PR #11, opened 2017-11-28
aether-38: extend jsonpath_ng with custom wildcards

quesebifurcan:
Resolves https://jira.ehealthafrica.org/browse/AET-38

TODO:
 - [x] add a predicate function `is_custom_jsonpath` + some more tests.

### DIFF
--- a/gather2-core/gather2/core/api/utils.py
+++ b/gather2-core/gather2/core/api/utils.py
@@ -1,7 +1,9 @@
 import string
 import json
+import re
 import uuid
 
+import jsonpath_ng
 from jsonpath_ng import parse
 
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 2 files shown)


gather2-core/gather2/core/api/utils.py
```diff
@@ -58,6 +60,48 @@ def json_printable(obj):
         return obj
 
 
+custom_jsonpath_wildcard_regex = re.compile('[a-zA-Z0-9_-]+\*')
```
admbtlr: Do we need a `.` in there to allow for more nested json paths, e.g. `$.facility.dose-*.id`?

admbtlr (changes requested):

gather2-core/gather2/core/api/utils.py
quesebifurcan: Good point -- @shawnsarwar said yesterday that aet-38 should only handle "top-level" custom wildcards. Which might be a good call -- the nested implementation would be significantly more complicated.

quesebifurcan: Come to think of it, I should probably add some more tests to make sure that we always fall back to `jsonpath_ng.parse` when we should. Here's a good list: https://github.com/json-path/JsonPath#path-examples

admbtlr: Yup, good call on the tests. And fair enough about ignoring nesting for now.

admbtlr (approved):

PR #14, opened 2017-12-04
aether-46: catch entity extraction errors

quesebifurcan:
If entity extraction fails during the creation of a
Response instance, raise `serializers.ValidationError`.

Resolves: https://jira.ehealthafrica.org/browse/AET-46


realdouno:
LGTM

PR #15, opened 2017-12-04
Changed Response to Submission

realdouno:
Changed Response to Submission
Solved: https://jira.ehealthafrica.org/browse/AET-60


gather2-core/entrypoint.sh
```diff
@@ -49,7 +49,7 @@ setup_db() {
     set -e
     psql -c "ALTER TABLE IF EXISTS api_project       RENAME TO core_project;"
     psql -c "ALTER TABLE IF EXISTS api_mapping       RENAME TO core_mapping;"
-    psql -c "ALTER TABLE IF EXISTS api_response      RENAME TO core_response;"
+    psql -c "ALTER TABLE IF EXISTS api_submission      RENAME TO core_submission;"
```
quesebifurcan, line 52: These sql statements are probably not necessary anymore, are they? If we can remove them, that would be great.

quesebifurcan (changes requested):
Looks good! Just that one thing with the sql scripts (see below)

gather2-core/entrypoint.sh
quesebifurcan: Fixed in https://github.com/eHealthAfrica/aether/pull/17

quesebifurcan (approved):

PR #18, opened 2017-12-05
update couchdb version to fix vulnerability

femilofin:


quesebifurcan:
I think this one might need an update too: https://github.com/eHealthAfrica/aether/blob/develop/gather2-couchdb-sync/couchdb/Dockerfile

Also, we might want to go for version `1.7.1` instead of `1.7.0` -- see https://github.com/apache/couchdb/issues/974

@femilofin

femilofin:
@quesebifurcan Updated to v 1.7.1

MrPink (approved):

PR #21, opened 2017-12-05
aether-47: exceptions

quesebifurcan:
Do not raise bare exceptions -- replace all instances of `raise Exception` with a more specific error type.


realdouno:
LGTM

PR #22, opened 2017-12-05
Broke extract_entity into multiple functions. Added unit tests for ea…

shawnsarwar:
…ch new function

Had to patch because of the rebase.


shawnsarwar:
Trying to fix github error

PR #29, opened 2017-12-07
update billing tags

femilofin:


terraform/dev/variables.tf
```diff
@@ -1,4 +1,4 @@
 variable "environment" { default = "dev" }
 variable "project" { default = "aether" }
-variable "project_billing_id" { default = "aether" }
+variable "project_billing_id" { default = "47200/NIGXXX" }
```
admbtlr: Can you put in a comment that says something like "Bill to Lake Chad Basin"? Just cos that code is not very obvious, and it might be useful to know what it means

admbtlr (approved):

PR #30, opened 2017-12-12
AET-79: add /schema endpoint

quesebifurcan:
Integrate drf-openapi.

kernel: change versioning scheme from NamespaceVersioning to
URLPathVersioning. This is necessary since drf-openapi only
supports the latter. See
https://github.com/limdauto/drf_openapi#4-constraints

**NOTE** this PR pins the django-restframework version to 3.6.4,
which is the highest available version which drf-openapi supports.
As soon as https://github.com/limdauto/drf_openapi/issues/94 is
addressed, we can remove this version constraint.

Resolves https://jira.ehealthafrica.org/browse/AET-79


quesebifurcan:
API documentation can be accessed locally at: http://kernel.aether.local:8000/v1/schema/

Example:
2017-12-12-135457_2246x2102

realdouno:
Good stuff

admbtlr (approved):

PR #31, opened 2017-12-13
Create avro schemas from salad schema

realdouno:
Create avro schemas from salad schema
Solves: https://jira.ehealthafrica.org/browse/AET-25


realdouno:
Note: Schemas are now unique per project, so multiple schemas can have the same name. We need to rethink the routing of schemas otherwise we will run into MultipleObjectsReturned exceptions

quesebifurcan:
> Note: Schemas are now unique per project, so multiple schemas can have the same name. We need to rethink the routing of schemas otherwise we will run into MultipleObjectsReturned exceptions

Good point, should be addressed by https://jira.ehealthafrica.org/browse/AET-70 (in progress)

quesebifurcan:
Will eventually be resolved in https://jira.ehealthafrica.org/browse/AET-96

PR #34, opened 2017-12-20
fix: avoid raising EntityExtractionError when no matches are found

quesebifurcan:
When jsonpath lookups fail, do not raise `EntityExtractionError` -- instead, append error to `aether_errors`.

The error formatting in `aether_errors` should probably be reworked to be more easily machine-parseable -- right now, it's just a custom message with the failed path attached.


admbtlr (approved):
Let's make a ticket to come up with a spec for log messages so that we can make them easily machine-parseable

PR #35, opened 2017-12-21
chore: build UI in both places

MrPink:
it seems to be clever enough to pick up the cached UI image so it doesn't rebuild it


djmoran64 (approved):
Looks good to me

admbtlr (approved):

PR #39, opened 2018-01-09
 feat(common): copy features from gather

obdulia-losantos:
- [x] Imported features from `gather2.common` module.
- [x] Added compatibility with Django 2.x

Once this is merged the distributed `common` packaged should be copy to `gather2` repo in the `gather3` "branch" as a replacement of `gather2.common`. Every module that links to Aether might require it...


quesebifurcan (approved):
Nice work!

PR #49, opened 2018-01-16
feat: make Submission.mapping non-nullable

quesebifurcan:
Resolves https://jira.ehealthafrica.org/browse/AET-111


admbtlr (approved):

obdulia-losantos (changes requested):
Sorry I merged my PR before :cry: You need to fix the migration files order...

aether-kernel/aether/kernel/api/models.py
```diff
@@ -90,7 +90,7 @@ class Submission(models.Model):
     map_revision = models.TextField(default='1')
     date = models.DateTimeField(auto_now_add=True, db_index=True)
     payload = JSONField(blank=False, null=False)
-    mapping = models.ForeignKey(to=Mapping, on_delete=models.CASCADE, blank=True, null=True)
+    mapping = models.ForeignKey(Mapping, related_name='submissions')
```
obdulia-losantos: could you reinclude the `on_delete...` option? :grimacing: It's needed to migrate to Django 2 (if we are going to do that). No migration file is required for this. Also the `related_name='submissions'` is useless with the new Meta option.

obdulia-losantos (approved):

PR #52, opened 2018-01-17
chore: upgrade postgres@9.6

obdulia-losantos:
This implies that the current data is going to be lost... not sure about how many deployments with data we have but no so much I think and either critical. :thinking: 

### To skip the problem locally

1.- Before start the app change the value back in `docker-compose-base.yml`

```yml
  postgres-base:
    image: postgres:9.5.2
```

2.- Rebuild the containers

```bash
docker-compose build
```

3.- In each container create a dump of the current database

```bash
docker-compose run {container} bash

export PGPASSWORD=$RDS_PASSWORD
export PGHOST=$RDS_HOSTNAME
export PGUSER=$RDS_USERNAME
export PGPORT=$RDS_PORT

pg_dump $RDS_DB_NAME > /media/data-backup.sql
```

The database backup will be in `tmp/{container}/data-backup.sql` in the project directory.

4.- Revert `docker-compose-base.yml` change

```yml
  postgres-base:
    image: postgres:9.6
```

5.- Rebuild containers

6.- Restore previous backups

```bash
docker-compose run {container} bash

export PGPASSWORD=$RDS_PASSWORD
export PGHOST=$RDS_HOSTNAME
export PGUSER=$RDS_USERNAME
export PGPORT=$RDS_PORT

dropdb -e $RDS_DB_NAME
createdb -e $RDS_DB_NAME -e ENCODING=UTF8
psql -e $RDS_DB_NAME < /media/data-backup.sql
```

7.- Start containers as usual.

### DIFF
--- a/aether-couchdb-sync-module/conf/docker/apt-packages.txt
+++ b/aether-couchdb-sync-module/conf/docker/apt-packages.txt
@@ -1,4 +1,4 @@
 curl
 gettext-base
-postgresql-client-9.5
+postgresql-client-9.6
 vim
--- a/aether-kernel/conf/docker/apt-packages.txt
+++ b/aether-kernel/conf/docker/apt-packages.txt
@@ -1,5 +1,4 @@
 curl
 gettext-base
-postgresql-client-9.5
-python-pypy.sandbox
+postgresql-client-9.6
 vim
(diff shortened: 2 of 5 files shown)


admbtlr (approved):
I think that now is probably a good time to do this, since there isn't any valuable data in Aether anywhere, but yeah, we should make everyone knows what to do...

shawnsarwar (commented):
I have no issue from an Aether standpoint. I'll deffer to someone else's judgement.

quesebifurcan (approved):

PR #56, opened 2018-01-18
chore: cleanup docker-compose setup

obdulia-losantos:
There is nothing like reading the docs :nerd_face: 

`./scripts/docker_start.sh` does the trick.


scripts/docker_start.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
```
admbtlr, line 1: If I start kernel then odk, `docker-compose` is smart enough to know that kernel is already running, isn't it?

obdulia-losantos: The next line will kill all running containers first... I can remove/comment that line in that case.

obdulia-losantos: docker-compose is smart enough to know which containers are running or not and does not restart them again or complain ;-)

admbtlr: cool. maybe you could hide the `set -e` behind a flag? like maybe `-f` to force killing and restarting the containers?

obdulia-losantos: better now?

obdulia-losantos:
@MrPink this one fixes all the chaos with the DC (AKA _Detective Comics_) files

scripts/docker_start.sh
admbtlr: beautiful!

admbtlr (approved):

obdulia-losantos:
😊

PR #57, opened 2018-01-19
added kernel client and tests

shawnsarwar:
The long awaited python client for the Aether Kernel API.
Integration testing with decent coverage.


aether-client/aether/client.py
```diff
@@ -0,0 +1,548 @@
+import json
+import requests
+
+
+class GenericClient(object):
+
+    # Base client type to be subclassed.
+    # Holds access to http requests
```
quesebifurcan: Can we turn these comments into class-level docstrings?

aether-client/aether/client.py
```diff
@@ -0,0 +1,548 @@
...
+    def refresh(self):
+        self.__dict__['Resource'] = DottedDict()
+        resource = self.__dict__['Resource']
+        for key, name in self._resource_types.items():
+            resource[key] = ResourceCollection(name, self)
+
+    def delete(self, url):
+        req = requests.delete(url, auth=self.auth)
+        ok = req.status_code in [requests.codes.ok, requests.codes.no_content]
+        return ok
```
quesebifurcan: Will it be possible for an application using this client to get access to the response codes? It looks like this will just return a boolean `False` on request failure, but perhaps this is handled somewhere else?

aether-client/aether/client.py
```diff
@@ -0,0 +1,548 @@
...
+
+    _type_lookup = {v: k for k, v in _resource_types.items()}
+    _holds_resources = {
+        "mappings": "Submission"
+    }
+
+    def refresh(self):
+        # Refreshes all Resource assets. Should be used sparingly.
+        # Better to call refresh on the asset directly instead of updating
+        # the entire tree.
```
quesebifurcan: Perhaps make this into a docstring too?

aether-client/aether/tests/test_client.py
```diff
@@ -0,0 +1,393 @@
...
+                fake_credentials = {"missing":"values"}
+                KernelClientCase.client = client.KernelClient(kernel_url, **fake_credentials)
+            except KeyError:
+                pass
+            try:
+                KernelClientCase.client = client.KernelClient(kernel_url, **kernel_credentials)
+                print("Connected client to Aether on %s" % kernel_url)
+            except Exception as e:
+                print(e)
+                self.fail("Could properly connect to Aether Kernel / setup client")
```
quesebifurcan: Could properly -> Could not properly?

docker-compose-test.yml
```diff
@@ -101,3 +101,14 @@ services:
...
     ports:
       - "9666:9666"
+
+  # ---------------------------------
+  # Aether Client container
+  # ---------------------------------
+  client-test:
+    build: aether-client
+    image: aether/client
+
```
quesebifurcan, line 111: If we add

```
depends_on:
  - kernel-test
  - db-test
```
it'll be easier to run this test in isolation.

aether-client/aether/tests/test_client.py
```diff
@@ -0,0 +1,393 @@
...
+            self.fail("Integration test failed with %s faults." % len(self.failures))
+        else:
+            print("Completed %s tasks without issue." % self.success)
+
+
+    def _steps(self):
+        # iterator to step through sequential test steps
+        for name in sorted(dir(self)):
+            if name.startswith("step"):
+                yield name, getattr(self, name)
```
quesebifurcan, line 70: This definitely works, nice idea. https://docs.python.org/3.6/library/unittest.html#unittest.TestLoader.sortTestMethodsUsing might be an alternative option.

aether-client/aether/tests/test_client.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def test_steps(self):
+        # The entrypoint for the test functions.
+        for name, step in self._steps():
+            try:
+                step()
+                self.success +=1
+                print(".")  # cheesey dot replicating normal unittest behavior
```
quesebifurcan: To save vertical space: `sys.stdout.write('.')` (dot without newline)

quesebifurcan (changes requested):
Looks great! 

Just some minor nitpicks (see comments).

How about including the `tests` directory in the flake8 config as well? Or just make a ticket to revisit it later.

aether-client/aether/tests/test_client.py
shawnsarwar: Yeah, I tried that and a few other tricks. Unfortunately unittests in python REALLY want to be unittests.

aether-client/aether/tests/test_client.py
shawnsarwar: I did this. Unfortunately they now all print at the end of all the tests which is a bit weird.

PR #64, opened 2018-01-23
chore: use port mapping

MrPink:
This is in relation to https://jira.ehealthafrica.org/browse/AET-110, I'm allocating ports instead of using the host network mode


docker-compose-kafka.yml
```diff
@@ -1,9 +1,9 @@
 # Adds Confluent Platform
-version: "2"
+---
```
obdulia-losantos: won't this complain? :thinking:

PR #65, opened 2018-01-25
 entity upsert with merge left or right options

lordmallam:
#### What this PR does
Creates the entities/put endpoint to update, accepts merge options and an existing entity id

#### Usage
`/entities/[entity uuid]/?merge=[merge option]`
or 
include a merge property in the body
`{
...
"merge": "[merge option]" 
...
}`

#### Merge Options
`last_write_wins` : target payload takes primacy
`first_write_wins` : source payload takes primacy
`overwrite` : target overwrites source (no merge is performed)

####  Related JIRA Ticket
[AET-71](https://jira.ehealthafrica.org/browse/AET-71)


shawnsarwar (changes requested):
Looks great overall. It may be difficult for API users to remember which integer based argument coincides with which merge strategy. I think a more descriptive string based arg would be preferable. Merge left and right aren't very clear. Perhaps something descriptive like, "overwrite", "prefer_new", and "prefer_existing".

shawnsarwar (approved):

aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -9,6 +9,14 @@
 
 import urllib
 
+MERGE_CHOICES = (
+    (utils.MergeOptions.overwrite.value, 'Overwrite (Do not merge)'),
+    (utils.MergeOptions.perfer_new.value, 'Prefer New (Target to Source)'),
```
quesebifurcan: Spelling: "perfer" -> "prefer"?

aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -243,20 +251,47 @@ class EntitySerializer(DynamicFieldsMixin, serializers.ModelSerializer):
         source='projectschema',
         view_name='projectschema-detail',
     )
+    update_id = serializers.CharField(allow_blank=True, default=None)
+    merge = serializers.ChoiceField(MERGE_CHOICES, default=utils.MergeOptions.overwrite.value)
 
     def create(self, validated_data):
```
quesebifurcan: Do we really want this logic inside of the `create` method? As I understood the issue, merge options need only be available when using `PUT` or `PATCH`. And if you override the `update` method instead, you would already have access to the instance id (and can skip the `update_id` parameter).

aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -243,20 +251,47 @@ class EntitySerializer(DynamicFieldsMixin, serializers.ModelSerializer):
...
+                                or merge_value == M_OPTIONS.perfer_new.value):
+                            entity.payload = \
+                                utils.merge_objects(payload, entity.payload, merge_value)
+            entity.payload['_id'] = str(entity.id)
+            entity.save()
+            return entity
+        except Exception as e:
+            raise serializers.ValidationError({
+                'description': 'Submission validation failed'
+            })
```
quesebifurcan, line 274: It would be good to have tests for this (to make sure we have 100% coverage of all if-branches).

quesebifurcan (changes requested):

aether-kernel/aether/kernel/api/serializers.py
lordmallam: Test included for merge option (fww)

aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -234,6 +242,7 @@ class Meta:
 
 
 class EntitySerializer(DynamicFieldsMixin, serializers.ModelSerializer):
+    lookup_field = 'merge'
```
quesebifurcan, line 245: Quoting from Slack: 

> what do you think about putting the upsert logic in a separate function based view under its own endpoint? That would keep things decoupled and easily modifiable (no need to change the existing Entity.create method). Something like:
> 
> ```/entity-merge/<primary-key-of-target>?strategy=prefer_new```

This would mean having a separate (view) function (in `aether/kernel/views.py`) which contains all the merge logic. And there would be no need to touch this serializer. The problem here is that we're modifying `lookup_field`, which impacts both routing and the hyperlinks between serializers.

quesebifurcan (changes requested):

quesebifurcan (approved):

PR #68, opened 2018-01-26
fix(odk): remove first level (None) in submissions payload

obdulia-losantos:
Related to: https://jira.ehealthafrica.org/browse/AET-125

### DIFF
--- a/aether-odk-module/aether/odk/api/views.py
+++ b/aether-odk-module/aether/odk/api/views.py
@@ -1,10 +1,5 @@
 import json
-import re
 import requests
-import xmltodict
-
-from dateutil import parser
-from geojson import Point
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


aether-odk-module/aether/odk/api/views.py
```diff
@@ -383,34 +379,6 @@ def xform_submission(request):
 
     '''
 
-    def walk(obj, parent_keys, coerce_dict):
```
admbtlr, line 386: It's such a relief to get this code out of the view...

admbtlr (approved):
🎉 

Can we also remove the `data.` from headings in the UI and CSV export? I guess that would be a Gather ticket...

obdulia-losantos:
I think that remove "data." is half Kernel half Gather :thinking:

PR #69, opened 2018-01-26
feat(common): CSVRenderer: parse header labels

obdulia-losantos:
Continues https://jira.ehealthafrica.org/browse/AET-125

Include in the url the param `parse_columns` with the rules to generate csv headers. See tests for more examples.

In this use case: `parse_columns=remove-prefix:payload.`

For CHAMPS: `rule_sep=$&parse_columns=remove-prefix$payload.,replace$.$:`


aether-common-module/aether/common/drf/renderers.py
```diff
@@ -40,7 +37,89 @@ def render(self, data, media_type=None, renderer_context={}, *args, **kwargs):
...
+            return value[len(prefix):]
+        return value
+
+    if rule.startswith('remove-suffix:') and rule != 'remove-suffix:':
+        suffix = rule.split(':')[1]
+        if value.endswith(suffix):
+            return value[:-len(suffix)]
+        return value
+
+    if rule.startswith('replace:'):
```
admbtlr: Wow, this is amazing, you actually read my mind... I'd forgotten that the final thing we need to do for CHAMPS to change the '.' separator on the column headings into a ':' 

💯!

aether-common-module/aether/common/drf/renderers.py
```diff
@@ -40,7 +37,89 @@ def render(self, data, media_type=None, renderer_context={}, *args, **kwargs):
...
+        old_sep = rule.split(':')[1]
+        new_sep = rule.split(':')[2]
+        return value.replace(old_sep, new_sep)
+
+    if rule.startswith('split:') and rule != 'split:':
```
admbtlr: Is this any different to `replace:<sep>: `?

admbtlr (changes requested):

aether-common-module/aether/common/drf/renderers.py
obdulia-losantos: Nop, just a shortcut of `replace:<sep>: :` :thinking:

aether-common-module/aether/common/drf/renderers.py
obdulia-losantos: In this case I'm not sure if I should use another rule separator like... `|` maybe

aether-common-module/aether/common/drf/renderers.py
admbtlr: OK cool, just checking

aether-common-module/aether/common/drf/renderers.py
admbtlr: `\:`?

obdulia-losantos: I meant `replace|.|:` I'm quite sure that there are no `|` in the object keys :crossed_fingers:

obdulia-losantos:
```
self.assertEqual(apply_label_rules(['replace:.:\:'], 'a.bcd'), 'a:bcd')

AssertionError: 'a\bcd' != 'a:bcd'
- a\bcd
+ a:bcd
```

admbtlr (approved):

PR #78, opened 2018-02-08
Aet 42 add producer

shawnsarwar:
This PR includes the aether-connect Kafka producer as well as Kafka integration. We need a testing strategy for Kafka connect to include the aether-producer. That's not covered in this PR. Additionally, the requirement for the aether-client library for aether-producer is currently being satisfied in the same way that we include aether.common -- via a wheel in the conf/pip/dependencies folder. I didn't find the automated build process for aether.common, so someone more knowledgeable in that department should sound off on how best to add build of aether-client for inclusion in modules that need it.


quesebifurcan:
What's the best way to test this (i.e. run locally)? I tried with various combinations of the different compose files but didn't get very far.

shawnsarwar:
The simplest setup would be to post a schema & projectschema as seen in the aether-client/tests and then post mock data against it, also as seen in aether-client/tests. Then you'd need to have a simple kafka-consumer look at the topic for the entity you created and make sure everything you posted to Aether was available in Kafka. It would pretty closely mirror the AetherToken Game.

quesebifurcan (approved):

PR #79, opened 2018-02-12
Aet 37 entity linking

lordmallam:
#### What does this PR do??
Includes linked entity data as nested objects on an entity `GET` request

#### How to test
Make a `GET` request on a entity with a `include_linked_data` and `depth(optional)` query strings
`/entities/<entityId>/?include_linked_data=true&depth=2`
This should return nested linked entities

#### Related JIRA ticket
[AET](https://jira.ehealthafrica.org/browse/AET-37)


aether-kernel/aether/kernel/api/views.py
```diff
@@ -62,6 +85,22 @@ class EntityViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.EntitySerializer
     filter_class = filters.EntityFilter
 
+    def retrieve(self, request, pk=None):
+        try:
+            selected_record = models.Entity.objects.get(pk=pk)
+        except Exception as e:
+            return Response(str(e), status=HTTPStatus.NOT_FOUND)
+        include_linked_data = request.GET.get('include_linked_data')
```
quesebifurcan: You can use `query_params` here instead of `GET`. See: http://www.django-rest-framework.org/api-guide/requests/#query_params

aether-kernel/aether/kernel/api/views.py
```diff
@@ -1,10 +1,33 @@
 from django.db.models import Count, Min, Max
 from rest_framework import viewsets, permissions
 from drf_openapi.views import SchemaView
+from rest_framework.response import Response
+from http import HTTPStatus
 
 from . import models, serializers, filters
 
 
+def get_entity_linked_data(entity, request, depth, start_depth=0):
```
quesebifurcan: Test coverage drops from 84 to 81 percent in this PR -- we need to make sure all new functions and classes are properly covered, so please add a couple of tests for this function.

aether-kernel/aether/kernel/api/views.py
```diff
@@ -62,6 +85,22 @@ class EntityViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.EntitySerializer
     filter_class = filters.EntityFilter
 
+    def retrieve(self, request, pk=None):
```
quesebifurcan, line 88: Same here -- tests please :-)

aether-kernel/aether/kernel/api/views.py
```diff
@@ -62,6 +85,22 @@ class EntityViewSet(viewsets.ModelViewSet):
...
+            selected_record = models.Entity.objects.get(pk=pk)
+        except Exception as e:
+            return Response(str(e), status=HTTPStatus.NOT_FOUND)
+        include_linked_data = request.GET.get('include_linked_data')
+        if include_linked_data == 'true':
+            depth = request.GET.get('depth')
+            if depth:
+                depth = int(depth)
+            else:
+                depth = 1
```
quesebifurcan: Maybe it's possible to use only `depth` here and skip `include_linked_data`? Since we're checking `depth` anyway, `include_linked_data` is a bit redundant.
If depth is present in the query params, attempt to retrieve linked data; otherwise, skip this entire block.

quesebifurcan (changes requested):
Looks good, great work!

Added some comments, but would like to test it properly. Which schemas and submissions data did you use for testing locally? Can you share them somehow?

aether-kernel/aether/kernel/api/views.py
lordmallam: For sample data, I ran the `./generate.sh` script from the `dev-demo` branch against the endpoints

quesebifurcan: Ok, thanks! I didn't get that branch to work, but `dev-demo-confluent` seems to generate entities.

aether-kernel/aether/kernel/api/views.py
```diff
@@ -1,10 +1,33 @@
...
+                linked_data_ref = entity.payload.get(item['name'])
+                linked_entity = models.Entity.objects.filter(payload__id=linked_data_ref)
+                entity.payload[item['name']+'_data'] = serializers.EntitySerializer(
+                    linked_entity, many=True, context={'request': request}).data[0]
+                while (start_depth < depth):
+                    start_depth = start_depth + 1
+                    entity.payload[item['name']+'_data'] = serializers.EntitySerializer(
+                        get_entity_linked_data(linked_entity.first(), request, start_depth),
+                        context={'request': request}).data
+                    Response('' + start_depth)
```
quesebifurcan: What is this line doing?

aether-kernel/aether/kernel/api/views.py
```diff
@@ -1,10 +1,33 @@
...
+def get_entity_linked_data(entity, request, depth, start_depth=0):
+    schema_definition_fields = entity.projectschema.schema.definition['fields']
+    jsonld_fields = [item for item in schema_definition_fields if item.get('jsonldPredicate')]
+    for item in jsonld_fields:
+        try:
+            if item['jsonldPredicate'].get('_id'):
+                linked_data_ref = entity.payload.get(item['name'])
+                linked_entity = models.Entity.objects.filter(payload__id=linked_data_ref)
+                entity.payload[item['name']+'_data'] = serializers.EntitySerializer(
+                    linked_entity, many=True, context={'request': request}).data[0]
```
quesebifurcan: We would probably need to put the payload of the linked object where the id should have been.

Guideline: the entity with resolved links _must_ conform to the same schema as the entity without resolved links (otherwise other validations would fail).

For example:

```
{                                                                                                                                                                       
  "payload": {                                                                                                                                                          
    "id": "46f0d0f3-7117-497a-a2d4-96b8e88072f9",                                                                                                                       
    "_id": "220af2d0-641e-44ae-9903-d754e24ce07a",                                                                                                                      
    "age": 98,                                                                                                                                                          
    "isAlive": true,                                                                                                                                                    
    "lastName": "knight",                                                                                                                                               
    "firstName": "lee",                                                                                                                                                 
    "householdID": {                                                                                                                                                    
      "id": "05151ef6-f819-4b15-8a5a-132c4b0927b7",                                                                                                                     
      "_id": "38a722a5-f96d-460f-8c18-b63a1ae7efb4",                                                                                                                    
      "sectionID": "2",                                                                                                                                                 
      "hasBedNets": true,                                                                                                                                               
      "locationID": "d795419e-1e02-45fc-993a-e9fbdb5d2396",                                                                                                             
      "headOfHouseHold": "46f0d0f3-7117-497a-a2d4-96b8e88072f9",                                                                                                        
      "locationID": {                                                                                                                                                   
        "id": "d795419e-1e02-45fc-993a-e9fbdb5d2396",                                                                                                                   
        "_id": "53df87f3-d9b1-46ac-8424-5fab8f325253",                                                                                                                  
        "lat": 6.976889,                                                                                                                                                
        "lng": -9.059241        
```
(Removed some fields for simplicity)

Ping @shawnsarwar

quesebifurcan (changes requested):

aether-kernel/aether/kernel/api/views.py
shawnsarwar: I agree that depth would be sufficient. If depth is greater than the max depth we support (3), we should probably have an error as well.

aether-kernel/aether/kernel/api/views.py
quesebifurcan: Or some other solution, I'll bring it up in the channel.

aether-kernel/aether/kernel/api/views.py
lordmallam: Included tests for this

quesebifurcan (approved):

PR #83, opened 2018-02-20
Feat: docker versioning

MrPink:
* This currently breaks because of the directory structure  i.e `aether-odk-module`


shawnsarwar (commented):
If I understand correctly, now if we run release.sh it'll we'll make new images and send them to docker hub as the version we set in aether-kernel/VERSION? Or do we need to set each version manually before we release?

MrPink:
@shawnsarwar it sets whatever is in VERSION correct, I guess having a git tag creation script that you can run yourself would also be beneficial

admbtlr:
Wouldn't it make sense to have just one VERSION file for the whole repo, since we're going to be releasing everything in lockstep?

MrPink:
@adamvert @quesebifurcan @shawnsarwar I could set a Travis env for the version and pass it in as a build-arg, we can have a symbolic `VERSION` file in the root of the repository

quesebifurcan:
@MrPink that sounds good. Would simplify things a bit and still make version bumping subject to code review.

admbtlr:
So you mean you would set the env var based on the value contained in the `VERSION` file? If so, 👍

MrPink:
@adamvert yeah exactly

admbtlr (approved):
LGTM

quesebifurcan (approved):
Nice!

PR #85, opened 2018-02-21
feat: implement retrieve/list views with POST

obdulia-losantos:
_Copying last **Gather2** patch here_


aether-kernel/aether/kernel/api/views.py
```diff
@@ -1,17 +1,39 @@
 from django.db.models import Count, Min, Max
 from rest_framework import viewsets, permissions
+from rest_framework.decorators import detail_route, list_route
 from drf_openapi.views import SchemaView
 
 from . import models, serializers, filters
 
 
-class ProjectViewSet(viewsets.ModelViewSet):
+class CustomViewSet(viewsets.ModelViewSet):
```
obdulia-losantos, line 9: As a comment here  `rest-framework@3.8` is going to replace the ` detail_route` and `list_route` decorators with `action` decorator that will also allow to indicate the url path. In that case the url pattern could be the same, now is based only on method name.

https://github.com/encode/django-rest-framework/blob/master/rest_framework/decorators.py#L133-L154

admbtlr (approved):

PR #86, opened 2018-02-23
add kong component

lordmallam:
#### What does this PR do?
Adds Kong as a component to the docker compose structure reachable on kong admin: `http://kong:8001` kong apis: `http://kong:8000`

#### How to test
Run `docker-compose up --build kernel`
Kong container is started with an ephemeral `kong-migration` container that handles migrations to the db
Kong admin should be reachable via `http://kong:8001` from within the `kernel` container or `kong.aether.local:8001` remotely (this requires that `kong.aether.local` be added to the /etc/host file)

#### Related JIRA ticket
[AET-162](https://jira.ehealthafrica.org/browse/AET-162)


shawnsarwar (commented):
Looks good. I can hit it on localhost:8001
Oops. I meant this for #87

PR #88, opened 2018-02-26
Aet 163 define api routing

lordmallam:
#### What does this PR do?
adds `kongfig` as a component and defines apis for `kernel`, `odk`, `couch-sync`

#### How to test
`docker-compose -f ./docker-compose.yml up --build`
check that you can reach kernel from `http://localhost/aether/v1` or `kong.aether.local/aether/v1`

#### Related JIRA ticket
[AET-163](https://jira.ehealthafrica.org/browse/AET-163)


shawnsarwar (changes requested):
Looks really good! The only issue I'm having is that static assets from django aren't being loaded when accessed through Kong. I'll paste a screenshot in slack.

shawnsarwar (approved):
Have't had a check on the static files, but assuming that's sorted, go ahead -> 150

PR #93, opened 2018-03-01
Fix: avro schema validation issues: aet 172

shawnsarwar:
This patch addresses issues in the avro-python3 library by replacing it with fastavro in the kernel module.  avro-python3 was not allowing some valid schemas that passed validation with the python2 version of the Apache maintained avro python library. Since the two versions of official avro-python are so divergent, we should start migrating all instance of avro-python3 to fastavro. We're still somewhat stuck with python-avro as it's used in schema-salad-tool.


aether-kernel/conf/pip/requirements.txt
```diff
@@ -16,11 +16,11 @@ chardet==3.0.4
 click==6.7
```
lordmallam, line 16: All good. I'm guessing `avro-python3` is no longer a dependency?

shawnsarwar: Yeah, we should try to avoid using avro-python3 in favor of fastavro. avro-python3 is not always compatible with schemas that are valid using the python2 avro package. Since schema-salad requires the py2 version, it's permissible when using python2 but standardizing on fastavro should be our goal.

lordmallam (approved):

PR #94, opened 2018-03-05
fix: make client tests return a non-zero exit code on test failure.

quesebifurcan:
Make client tests return a non-zero exit code on test failure.


admbtlr:
So is it expected that this PR fails on Travis?

quesebifurcan:
@adamvert yes, that's the fix. Shawn is working on an update to the client code to make it pass.

shawnsarwar:
Issues are fixed and it's ready for merge into 150.

admbtlr (approved):
LGTM

PR #98, opened 2018-03-07
Chore: add helm

MrPink:
Welcome to Helm! More information about Helm can be found here https://helm.sh/

**These will most likely evolve in time and come with fewer defaults**

* This allows us to package Aether Kubernetes resources as Helm packages which removes the complexity of distributing these resources in an organised and efficient manner. This is comparable to Debian's APT package system

* We host these Charts on http://charts.eha.im:8080/api/charts 
* The Helm charts use an override YAML where we can specify things to specific to the environment

**TO DO**
* Helm release and deployment workflow using Travis CI


helm/kernel/values.yaml
```diff
@@ -0,0 +1,38 @@
...
+release: 0.5.1
+service:
+  name: nginx
+  type: NodePort
+  externalPort: 80
+  internalPort: 80
+environment: "dev"
+ums_url: ums-dev
+domain: gather2.org
+trusted_origins: .ehealthafrica.org
```
quesebifurcan: Not sure if `.ehealthafrica.org` is enough -- we might need `.gather2.org` instead/as well. Let's test and see.

helm/odk/templates/NOTES.txt
```diff
@@ -0,0 +1,19 @@
...
+  http://{{ . }}
+{{- end }}
+{{- else if contains "NodePort" .Values.service.type }}
+  export NODE_PORT=$(kubectl get --namespace {{ .Release.Namespace }} -o jsonpath="{.spec.ports[0].nodePort}" services {{ template "odk.fullname" . }})
+  export NODE_IP=$(kubectl get nodes --namespace {{ .Release.Namespace }} -o jsonpath="{.items[0].status.addresses[0].address}")
+  echo http://$NODE_IP:$NODE_PORT
+{{- else if contains "LoadBalancer" .Values.service.type }}
+     NOTE: It may take a few minutes for the LoadBalancer IP to be available.
+           You can watch the status of by running 'kubectl get svc -w {{ template "odk.fullname" . }}'
+  export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ template "odk.fullname" . }} -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
```
quesebifurcan, line 13: Does the helm linter catch jsonpath lookup errors?

helm/odk/templates/db-init-config-map.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ConfigMap
+metadata:
+  name: odk-db-init 
+data:
+  db-init.sh: |
+    #!/bin/bash
+
+    psql -c "CREATE DATABASE $RDS_DB_NAME;"
+    psql -c "CREATE USER $RDS_USERNAME WITH PASSWORD '$RDS_PASSWORD';"
+    psql -c "GRANT ALL PRIVILEGES ON DATABASE $RDS_DB_NAME TO $RDS_USERNAME;"
```
quesebifurcan: Should we run bash scripts with some options perhaps?
`set -e` or `set -Eeuo(x) pipefail` (depending on the context)?

quesebifurcan (changes requested):

helm/odk/templates/db-init-config-map.yaml
MrPink: well caught, I think `set -e` suffices

helm/odk/templates/NOTES.txt
MrPink: this outside the boundary of Helm's duties

quesebifurcan (dismissed):
Great! Looks good to me

quesebifurcan (dismissed):

quesebifurcan (approved):

PR #99, opened 2018-03-07
settings and environment variables

quesebifurcan:
Disallow unset variables in all bash scripts. (`-u`)
Immediately exit a pipeline when encountering a non-zero exit code (`-o pipefail`)

Add assertion for and documentation of `AETHER_KERNEL_URL` in `common.settings`.

This PR partly addresses some of the issues we encountered during the deployment of gather-on-aether.

Suggestions for further improvements:
- Avoid creation of new environment variables in the entrypoint files
- Avoid lookups on os.environ outside of the settings file (this results in a more predictable application state)


aether-client/conf/docker/setup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -Eeuo pipefail
 set -x
```
quesebifurcan: note to self: bring the `x`s up one line.

shawnsarwar (approved):
-Eeuox pipefail is interesting.

PR #100, opened 2018-03-08
chore: upgrade and normalize pip dependencies

obdulia-losantos:
- `psycopg2` is going to be replaced by `psycopg2-binary` in future releases.
- `avro-python3` missed as a primary requirement in `kernel`

It's up to you to continue to use this `pip_freeze` endpoint or to remove it and do the upgrades manually.


shawnsarwar (approved):
Agreed. I think we should continue to use pip freeze.

PR #103, opened 2018-03-08
Aet 181 provision token for individual users

lordmallam:
#### What does this PR do
Set kong plugin security per kong api as against globally
Setup oauth2 plugin on kernel apis; to be extended to all aether core modules
Setup key auth on token generation endpoint
Generates `client_id` and `client_secret` associated to a kong consumer which is in turn exchanged for an `access_token`
Generates `access_token` which expires in 24hr for a user already authenticated by a 3rd party app, used to grant access to kong gateway (kernel apis for now)

#### How to test
Try accessing kernel apis `api.aether.local/aether/v1`, returns 401
Request for a token from `POST api.aether.local/token` with the following data
`{redirect_uri:string, username:string, password:string, authenticated_userid:string}` also if an authorization token was gotten for the 3rd party auth service, include it in the header of your request, it will be passed onto the upstream service eventually
Your response should look like this : 
`{"refresh_token":"kFqYVKJIYX078Ul2XYdtqLfMQnw4cBRd","token_type":"bearer","access_token":"91kO3MgxQ27GWOUcK4F7kA05apyrSweT","expires_in":86400}
`
Use the `access_token` in your requests headers to the kernel apis now
You should be able to go beyond kong now

#### Related JIRA ticket
[AET-181](https://jira.ehealthafrica.org/browse/AET-181)


shawnsarwar (approved):
This looks good. Is there some way that we can generalize setup_kong_consumer more? Maybe make it or a subset of it accessible from aether.common?

lordmallam:
It's a good idea for the kong consumer to be setup in the common module, but there is no way of testing this feature because it depends on a lot of kernel and kong related objects

PR #110, opened 2018-03-13
feat: (kong) modularize kong configuration

lordmallam:
#### What does this PR do?
Adds a `kongfig.yml` file to all core modules for module-level kong elements registration

#### How to test
Modify `kongfig.yml` file on any of the core modules `kernel, odk, couchdb-sync`
Run `docker-compose -f  ./docker-compose.yml up --build`
Check that the apis, plugins, consumers are correctly registered on kong `http://api.aether.local:8001`

#### Related JIRA Ticket
[AET-167](https://jira.ehealthafrica.org/browse/AET-167)


shawnsarwar (changes requested):
Does Kongfig run from inside each container? Without Kong running there's no need to register with it, so it should be possible to run Kongfig from a single place which should keep us from having to install/ run node in each container.

lordmallam:
@shawnsarwar Yes, with this implementation `kongfig` runs on each container. So the issue was separation of concerns regarding individual modules kongfig, so as to specify how module-wide endpoints are registered, pointing to allowing some odk endpoints as publicly accessible as an example. This way the dev will only deal with say odk kongfig or ui-module kongfig. Core aether modules (`kernel`) depends on kong to be running. so technically kong will always be running. Alternatives to avoid the multiple install node issue will be: 

1. Have a centralized kongfig container with one kongfig.yml file where the dev will register all modules endpoints
2. Have individual kongfig.yml files in each module and have one kongfig container to compile all kongfig files and perform the registrations from one place.

PR #112, opened 2018-03-13
fix: handle duplicate submissions

quesebifurcan:
Fix duplicate submissions.

When handling submissions containing multiple attachments,
ODK Collect will split the submission into multiple POST
requests. Using the instance id of the submission, we can
assign the attached files to the correct submission,
avoiding duplicate submissions.

Add gather-specific `instanceID` filter to
`aether.kernel.api.models.Submission`.

The code makes some assumptions:
1. The submission has an instance id, accessible at
   `data['meta']['instanceID']`.
2. The instance id is globally unique.
3. The form data in the submission is identical for all
   POST requests with the same instance id.

Resolves:
https://jira.ehealthafrica.org/browse/AET-183
https://jira.ehealthafrica.org/browse/GTH-115


obdulia-losantos (dismissed):
LGTM!

aether-odk-module/aether/odk/api/views.py
```diff
@@ -281,32 +281,61 @@ def xform_submission(request):
...
+        # Collect will split the submission into multiple POST requests. Using
+        # the instance id of the submission, we can assign the attached files
+        # to the correct submission.
+        #
+        # The code in this block makes some assumptions:
+        #   1. The submission has an instance id, accessible at `data['meta']['instanceID']`.
+        #   2. The instance id is globally unique.
+        #   3. The form data in the submission is identical for all
+        #      POST requests with the same instance id.
+        instance_id = get_instance_id(data)
```
admbtlr: Maybe you could add a line like `(These assumptions match the OpenRosa spec linked to above)`, just to avoid future confusion

admbtlr (changes requested):
Nice!

obdulia-losantos (approved):
:writing_hand:

admbtlr (approved):

PR #114, opened 2018-03-14
chore: bump version -> 0.5.3

obdulia-losantos:


quesebifurcan:
I see why we need this, just need to check with Adam and Shawn what version strategy they want. If I understood things correctly, `develop` will only build nightlies and releases will be made from separate release branches. I'll check back as soon as I know more :smile:

obdulia-losantos:
OK, it's needed in this PR: https://github.com/eHealthAfrica/gather2/pull/364

Thanks!

obdulia-losantos:
Not needed

PR #124, opened 2018-04-05
feat: test aether connect

shawnsarwar:
This PR pertains to Aether Utilities (python client, salad-bar and data mocker); moving them into a file structure that makes sense and using them to perform integration tests.

To test integration run scripts/test_integration.sh
Integration tests run as part of scripts/test_all.sh

Integration tests live in "test-aether-integration-module"; suggest a better name, please.

The following functionality is tested:
- Automated registration of a schema using saladbar
- Automated generation of entities against the registered schemas using data mocker.
- Automated push from Aether -> Aether Connect using the Aether Producer
- Consistency check of data generated against the entities which can be found in Kafka.

I used pytest fixtures to handle the various resource connections. If you're curious they're in the __init__.py of /test-aether-integration-module/test/

We also have to run different parts of the integration test with python2 and python3, thanks to schema-salad-tool's using the python2 avro library. pip2 and pip3 keep things clean in the docker environment and pytest allows a function annotation that we've written which skips tests pertaining to the wrong version of python. The python2 tests run first since they also setup the schemas.


aether-producer/entrypoint.sh
```diff
@@ -98,6 +98,14 @@ case "$1" in
         ./manage.py # runserver 0.0.0.0:$WEB_SERVER_PORT
     ;;
 
+    start_test )
+        #setup_db
+        #setup_initial_data
```
quesebifurcan: Commented code, can we remove these lines?

aether-utils/aether-client/entrypoint.sh
```diff
@@ -84,8 +84,22 @@ case "$1" in
...
+        python setup.py bdist_wheel --universal
+
+        # remove useless content
+        rm -rf build
+        rm -rf aether.client.egg-info
+    ;;
+
     start )
-        echo 'start'        
+        echo 'start'
```
quesebifurcan: Is this used at all? Or just for debugging purposes?

aether-utils/aether-mock-data/entrypoint.sh
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
+set -e
```
quesebifurcan: If possible, I'd prefer the same header as in the other bash files:
```
set -Eeuox pipefail
```
Just to limit the number of runtime surprises.

aether-utils/aether-mock-data/entrypoint.sh
```diff
@@ -0,0 +1,116 @@
...
+        rm -rf aether.mocker.egg-info
+    ;;
+
+    start )
+        echo 'start'
+    ;;
+
+    start_dev )
+        echo 'start_dev'
+    ;;
```
quesebifurcan: Same here -- if they aren't really used, we could remove them (and update the help description).

quesebifurcan (changes requested):

shawnsarwar (commented):
Fixed issues raised.

quesebifurcan (approved):

PR #125, opened 2018-04-06
chore: improve node setup

obdulia-losantos:


aether-ui/entrypoint.sh
```diff
@@ -0,0 +1,208 @@
+#!/bin/bash
+set -e
```
quesebifurcan: In the other entrypoint scripts, we're using
```
set -Eeuox pipefail
```
to increase predictability a bit. It'll fail for unset variables, among other things. Would be nice to have here, too.

aether-ui/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+FROM python:3.6 AS ui_py
+
+
+################################################################################
+## setup container
+################################################################################
+
+## make node+npm available within the python container
+COPY --from=ui_node /usr/local/bin/nodejs        /usr/local/bin/nodejs
+COPY --from=ui_node /usr/local/lib/node_modules  /usr/local/lib/node_modules
```
quesebifurcan, line 40: Nice! Keeps the container size down

quesebifurcan (changes requested):

quesebifurcan (approved):

PR #129, opened 2018-04-10
fix(ui): remove useless components

obdulia-losantos:
Trying to keep it simple... if we need one of those components in the future we can bring it back.


aether-ui/aether/ui/assets/apps/components/common/index.jsx
```diff
@@ -1,25 +0,0 @@
-// import all components
```
lordmallam: I extracted the `NavBar` to the common folder and referenced in this file 
[https://github.com/eHealthAfrica/aether/blob/ui-pipeline-state/aether-ui/aether/ui/assets/apps/components/common/index.jsx#L12](https://github.com/eHealthAfrica/aether/blob/ui-pipeline-state/aether-ui/aether/ui/assets/apps/components/common/index.jsx#L12). I also think this will be useful down the line for common components

obdulia-losantos: I saw it in your PR :joy: 

What can we do?

lordmallam: Probably resolving the file conflict will suffice  😬

obdulia-losantos: :joy: 

:building_construction:

lordmallam (approved):
Nice tests

PR #131, opened 2018-04-12
Mock ui inputs entity types

lordmallam:
#### What does this PR do
Loads mock schema and converts to markup for pipeline input and entity types

PR #132, opened 2018-04-13
feat(pipeline): data model with validation

obdulia-losantos:
I'm trying to understand how the data model could be. Thereby I will have a better comprehension on what can be done in the frontend.

So far this is what I understood, I'm not sure of the data model but I think it's better if we make our comments here and redesign the data model. So feel free to make any comment, push commits with better designs and whatever you feel will help.

I'm not talking about how to communicate with kernel to validate the data or push it to create the project structure once everything is done... it's just about the data model needed to work on the frontend side.

Thanks!


aether-ui/aether/ui/api/models.py
```diff
@@ -153,6 +154,104 @@ class Meta:
...
+
+    +------------------+             +------------------+
+    | Pipeline         |             | Entity Type      |
+    +==================+             +==================+
+    | id               |<-----+      | id               |
+    | name             |      |      | name             |
+    | input            |      |      | payload          |
+    | mapping          |      |      +::::::::::::::::::+
+    | mapping_errors   |      +-----<| pipeline         |
+    +------------------+             +------------------+
```
lordmallam: First question: how do you come up with these amazing in-code ERDs? 😍

aether-ui/aether/ui/api/models.py
```diff
@@ -153,6 +154,104 @@ class Meta:
...
+        return self.name
+
+    def save(self, *args, **kwargs):
+        if self.mapping:
+            # in kernel: validate_mapping(submission_payload, entity_list, mapping)
+            self.mapping_errors = validate_mapping(
+                submission_payload=self.input,
+                entity_list=self.entity_types.all(),
+                mapping=self.mapping,
+            )
```
quesebifurcan: I wonder if it's necessary to save the error details in the database -- this is potentially also something which could be kept in the client state. And whenever we re-open a Pipeline, we just run the validation rules again (we'll have to do that anyway to fill the `output` field).

quesebifurcan (commented):
Looks good! Just one comment re:error details (see above)

aether-ui/aether/ui/api/models.py
obdulia-losantos: So, in your opinion it's better to call the validation method from the frontend every time you change the something in the pipeline.  :thinking: makes sense!

My concern was that the first page, pipeline list, shows the number of errors and that means that you are going to call the validation method every time the page is refreshed/loaded for each available pipeline. I was trying to avoid that.

quesebifurcan: I think both methods would work; it's also fine with me to have everything in one place like you do now. Might make it easier to test. A field `error_count` would also work.

aether-ui/aether/ui/api/models.py
```diff
@@ -153,6 +155,99 @@ class Meta:
...
+
+    # this is an example of the submission payload
+    input = JSONField(blank=True, null=True)
+    # this is the avro schema
+    schema = JSONField(blank=True, null=True)
+
+    # this represents the list of mapping rules
+    # {
+    #    "mapping": [
+    #      ['xpath-input-1', 'xpath-entity-type-1'],
```
quesebifurcan: Maybe change `xpath` to `jsonpath` to clarify that the input is always json?

quesebifurcan (changes requested):

aether-ui/aether/ui/api/models.py
obdulia-losantos: :+1:

quesebifurcan (approved):

obdulia-losantos:
@lordmallam I'll wait for your comments since this will affect what we, mostly you, are building in the frontend. :grimacing:

obdulia-losantos:
Reviewing #131 I saw that the entity types go in one text area, should I change the data model to reflect this and speed up the frontend develop? @lordmallam

lordmallam:
Yes, as far as i know the entityTypes on the ui will remain a json to generate a visual representation of the types for each pipeline

obdulia-losantos:
Ok, changing again! :D

lordmallam:
I think we will also need a way to associate the pipeline with a specific project to allow pipeline config for multiple projects (data submission sources)

obdulia-losantos:
Yeah! But we should discuss this point deeper and find out an easy way to do that :thinking: 

Maybe when we finalize the current tasks :building_construction:

lordmallam (approved):

aether-ui/aether/ui/api/models.py
obdulia-losantos: :joy: 

Made them manually! Takes time but they are really useful to have the big picture. ;-)

PR #133, opened 2018-04-17
DO NOT MERGE: request api (deprecated)

obdulia-losantos:
@quesebifurcan @lordmallam 

This is my first approach to the request API. I'm not quite sure if this is the best option but this is what I already implemented in other projects (and it's working fine). I'm trying to make it more "redux" complaint but I'm a little lost on that side (more than one year without using it make things slow down).

I would appreciate if you can lead me in the right direction :pray: 

I'm sure that the `PaginationContainer` component could use a  reducer to handle its state... I'll do my best on that...

- [x] implemented `request` utility to communicate with the backend side.
- [x] re-added components to control the data workflow.
- [x] implement "Add pipeline" using the request utility.
- [x] cleaning and testing.
- [ ] styling new components :woman_artist: 

Open questions:
- if we are not going to use reducers to handle the data workflow we could remove `redux-thunk`
- otherwise someone could extract all the "api" functionality into an "apiReducer"


aether-ui/aether/ui/assets/apps/utils/request.jsx
```diff
@@ -0,0 +1,178 @@
+import 'whatwg-fetch'
```
lordmallam, line 1: An implantation of a request middleware which plugs into redux via redux-thunx was already implemented. `https://github.com/eHealthAfrica/aether/blob/aet-ui-setup/aether-ui/aether/ui/assets/apps/redux/middleware/client.jsx#L3` this require you pass redux types on a success and failure together with the url, thus, the resulting response will be stored in the redux store and can be consumed like in the `paginationContainer`. I will suggest that implementation be adopted. I could update tests scripts on that and also to serve as usage examples.

obdulia-losantos: Could you help me with that?

lordmallam: Yup sure... on it 👮

PR #135, opened 2018-04-18
Ui request middleware

lordmallam:
Test and Usage of request api plugged into `redux` via `redux thunk`


aether-ui/aether/ui/assets/apps/utils/api.jsx
```diff
@@ -1,34 +1,53 @@
-import superagent from 'superagent'
-import cookie from 'react-cookie'
-
+/* global fetch, jQuery */
```
obdulia-losantos, line 1: Instead of using the global variables you can do this

```javascript
import 'whatwg-fetch'
import jQuery from 'jquery'
```

With that you don't need to change the test setup ;-)

aether-ui/aether/ui/assets/apps/utils/urls.jsx
```diff
@@ -1,6 +1,6 @@
 const baseUrl = process.env.HOSTNAME
 
 export default {
-  SAMPLE_URL: `${baseUrl}/sample`,
+  MOCK_PIPELINES_URL: 'http://ui:8004/static/mock/pipelines.mock.json',
```
obdulia-losantos: As far as I know `/static/mock/pipelines.mock.json` this should also work

lordmallam: The problem (travis fails) is with the boot sequence.. the static files aren't serve-able during the tests.. working on that...

obdulia-losantos: You can use `nock` for that ;-)

aether-ui/aether/ui/assets/apps/utils/api.jsx
obdulia-losantos: I don't mind to have globals but in that case instead of changing `jest.setup.jsx` the `global` variables should go to `UiTestsEnvironment.jsx` with the rest. :nerd_face:

obdulia-losantos (approved):
Any change in this can be done later.

PR #136, opened 2018-04-19
feat(ui): mapping rules

obdulia-losantos:
the new playground...  not finished yet but you can start checking it.

PR #138, opened 2018-04-19
Ui api consumption

lordmallam:


aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -98,10 +97,10 @@ class Pipeline extends Component {
 
           <div className='pipeline-sections'>
             <div className='pipeline-section__input'>
-              <Input />
+              <Input data={this.props.selectedPipeline && this.props.selectedPipeline.schema} />
```
obdulia-losantos: If you take a look at #136, I made a different approach to this ;-)

obdulia-losantos:
@lordmallam could you merge `aet-ui-setup` again? Looks like I'm reviewing the same code twice :nerd_face:

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
```diff
@@ -25,6 +26,11 @@ export const addPipeline = newPipeline => ({
...
 })
 
+export const getPipelineById = id => ({
+  type: types.GET_BY_ID,
+  payload: id
+})
+
 export const updatePipeline = pipeline => ({
   type: types.PIPELINE_UPDATE,
   payload: pipeline
```
obdulia-losantos: Should this be a promise using `client.put`...? Else, how are the pipeline changes pushed to the database?

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
```diff
@@ -37,7 +43,7 @@ export const selectedPipelineChanged = selectedPipeline => ({
 
 export const getPipelines = () => ({
   types: ['', types.GET_ALL, types.GET_ALL_FAILED],
-  promise: client => client.get(urls.MOCK_PIPELINES_URL, { 'Content-Type': 'application/json' })
+  promise: client => client.get(urls.PIPELINES_URL, { 'Content-Type': 'application/json' })
```
obdulia-losantos: I'll suggest to use the parameter `page_size` with a really BIG number. So far we are not paginating the data but somehow we need to be sure that at least all the data is fetched (as an starting point). :thinking:

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
lordmallam: This is an update to only the redux state for now. I should confirm the UX flow with @sabeen on when the final update is made to the database or if it is done in bits via `add to pipeline` and `apply mapping rules to pipeline` buttons

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -17,7 +17,9 @@ class PipelineList extends Component {
   }
 
   componentWillMount () {
-    this.props.getPipelines()
+    if (!this.props.pipelineList.length) {
```
obdulia-losantos, line 20: what happens here if there are no pipelines yet? eternal loop? :thinking:

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
```diff
@@ -21,13 +22,23 @@ export const INITIAL_PIPELINE = {
...
-  payload: newPipeline
+  types: ['', types.PIPELINE_ADD, types.PIPELINE_ERROR],
+  promise: client => client.post(`${urls.PIPELINES_URL}`,
+    { 'Content-Type': 'application/json' },
+    { data: { name: newPipeline.name } })
+})
+
+export const getPipelineById = id => ({
+  type: types.GET_BY_ID,
+  payload: id
```
obdulia-losantos: not sure if this is a good approach, this means that your local version in the list is always up to date, and I can imagine a couple of situations in which is not true, like having two users alternating with pipelines :thinking:

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
lordmallam: `componentWillMount()` happens only once

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
lordmallam: Ok, i'll get from api

aether-ui/aether/ui/assets/apps/pipeline/sections/Input.jsx
```diff
@@ -44,7 +44,7 @@ class Input extends Component {
       // generate a new input sample
       const input = type.random()
 
-      this.props.updatePipeline({ schema, input })
+      this.props.updatePipeline(Object.assign(this.props.selectedPipeline, { schema, input }))
```
obdulia-losantos: I don't mind using `Object.assign` but if this is going to be open source, sometimes is nice to check the compatibility table.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

`{... this.props.selectedPipeline, schema, input }` does the same work and Babel will take care of compatibility.

obdulia-losantos: Another approach is just indicating the `id` and in the `updatePipeline` use `patch` instead of `put`; in this case it will partially update the given fields, the rest remained the same and the returned value is complete.

```
{
  "id": "ab7091db-b3d0-4281-b39f-20746337febe",
  "url": "http://ui.aether.local:8004/api/ui/pipelines/ab7091db-b3d0-4281-b39f-20746337febe/",
  "created": "...",
  "modified": "...",
  "name": " ... ",
  "schema": { ... },
  "input": { ... },
  "entity_types": [ ... ],
  "mapping": [ ... ],
  "mapping_errors": [ ... ],
  "output": [ ... ]
}
```

As you see, the url for patching, updates or even delete is included in the payload :grimacing:

obdulia-losantos: On the other hand, entity types and mapping were not updated with this change. :wink:

obdulia-losantos:
This line is wrong https://github.com/eHealthAfrica/aether/blob/aet-ui-setup/aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx#L153 :man_facepalming: 

Could you fix that for me? :grimacing: 

It should be `mapping_errors: this.props.selectedPipeline.mapping_errors || []`

obdulia-losantos:
changing the address bar manually does not reload the page with the new pipeline, is this expected?

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
```diff
@@ -36,41 +51,44 @@ export const selectedPipelineChanged = selectedPipeline => ({
...
+const reducer = (state = INITIAL_PIPELINE, action) => {
   const newPipelineList = clone(state.pipelineList)
 
   switch (action.type) {
     case types.PIPELINE_ADD: {
       const newPipeline = clone(action.payload)
       newPipelineList.unshift(newPipeline)
-
-      return { ...state, pipelineList: newPipelineList, selectedPipeline: newPipeline }
+      return { ...state, pipelineList: newPipelineList, selectedPipeline: newPipeline, error: null }
```
obdulia-losantos, line 65: what about `notFound` entry? shouldn't it be `null`?

aether-ui/aether/ui/assets/apps/utils/api.jsx
```diff
@@ -37,15 +38,23 @@ export default class ApiClient {
...
-            } else {
-              return { error: res.statusText, status: res.status }
-            }
-          })
-          .catch(err => ({ error: err.message, status: err.errno }))
+        return new Promise((resolve, reject) => {
+          window.fetch(path, options)
+            .then(res => {
+              if (res.status >= 200 && res.status < 400) {
+                res.json().then(res => resolve(res)).catch(err => reject(err)) // Should be extended to cater for other content-types or than json
```
obdulia-losantos, line 45: hopefully we'll only work with json content types :crossed_fingers:

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
lordmallam: Well... `notFound` isn't used during an add operation. `error` used for operation that make changes to the `db`

aether-ui/aether/ui/assets/apps/pipeline/sections/EntityTypes.jsx
```diff
@@ -44,7 +44,7 @@ class EntityTypes extends Component {
       // generate sample output with new enity types (TO BE REMOVED!!!)
       const output = schemas.map(et => avro.parse(et).random())
 
-      this.props.updatePipeline({ entity_types: schemas, output })
+      this.props.updatePipeline({ ...this.props.selectedPipeline, entity_types: schemas, output })
```
obdulia-losantos: remove output here, it was only necessary with mock data, the validation step will provide it.

aether-ui/aether/ui/assets/apps/pipeline/sections/EntityTypes.jsx
```diff
@@ -41,9 +41,6 @@ class EntityTypes extends Component {
     try {
       // validate schemas
       const schemas = JSON.parse(this.state.entityTypesSchema)
-      // generate sample output with new enity types (TO BE REMOVED!!!)
-      const output = schemas.map(et => avro.parse(et).random())
```
obdulia-losantos, line 44: The replacement can be `schemas.forEach(et => { avro.parse(et) })` to check that each entity type is a valid AVRO schema.

obdulia-losantos (approved):
:man_dancing:

PR #140, opened 2018-04-25
Ui mapping styles

sabeen:
In this branch I finalized the mapping styles.
Plus I added a first version of a styleguide that can be reached via
http://ui.aether.local:8004/styleguide/


obdulia-losantos (approved):
looks awesome!!! :clap:

PR #141, opened 2018-04-25
feat: local development with minikube

quesebifurcan:
Setup travis to run kernel and odk tests with minikube, helm and tiller.

Create local secrets (development only).

Modify nginx config maps to enable usage of internal kubernetes DNS.

Create override files for environments "local" and "test".

Modify `livenessProbe.initialDelaySeconds` and `readinessProbe.initialDelaySeconds` to speed up local development.

**NOTE** in this PR, kernel- and odk-test are running inside a kubernetes cluster. All tests (including kernel, odk, couchb-sync, client, producer etc) are _also_ expected to run to completion using our standard docker-compose setup. 

The duration of the kubernetes-based test suite is ca. 3 minutes, including installation and startup of minikube, helm, tiller.

### Dependencies
- minikube ([installation instructions](https://kubernetes.io/docs/tasks/tools/install-minikube/))
- helm ([installation instructions](https://github.com/kubernetes/helm/blob/master/docs/install.md))

### Installation
If you are using ubuntu, you can probably make use of the installation process we use for Travis CI.

To install minikube on ubuntu, run:
```
sudo ./scripts/kubernetes/install_minikube.sh
```
To install helm on ubuntu, run:
```
sudo ./scripts/kubernetes/install_helm.sh --version v2.8.1
```

### Environments
To start the `kernel` and `odk` modules in a local minikube cluster with code reloading enabled, we need to mount our aether repository in minikube:

```
minikube mount $(realpath .):/aether
```
This process needs to keep running in order for our mount to work, so start it in a separate terminal tab or background it in your current one.

Once the mount process is running, we can do:
```
./scripts/kubernetes/install_secrets.sh && ./scripts/kubernetes/start_cluster.sh ./helm/overrides/local
```
This will bring up both applications with auto-reloading of django code enabled. 

### Accessing the aether APIs from the host

The ingress resources in the aether modules use hostnames to direct incoming traffic. In order to be able to access the aether APIs from a browser or a command line client on the host machine, we need to add a couple of entries to our `/etc/hosts` file.

To get a list of the exposed endpoints, run:
```
kubectl get ingress
```
The result should look similar to this:
```
NAME      HOSTS                 ADDRESS          PORTS     AGE
kernel    kernel.aether.local   192.168.99.100   80, 443   28m
odk       odk.aether.local      192.168.99.100   80, 443   28m
```
For each entry, add the address and the hostname to your `/etc/hosts` file:
```
# /etc/hosts
...
192.168.99.100 kernel.aether.local
192.168.99.100 odk.aether.local
...
```

## Running the tests
If you already have kernel and odk running, do:
```
./scripts/kubernetes/run_tests.sh test_all
```
To test a single module, use `./run_tests.sh test_<module-name>`. For example:
```
./scripts/kubernetes/run_tests.sh test_kernel
./scripts/kubernetes/run_tests.sh test_odk
```
To delete everything in the cluster, bring it back up and then run the tests, do:
```
./scripts/kubernetes/delete_all.sh && \
    ./scripts/kubernetes/install_secrets.sh && \
    ./scripts/kubernetes/start_cluster.sh ./helm/overrides/local && \
    ./scripts/kubernetes/run_tests.sh test_all
```


scripts/kubernetes/install_helm.sh
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/env bash
```
quesebifurcan, line 1: (This script is copied from https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get.)

helm/kernel/templates/deployment.yaml
```diff
@@ -57,13 +57,13 @@ spec:
           mountPath: {{ .Values.application.data_mount }}
       containers:
       - name: {{ .Chart.Name }}
-        image: "{{ .Values.kernel.repository }}:{{ .Chart.Version }}"
+        image: "{{ .Values.kernel.repository }}:{{ .Values.release }}"
```
quesebifurcan, line 60: @MrPink is this ok with you? Background: for local development, we need to be able to reference e.g. an "aether-kernel:latest" release. It's nice to have `.Chart.Version` as tag though, so If you can think of a better way, I'm all ears.

helm/kernel/values.yaml
```diff
@@ -31,9 +34,7 @@ provider:
...
   enabled: true
+  hosts:
+    - kernel.aether.local
   annotations:
-    alb.ingress.kubernetes.io/scheme: "internet-facing"
-    alb.ingress.kubernetes.io/healthcheck-path: /health
-    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP":80,"HTTPS": 443}]'
-    alb.ingress.kubernetes.io/certificate-arn: "arn:aws:acm:eu-west-1:387526361725:certificate/dcd71c49-1b5b-4b19-bd2d-6de98a3f6477"
-    alb.ingress.kubernetes.io/security-groups: "alb-ingress-controller.dev.k8s.eha.to"
+    nginx.ingress.kubernetes.io/ingress.class: "nginx"
```
quesebifurcan, line 40: @MrPink moved the aws specific ingress config out of the way to keep the default config. So those values would need to be passed in explicitly during prod deploy.

quesebifurcan: (An alternative solution would be to keep the `alb.ingress.kubernetes.io` values in this file and override them in local config files.)

helm/kernel/templates/deployment.yaml
MrPink: I guess it makes sense to have a `release` value, I just didn't want to have two values to update

helm/kernel/values.yaml
MrPink: I think moving it out of the way is the cleanest approach, the override specific YAML's should be specific to what platform its deploying too.

MrPink (approved):
Superb work @quesebifurcan  🥇

PR #145, opened 2018-04-30
chore: switch avro library from avro-js to avsc

quesebifurcan:
Switch from [avro-js](http://svn.apache.org/repos/asf/avro/trunk/) to [avsc](https://github.com/mtth/avsc). `avsc` can derive schemas from arbitrary json data, which is needed in https://jira.ehealthafrica.org/browse/AET-234. 

Example:
```node
> var avro = require('avsc')
undefined
> var data = {a: 1, b: {c: {d: [2, "a", "b"]}}}
undefined
> avro.Type.forValue(data)
<RecordType {"fields":[{"name":"a","type":"int"},{"name":"b","type":{"type":"record","fields":[{"name":"c","type":{"type":"record","fields":[{"name":"d","type":{"type":"array","items":["int","string"]}}]}}]}}]}>
```

See https://github.com/mtth/avsc/issues/108#issuecomment-302436388 for info on how to use type hooks to deal with anonymous records and enums.

### DIFF
--- a/aether-ui/aether/ui/assets/apps/components/AvroSchemaViewer.jsx
+++ b/aether-ui/aether/ui/assets/apps/components/AvroSchemaViewer.jsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react'
 import { FormattedMessage } from 'react-intl'
-import avro from 'avro-js'
+import avro from 'avsc'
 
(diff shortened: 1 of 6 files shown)


admbtlr (approved):
Well that was easy :)

PR #147, opened 2018-05-02
feat: schema/data toggle

quesebifurcan:


aether-ui/aether/ui/assets/apps/components/NavBar.jsx
```diff
@@ -34,7 +34,7 @@ class NavBar extends Component {
         }
 
         <div data-qa='navbar-user' className='top-nav-user'>
-          <span class='user-name'>
+          <span className='user-name'>
```
obdulia-losantos, line 37: good catch!

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -58,7 +58,7 @@ class Pipeline extends Component {
     return (
       <div className={'pipelines-container show-pipeline'}>
         <NavBar showBreadcrumb>
-          <div classname='breadcrumb-links'>
+          <div className='breadcrumb-links'>
```
obdulia-losantos, line 61: good catch! again!

obdulia-losantos:
LGTM!

One concern I have is that if the user updates the input data without changing the schema but just given meaningful values to the properties this makes the AVRO schema to be regenerated and the other way around... is there a way to detect those cases?

lordmallam (approved):
Looks good

quesebifurcan:
@obdulia-losantos good point -- me, Adam and Sabeen discussed this offline on Wednesday. It will be addressed in a future PR. The idea was that, when editing an Avro schema generated from a data sample, a gentle reminder will appear above the text area, notifying the user that committing these changes will result in in the sample data being generated anew (if no longer compatible with the edited schema).

PR #150, opened 2018-05-03
feat(odk): include original xml as an attachment

obdulia-losantos:
This is the simplest and quickest way I found. :nerd_face:


quesebifurcan (approved):
Looks good to me :+1:

PR #151, opened 2018-05-04
Aet ui push to kernel

lordmallam:
#### What does this PR do?
Implements the pipeline publish api endpoint

- Transforms pipeline data into kernel objects (project, schema, mapping)
- Checks if the pipeline objects have links to existing kernel objects (if yes, to be handled in an upcoming PR; if no, creates a new kernel object and save reference to the object on the UI pipeline)

#### How to test
Make a `GET` request `/ui/publish/<pipeline_id>/<project_name>` `pipeline_id` : Id of a pipeline with valid mappings
On kernel, make a submission with the mapping created, Entities should be generated as expected

#### Related JIRA Tickets
[AET-220](https://jira.ehealthafrica.org/browse/AET-220)
[AET-222](https://jira.ehealthafrica.org/browse/AET-222)
[AET-223](https://jira.ehealthafrica.org/browse/AET-223)


aether-ui/aether/ui/api/tests/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+                        'rdf_definition': '[]'
+                      }
+        entity_type = PIPELINE_EXAMPLE_1['entity_types'][2]
+        schema_data = {
+                        'revision': '123',
+                        'name': entity_type['name'],
+                        'type': entity_type['type'],
+                        'definition': entity_type
+                      }
+        try:
```
obdulia-losantos: is it a good practice to use `try-catch` within a test? just asking :thinking:

lordmallam: hehehehe.. nope. That was a test within a test. Ill take that out

quesebifurcan:
Are there any alternatives to using a `GET` request to publish the pipeline? Just asking, since `GET`s do not usually have any side effects.

aether-ui/aether/ui/api/views.py
```diff
@@ -144,3 +146,63 @@ def valid_header(name):
                                     *args,
                                     **kwargs)
         return HttpResponse(response, status=response.status_code)
+
+
+def PublishPipeline(requests, pipelineid, projectname):
```
obdulia-losantos: Could it be possible instead of using an external method outside the ViewSet to include it inside?

https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

```python
from django.shortcuts import get_object_or_404
from rest_framework.decorators import action


    queryset = models.Pipeline.objects.all()
    serializer_class = serializers.PipelineSerializer
    ordering = ('name',)

        # ....
        pipeline = get_object_or_404(models.Pipeline, pk=pk)
        project_name = request.data['project_name']  # project name is part of the submission, not the url

```

The advantage of this approach is that the rest framework library takes care of the security and the permissions are inherited (so far we don't have any special concern here but who knows later), also you don't need to define a new entry in the `urlpatterns` list. The default url would be `/ui/pipelines/{id}/publish`.

aether-ui/aether/ui/api/tests/__init__.py
```diff
@@ -0,0 +1,844 @@
...
+    'project': '12345',
+    'schema': {
+      'Person': '1234'
+    },
+    'projectSchema': {
+      'Person': '1234'
+    },
+    'mapping': '1234'
+  }
+}
```
quesebifurcan, line 376: This looks good, but could we reduce the size of this example a bit? It's a lot to read and scroll through when debugging. Simple example schema:

```
{
  "name": "Example",
  "type": "record",
  "fields": [
    {
      "name": "a",
      "type": "string"
    }
  ]
}
```

aether-ui/aether/ui/api/views.py
lordmallam: With the current implementation any method would suffice `GET` or `POST`, as it is simply to trigger the requests to the kernel. But i see sense in @obdulia-losantos's approach so i'll refactor to use a post

aether-ui/aether/ui/api/tests/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class ViewsTest(TestCase):
+
+    def setUp(self):
+        username = 'test'
+        email = 'test@example.com'
+        password = 'testtest'
+        self.user = get_user_model().objects.create_user(username, email, password)
+        self.client.login(username=username, password=password)
```
obdulia-losantos: I removed this line and the tests passed!!! This view is skipping authentication!!! :scream:

aether-ui/aether/ui/api/views.py
obdulia-losantos: Regarding "safe" methods... 
http://www.django-rest-framework.org/api-guide/permissions/#api-reference

So far we only required authenticated users

https://github.com/eHealthAfrica/aether/blob/develop/aether-common-module/aether/common/conf/settings.py#L128

aether-ui/aether/ui/api/views.py
```diff
@@ -144,3 +146,63 @@ def valid_header(name):
...
+        if ui_utils.is_object_linked(pipeline.kernel_refs, 'mapping'):
+            # Notify user of existing object, and confirm override
+            pass
+        else:
+            ui_utils.create_new_kernel_object('mapping', pipeline, mapping_data, projectname)
+
+        return JsonResponse(serializers.PipelineSerializer(pipeline, context={'request': requests}).data,
+                            status=200, safe=False)
+    except Exception as e:
+        return JsonResponse({'message': str(e), 'error': 'Bad request'}, status=400)
```
quesebifurcan: What happens if some of these calls fail but others succeed? My impression is that we will then end up with an invalid representation of the pipeline in the kernel database.

An alternative strategy would be to add a new (`POST`) view to kernel, which accepts a JSON doc representing an entire project (including schemas, projectschemas, mappings etc). This view can then be wrapped in `@transaction.atomic` to make sure that kernel is always consistent. See: https://docs.djangoproject.com/en/2.0/topics/db/transactions/#controlling-transactions-explicitly

quesebifurcan (changes requested):

aether-ui/aether/ui/api/views.py
quesebifurcan: Example post data:
```
{
  "project_name": "test",
  "schemas": [...],
  "mappings": [...],
   ...etc.
}
```
Would be great to validate this with a schema while we're at it.

lordmallam: The eventual result is to get the mappings saved to the kernel, which depends on the `project, schemas, and project_schemas` to be setup correctly. During the publish process a number of items can succeed while others fail. Handling of this scenario was skipped in this PR due to lack of clarity as to how the `ui` will handle the possible outcomes. An update with the logic to report happenings on the kernel will be pushed shortly

lordmallam: Also, the aim is to leave the `kernel` as is and all logic concerning `ui` (pipelines) should be on the ui module

quesebifurcan: That makes sense :+1:  We can discuss this further in the channel and/or standup. My main point was that handling these errors is much easier and less error prone if we do it in the kernel, so it would make sense for us to add a new endpoint for submitting a project in one swoop.

aether-ui/aether/ui/api/tests/test_utils.py
obdulia-losantos: Could you create a test without "login"? Just to check that anonymous users cannot break anything :crossed_fingers:

obdulia-losantos: Actually, login is not needed here but in view tests :grimacing:

aether-ui/aether/ui/api/utils.py
```diff
@@ -114,3 +119,156 @@ def validate_pipeline(pipeline):
...
+                            'is_encrypted': False,
+                            'project': pipeline.kernel_refs['project'],
+                            'schema': schema_id
+                        }
+    if not is_object_linked(pipeline.kernel_refs, 'projectSchema', entity_name):
+        create_new_kernel_object('projectSchema',
+                                 pipeline, project_schema_data, entity_name=entity_name)
+
+
+def publishPipeline(requests, pipelineid, projectname):
```
obdulia-losantos: The `requests` argument is not used at all and creates a little bit of confusion, could you take it out?

aether-ui/aether/ui/api/views.py
```diff
@@ -3,16 +3,27 @@
...
     ordering = ('name',)
 
+    @action(methods=['post'], detail=True)
+    def publish(self, request, pk=None):
+        '''
+        This view transform the supplied pipeline to kernal models,
+        publish and update the pipeline with related kernel model ids.
+        '''
+        project_name = request.data['project_name'] if 'project_name' in request.data else 'Aux'
+        return ui_utils.publishPipeline(request, pk, project_name)
```
obdulia-losantos: I would suggest to pass as an argument the `pipeline`, not the `pk`.

There is an specific method in Django that will throw a 404 error if the object does not exist. I think this is the expected error when you try to publish a non-existent pipeline :thinking: 

```python
from django.shortcuts import get_object_or_404

def .... (..., pk):
    pipeline = get_object_or_404(models.Pipeline, pk=pk)
    # ...
    return ui_utils.publishPipeline(pipeline, project_name)

```

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -6,7 +6,7 @@ import { PROJECT_NAME } from '../utils/constants'
 import { NavBar } from '../components'
 
 import NewPipeline from './NewPipeline'
-import { addPipeline, selectedPipelineChanged, getPipelines } from './redux'
+import { addPipeline, selectedPipelineChanged, getPipelines, publishPipeline } from './redux'
```
obdulia-losantos: Is the `publishPipeline` used somewhere in this file? or are you going to include the button later?

aether-ui/aether/ui/assets/apps/pipeline/redux.spec.jsx
```diff
@@ -172,4 +174,44 @@ describe('Pipeline actions', () => {
           })
       })
   })
+
+  it('should dispatch an publish pipeline action and save respose in the redux store', () => {
```
obdulia-losantos: typo `respose` :wink:

obdulia-losantos: `dispatch AN publish` or `dispatch A publish` ?

aether-ui/aether/ui/assets/apps/utils/api.jsx
```diff
@@ -47,7 +47,9 @@ export default class ApiClient {
                 if (res.status === 404) {
                   reject(new NotFoundError())
                 } else {
-                  resolve({ error: res.statusText, status: res.status })
+                  res.json().then(error => {
+                    resolve({ error, status: res.status })
```
obdulia-losantos, line 51: `resolve` or `reject` ? :thinking: :flushed:

aether-ui/aether/ui/assets/apps/pipeline/redux.spec.jsx
```diff
@@ -172,4 +174,44 @@ describe('Pipeline actions', () => {
...
+        expect(store.getState().publishSuccess).toEqual(
+          expected
+        )
+        expect(store.getState().publishError).toEqual(
+          null
+        )
+      })
+  })
+
+  it('should dispatch a wrong publish pipeline action and save respose in the redux store', () => {
```
obdulia-losantos: `respose`

aether-ui/aether/ui/api/views.py
lordmallam: There are a number of expected errors while publishing a pipeline, an invalid pipeline id being one. The `publishPipeline` utils function https://github.com/eHealthAfrica/aether/pull/151/files#diff-9ed92cc050ab97d135d3e96acaec24a6R195, I am just trying not to make extra branches on the view.

aether-ui/aether/ui/api/utils.py
```diff
@@ -114,3 +119,156 @@ def validate_pipeline(pipeline):
...
+    outcome = {
+        'successful': [],
+        'failed': [],
+        'exists': []
+    }
+    try:
+        pipeline = models.Pipeline.objects.get(pk=pipelineid)
+        if pipeline.mapping_errors:
+            outcome['failed'].append({'message': 'Mappings have errors', 'status': 'Bad Request'})
+            return JsonResponse({'links': None, 'info': outcome}, status=400)
```
obdulia-losantos, line 208: I wonder if it's better to return the `outcome` object in this method and let the `publish` viewset method to deal with the response and the statuses codes... but it's up to you.

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
lordmallam: Not yet... a publish is to be possible from the list as well, i'll take it out pending when the mockups are ready

obdulia-losantos: You can leave it there if you want, sooner or later is going to be there...

aether-ui/aether/ui/api/views.py
obdulia-losantos: I can imagine that the list of unexpected errors is longer :joy:

aether-ui/aether/ui/assets/apps/utils/api.jsx
lordmallam: Actually `resolve` ref: https://github.com/eHealthAfrica/aether/blob/aet-ui-push-to-kernel/aether-ui/aether/ui/assets/apps/redux/middleware/client.jsx#L16, an error on `res.json()` will then result in a `reject`

aether-ui/aether/ui/api/utils.py
lordmallam: Not sure i understand that approach as to how to determine what the status code is on the viewset. The `outcome` object is built at different stages with different status per action performed on the kernel

obdulia-losantos: Let me explain better but as I said before do whatever you want...

```python
# utils.py

def publishPipeline(pipelineid, projectname):
    '''
    Transform pipeline to kernel data and publish
    '''
    outcome = {
        'successful': [],
        'failed': [],
        'exists': []
    }

    # do magic....
    if error_somewhere:
        outcome['failed'].append('some reason')
        return outcome  # instead of JsonResponse

    return outcome

# ------------------------------------------------

# views.py


    queryset = models.Pipeline.objects.all()
    serializer_class = serializers.PipelineSerializer
    ordering = ('name',)

        '''
        This view transform the supplied pipeline to kernal models,
        '''
        outcome = ui_utils.publishPipeline(pk, project_name)

        if not len(outcome['failed']) and not len(outcome['exists']):
            # somewhere pipeline = ...
            return Response(data=serializers.PipelineSerializer(pipeline, context={'request': request}).data)
        else:
            return Response(data=outcome, status=400)

```

aether-ui/aether/ui/api/utils.py
```diff
@@ -192,7 +192,7 @@ def create_project_schema_object(name, pipeline, schema_id, entity_name):
                                  pipeline, project_schema_data, entity_name=entity_name)
 
 
-def publishPipeline(requests, pipelineid, projectname):
+def publishPipeline(pipelineid, projectname):
```
obdulia-losantos: :D thanks!

aether-ui/aether/ui/assets/apps/utils/api.jsx
obdulia-losantos: What I meant is that the `resolve` is already done and here you are dealing with >= 400 status codes, so it should be `reject` because is an error in the server response.

aether-ui/aether/ui/api/utils.py
lordmallam: I see.. Just for clarity, i am going with this method because from [L247](https://github.com/eHealthAfrica/aether/blob/aet-ui-push-to-kernel/aether-ui/aether/ui/api/utils.py#L247) to [L270](https://github.com/eHealthAfrica/aether/blob/aet-ui-push-to-kernel/aether-ui/aether/ui/api/utils.py#L270) `exists` or `failed` in the outcome object does not necessarily mean the request failed. I think making a selection when the required info is available to make that decision, should be made rather than deffering, especially in this case where a request is made up of multiple requests.

obdulia-losantos: I took that line to make the difference between fail and success as I could have taken another one; but what it's better in your opinion it's going to be great.

aether-ui/aether/ui/assets/apps/utils/api.jsx
lordmallam: Rejections have to be of `Error` Type, since this plugs into `redux` i need an object in the format it is to be sent as the failure payload. Note: a failure may come along with an object detailing the exact reason of failure on the server like in the `outcome` case, i need the user to have all that juicy info.

lordmallam: All requests go through the [middleware](https://github.com/eHealthAfrica/aether/blob/aet-ui-push-to-kernel/aether-ui/aether/ui/assets/apps/redux/middleware/client.jsx) where this concern is addressed

aether-ui/aether/ui/api/models.py
```diff
@@ -194,6 +194,7 @@ class Pipeline(TimeStampedModel):
     # }
     mapping_errors = JSONField(blank=True, null=True, editable=False)
     output = JSONField(blank=True, null=True, editable=False)
+    kernel_refs = JSONField(blank=True, null=True, editable=False)
```
quesebifurcan, line 197: Is there any validation of the data which can go into this field? Since other parts of the program depend on it conforming to a certain structure, it would be great to have something there.

aether-ui/aether/ui/api/tests/test_utils.py
```diff
@@ -0,0 +1,56 @@
+import ast
+
+from django.test import TestCase
+
+from . import PIPELINE_EXAMPLE_1
+from ..models import Pipeline
+from ..utils import (kernel_data_request, create_new_kernel_object, is_object_linked,
+                     create_project_schema_object)
```
quesebifurcan: An alternative to these long tuples would be to `from .. import utils` and then call the functions like `utils.kernel_data_request()`. Not necessary to fix right now, just mentioning it.

aether-ui/aether/ui/api/tests/test_utils.py
```diff
@@ -0,0 +1,56 @@
...
+class ViewsTest(TestCase):
+    def test_kernel_data_request(self):
+        result = kernel_data_request('projects', 'get')
+        self.assertIn('count', result)
+        with self.assertRaises(Exception):
+            kernel_data_request('projects', 'post', {'wrong-input': 'tests'})
+
+    def test_create_new_kernel_object(self):
+        pipeline = Pipeline.objects.create(
+          name=PIPELINE_EXAMPLE_1['name'],
```
quesebifurcan: You might want to check the indentation here (should be 4 spaces). I'm surprised the linter doesn't complain :smile:

aether-ui/aether/ui/api/tests/__init__.py
```diff
@@ -0,0 +1,492 @@
...
+      'id': '0d8aebf3-50d0-4e77-a5ee-1045ffa5f29f',
+      'firstName': 'Sejflsd Dljljkf'
+    }
+  ],
+  'input': {
+    'person': {
+      'forename': 'Sejflsd Dljljkf'
+    }
+  }
+}
```
quesebifurcan, line 437: Alright, 500 lines is less than 800 lines, but it would be wonderful to have shorter test data. We might need to add to/alter the schemas sooner rather than later.

aether-ui/aether/ui/api/utils.py
```diff
@@ -114,3 +119,156 @@ def validate_pipeline(pipeline):
...
+                            'is_encrypted': False,
+                            'project': pipeline.kernel_refs['project'],
+                            'schema': schema_id
+                        }
+    if not is_object_linked(pipeline.kernel_refs, 'projectSchema', entity_name):
+        create_new_kernel_object('projectSchema',
+                                 pipeline, project_schema_data, entity_name=entity_name)
+
+
+def publishPipeline(pipelineid, projectname):
```
quesebifurcan: Minor thing -- all functions should use snake case: `publish_pipeline`

quesebifurcan (changes requested):

aether-ui/aether/ui/api/models.py
lordmallam: There is no validation. The field is used to store links to kernel objects, if a ref to a certain object isn't in the dict it is not linked. keys to an object are only added once the object has been created on the kernel

quesebifurcan (approved):
:+1: Nice work!

We can merge this now, but I think it's important to keep track of and eventually address a couple of issues:
- validation of `kernel_refs`
- interaction with kernel -- it's currently more complicated than it has to be. I would suggest extending the kernel API with views which are helpful for building the UI (among other things, that would allow us to use transactions when registering pipeline objects).

PR #154, opened 2018-05-08
chore: use virtualenv with pip freeze

obdulia-losantos:
pip freeze wasn't working as expected (it was taking removed dependencies), use `virtualenv` to skip this issue.


quesebifurcan (approved):
:+1: Nice work

PR #155, opened 2018-05-08
feat: filter entities by project

quesebifurcan:
Filter entities by project id.

Example:
```
curl http://kernel.aether.local:8000/entities/?project=<project-id>
```


admbtlr (dismissed):
Thank you!

aether-kernel/aether/kernel/api/filters.py
```diff
@@ -63,6 +63,11 @@ class Meta:
 
 
 class EntityFilter(filters.FilterSet):
+    project = filters.CharFilter(
```
obdulia-losantos, line 66: Once upon a time, in a far far far land there were a sort of files called "tests". They look useless for the end user but somehow they gracefully helped the humble developer to find errors and to check that what the code lines said was what the code lines did...

Joking apart, weren't in the `test_views` file any sort of filtering tests? :thinking: 

I found them in the former gather core but not here :thinking: 
https://github.com/eHealthAfrica/gather2/blob/develop/gather2-core/gather2/core/api/tests/test_views.py#L119

I know that there are very simple but they help to understand and to realize what to expect from the code (as an another developer, whose name I cannot remember (something with FR...), told me two lifetimes ago) :stuck_out_tongue:

obdulia-losantos (dismissed):

shawnsarwar (dismissed):
Generally useful thing to be able to do.

aether-kernel/aether/kernel/api/filters.py
quesebifurcan: Good point, will add a couple of tests :+1:

lordmallam (dismissed):

aether-kernel/conf/pip/primary-requirements.txt
```diff
@@ -9,6 +9,7 @@
 ################################################################################
 
 django<2
+django-autofixture
```
obdulia-losantos: Because this library is only used in tests, could you move it to the bottom with the rest of test libraries :wink:

obdulia-losantos:
*And they lived happily ever after.*

This kind of things happens when I speak too much so  :speak_no_evil:

aether-kernel/aether/kernel/api/tests/test_filters.py
```diff
@@ -0,0 +1,85 @@
...
+            kwargs = {'instanceID': instance_id}
+            url = reverse(viewname='submission-list')
+            response = json.loads(
+                self.client.get(url, kwargs, format='json').content
+            )
+            self.assertEqual(response['count'], 1)
+            self.assertEqual(
+                response['results'][0]['payload']['meta']['instanceID'],
+                instance_id,
+            )
```
obdulia-losantos, line 85: :clap: :clap: :clap: :clap: :clap:

obdulia-losantos (dismissed):
:speak_no_evil:

aether-kernel/aether/kernel/api/tests/test_filters.py
quesebifurcan: yeah, thanks for pointing this out :+1:  
The filter functionality is trivial to implement, but very hard to keep track of over time.

obdulia-losantos (approved):
:bridge_at_night:

PR #156, opened 2018-05-08
Introduce DB health check URL

tooxie:
It checks:
* DB connection
* Kernel connection


aether-common-module/aether/common/health/views.py
```diff
@@ -18,10 +18,26 @@
...
 
     return JsonResponse({})
+
+
+def deep_health(*args, **kwargs):
+    '''
+    Deep health check. Tests both the kernel and the database connections.
+    '''
+
+    kernel_reachable = test_connection()
```
obdulia-losantos: Does this line make sense within the kernel app? :stuck_out_tongue:

tooxie: You are right, but since it's a common file for all other modules as well I ignored that detail 😅 I could wrap it in a condition so it runs only if not in the kernel already:
```py
in_kernel = apps.get_app_config('kernel').name == 'aether.kernel'
if not in_kernel:
    # magic happens here
```

What do you think?

aether-common-module/aether/common/kernel/utils.py
```diff
@@ -79,6 +82,16 @@ def test_connection():
     return False  # it's not possible to connect with kernel :(
 
 
+def test_db_connection():  # pragma: no cover
```
obdulia-losantos: This is not `kernel` related, it fits better in the `health` module, doesn't it? :thinking: 

You are not checking the kernel database connection but yours... if I'm not wrong.

tooxie: True, it picks up the default DB and checks that it's reachable, it's not kernel specific. Will move to the `health` module.

aether-common-module/aether/common/health/views.py
obdulia-losantos: That will work if the kernel app is not installed, and I cannot imagine a case were it would happen outside the own app. :+1: 

What I would suggest it to include the "kernel check" as an argument in the view (you can pass it automatically as an extra option in the url definition).
https://docs.djangoproject.com/en/1.11/topics/http/urls/#passing-extra-options-to-view-functions

It might be the case, but very unlikely, that your app is not kernel but you don't mind if you are not connected to kernel... then your deep-health will always return false failures.

But do as you wish, I'm just :thinking: :thought_balloon: :thought_balloon: :thought_balloon:

tooxie: I see. Then what about getting rid of the kernel check (which is already possible under the `/check-kernel` endpoint) and have a common `/check-db` for all modules instead?

obdulia-losantos: Sounds good!

obdulia-losantos (approved):

quesebifurcan (approved):

PR #159, opened 2018-05-09
feature: added jsonpath wildcards to extractor

shawnsarwar:
Made improvements to kernel utils.find_json_path.

   From the docs
    '''
    This function wraps `jsonpath_ng.parse()` and
    `jsonpath_ng.jsonpath.Child.find()` in order to provide custom
    functionality as described in https://jira.ehealthafrica.org/browse/AET-38.

    If the any single element in `path` is a wildcard match prefixed by an arbitrary
    string, `find` will attempt to filter the results by that prefix. We then replace
    that section of the path with a singel wildcard which becomes a valid jsonpath.
    We filter matches on their adherence to the partial path.

    **NOTE**: this behavior is not part of jsonpath spec.
    '''
We can now handle wildcards at any depth.

Given:
`            'data': {
                'dose-1': {
                    'id': 1,
                },
                'dose-2': {
                    'id': 2,
                },
                'person-1': {
                    'id': 3,
                },
            }
        }
`
We can describe dose-1.id and dose2.id by: '$.data.dose-*.id'


quesebifurcan:
`jsonpath_ng.parse` is used in other places, it seems. `mapping_validation.py` is one, but there are also several examples of `jsonpath_ng.parse` in kernel utils. Did you go through them and check which of them should be using the custom jsonpath implementation? @shawnsarwar

PR #162, opened 2018-05-09
feat(odk): one-item list and altitude+accuracy in geopoint (#148)

admbtlr:
* chore(odk): remove geojson dependency

* feat(odk): include altitude+accuracy in geopoint

* test(odk): include expected parsed data

* fix(odk): detect list of one element

* test(odk): does not break with unexpected data

* fix(odk): check that one-item list are not list


obdulia-losantos:
:see_no_evil: :speak_no_evil: :hear_no_evil:

PR #167, opened 2018-05-15
Aet ui load mappings from kernel

lordmallam:
Fetch mapping data from the `kernel` and transform and store as `ui pipelines`
NOTE: 

- Pipeline input data and schema defaults to `{}`
- Subsequent mapping updates from the kernel are not effected to prevent overwriting changes on the ui.

#### How to test

- Setup kernel with project, schema, project schema and mappings
- Load ui
- A pipeline should be added to the pipeline list reflecting the mappings on kernel


quesebifurcan:
I'm currently not able to update kernel mappings or schemas. After

1. creating a pipeline
2. publishing the pipeline
3. editing the mapping and entity types
4. publishing the pipeline (again)

...kernel is still in a state which corresponds to step 2. Any ideas? @lordmallam

lordmallam:
@quesebifurcan as stated in this [PR](https://github.com/eHealthAfrica/aether/pull/151), updates to kernel objects created from a pipeline by republishing is to be determined by whether the user wants to overwrite existing kernel objects (schemas and mappings) or create new ones with new names. Thus, the dependence on [AET-254](https://jira.ehealthafrica.org/browse/AET-254)

aether-ui/aether/ui/api/utils.py
```diff
@@ -250,7 +250,7 @@ def publish_pipeline(pipelineid, projectname):
                     for rule in pipeline.mapping
                 ]
                 mapping_data = {
-                    'name': '{}-{}'.format(projectname, pipeline.name),
+                    'name': pipeline.name,
```
tooxie: Isn't it risky to change this? Couldn't we end up with 2 different naming schemas in the same DB?

lordmallam: As it is not in production yet, I don't think it is a concern. BTW, the change was made to match the pipeline naming on the ui to that of the kernel mapping being that the end product of a pipeline is a mapping

aether-ui/aether/ui/api/utils.py
```diff
@@ -272,3 +272,53 @@ def publish_pipeline(pipelineid, projectname):
...
+    }
+    linked_pipeline = models.Pipeline.objects.filter(**kwargs)
+    return True if len(linked_pipeline) else False
+
+
+def convertMappings(mapping_from_kernel):
+    result = []
+    for mapping in mapping_from_kernel:
+        result.append({'source': mapping[0], 'destination': mapping[1]})
+    return result
```
obdulia-losantos: The pythonic way of life...

```python
    return [
        {'source': mapping[0], 'destination': mapping[1]}
        for mapping in mapping_from_kernel
    ]
```

aether-ui/aether/ui/api/views.py
```diff
@@ -15,6 +14,16 @@ class PipelineViewSet(viewsets.ModelViewSet):
...
 
+    @action(methods=['get'], detail=False)
+    def fetch(self, request):
+        '''
+        This view gets kernel objects, transforms and loads into a pipeline
+        '''
+        ui_utils.kernel_to_pipeline()
+        pipelines = models.Pipeline.objects.all()
+        serialized_data = serializers.PipelineSerializer(pipelines, context={'request': request}, many=True).data
+        return JsonResponse(serialized_data, status=200, safe=False)
```
obdulia-losantos: You don't need to use `JsonResponse`, rest framework is going to transform it into JSON because the default parser is the JsonParser.

On the other hand this should not be a `GET` method because you are adding data from Kernel, and any method that creates/transforms/deletes data is not safe at all. I think it's better `POST`.

Safe methods are `HEAD`, `OPTIONS` and `GET`.

aether-kernel/aether/kernel/api/views.py
```diff
@@ -65,12 +65,34 @@ class ProjectViewSet(CustomViewSet):
...
+    def byname(self, request):
+        '''
+        This view returns projects filtered by the passed name
+        '''
+        name = self.request.query_params.get('name', None)
+        filtered_list = utils.find_by_name('Project', name)
+        serialized_data = serializers.ProjectSerializer(
+                           filtered_list, context={'request': request}, many=True).data
+        return Response(serialized_data, status=HTTPStatus.OK)
+
```
quesebifurcan: We might have what you're looking for already, implemented in `filters.py` for all models.

If you try (on develop):
```
curl http://kernel.aether.local:8000/projects/?name=AUX
```
Is that what you need?

aether-ui/aether/ui/api/utils.py
```diff
@@ -187,88 +187,143 @@ def create_project_schema_object(name, pipeline, schema_id, entity_name):
...
+def convertEntityTypes(entities_from_kernel):
+    result = {'schemas': [], 'ids': {}}
+    for entity in entities_from_kernel:
+        try:
+            project_schema = kernel_data_request(f'projectschemas/{entities_from_kernel[entity]}/')
+            schema = kernel_data_request(f'schemas/{project_schema["schema"]}/')
+            result['schemas'].append(schema['definition'])
+            result['ids'][schema['name']] = schema['id']
+        except Exception:
+            pass
```
quesebifurcan: I think we need to handle any errors here and not just let them pass through. If you expect `KeyError`s in the lookups, you could (for example) handle those explicitly. `except Exception` catches everything, including typos.

aether-ui/aether/ui/api/tests/test_utils.py
```diff
@@ -43,13 +43,24 @@ def test_create_new_kernel_object(self):
...
             utils.create_new_kernel_object('project', pipeline, {}, 'Aux-test')
         exception = ast.literal_eval(str(exc.exception))
         self.assertEqual(exception['object_name'], 'unknown')
         self.assertFalse(utils.is_object_linked(pipeline.kernel_refs, 'schema', 'Person'))
+
+    def test_convert_entity_types(self):
+        with self.assertRaises(Exception) as exc:
+            utils.convertEntityTypes({'Person': '123456'})
+            exception = ast.literal_eval(str(exc.exception))
+            self.assertEqual(exception['object_name'], 'unknown')
```
quesebifurcan, line 69: Could we add a test for the success case as well?

aether-ui/aether/ui/api/tests/test_views.py
```diff
@@ -353,8 +364,51 @@ def test_view_publish_pipeline(self):
...
+        outcome = {
+                    'successful': [],
+                    'error': [],
+                    'exists': []
+                }
+        outcome = utils.publish_preflight(pipeline, 'Aux', outcome)
+        self.assertGreater(len(outcome['error']), 0)
+        self.assertEqual(len(outcome['exists']), 0)
+
+    def test_view_02_pubish_pipeline(self):
```
quesebifurcan: Could we avoid using numbers in function here? `test_publish_pipeline` would be enough. Numbers are tricky to maintain -- what if we move tests around or want to add a new inbetween 2 and 3?

aether-kernel/aether/kernel/api/views.py
```diff
@@ -65,12 +65,34 @@ class ProjectViewSet(CustomViewSet):
...
+    def byname(self, request):
+        '''
+        This view returns mapping filtered by the passed name
+        '''
+        name = self.request.query_params.get('name', None)
+        filtered_list = utils.find_by_name('Mapping', name)
+        serialized_data = serializers.MappingSerializer(
+                           filtered_list, context={'request': request}, many=True).data
+        return Response(serialized_data, status=HTTPStatus.OK)
+
```
quesebifurcan: (Same as above)

quesebifurcan (changes requested):

aether-ui/aether/ui/api/tests/test_utils.py
lordmallam: Success case test can be found [here](https://github.com/eHealthAfrica/aether/pull/167/files#diff-a6f7f46e059290d2c7ee9e3cc99bb4deR425)

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -525,3 +525,9 @@ def validate_entity_payload(project_Schema, payload):
         return
     except Exception as err:
         raise err
+
+
+def find_by_name(model_name, name):
+    model = getattr(models, model_name)
+    queryset = model.objects.all()
+    return queryset.filter(name=name)
```
quesebifurcan: Do we still need this? Looks like it was only used by the `byname` methods you removed.

aether-ui/aether/ui/api/utils.py
```diff
@@ -159,16 +158,23 @@ def create_new_kernel_object(object_name, pipeline, data={}, project_name='Aux',
         pipeline.save()
 
 
+def update_kernel_object(object_name, id, data={}):
+    try:
+        kernel_data_request(f'{object_name.lower()}s/{id}/', 'put', data)
+    except Exception as e:
+        raise Exception(e)
```
quesebifurcan: I think this could be simplified to `kernel_data_request(f'{object_name.lower()}s/{id}/', 'put', data)` -- the exception is just caught and immediately reraised. Also, it's always nice to get the return value (especially the status code) from a function using `requests`.

aether-ui/aether/ui/api/utils.py
```diff
@@ -187,88 +194,190 @@ def create_project_schema_object(name, pipeline, schema_id, entity_name):
                             'project': pipeline.kernel_refs['project'],
                             'schema': schema_id
                         }
-    if not is_object_linked(pipeline.kernel_refs, 'projectSchema', entity_name):
         create_new_kernel_object('projectSchema',
                                  pipeline, project_schema_data, entity_name=entity_name)
 
 
-def publish_pipeline(pipelineid, projectname):
+def publish_preflight(pipeline, project_name, outcome):
```
quesebifurcan: This is a rather long function -- could you add a docstring which briefly outlines what it's supposed to do?

aether-ui/aether/ui/api/utils.py
```diff
@@ -187,88 +194,190 @@ def create_project_schema_object(name, pipeline, schema_id, entity_name):
...
+    return [
+        {'source': mapping[0], 'destination': mapping[1]}
+        for mapping in mapping_from_kernel
+    ]
+
+
+def convertEntityTypes(entities_from_kernel):
+    result = {'schemas': [], 'ids': {}}
+    for entity in entities_from_kernel:
+        project_schema = kernel_data_request(f'projectschemas/{entities_from_kernel[entity]}/')
```
quesebifurcan: Could you try `for entity, entity_id in entities_from_kernel.items()`?
Then we could avoid the dictionary lookups inside the format strings.

aether-ui/aether/ui/api/utils.py
```diff
@@ -187,88 +194,190 @@ def create_project_schema_object(name, pipeline, schema_id, entity_name):
...
+            'mapping': kernel_object['id']
+        }
+    )
+
+
+def kernel_to_pipeline():
+    mappings = kernel_data_request('mappings/')['results']
+    for mapping in mappings:
+        if not is_linked_to_pipeline('mapping', mapping['id']):
+            create_new_pipeline_from_kernel(mapping)
```
quesebifurcan: Could we make sure that every function returns a value? Otherwise, they can become quite difficult to debug.

quesebifurcan (changes requested):

aether-ui/aether/ui/api/views.py
```diff
@@ -15,14 +17,49 @@ class PipelineViewSet(viewsets.ModelViewSet):
...
     ordering = ('name',)
 
+    @action(methods=['post'], detail=False)
+    def fetch(self, request):
+        '''
+        This view gets kernel objects, transforms and loads into a pipeline
+        '''
+        ui_utils.kernel_to_pipeline()
+        pipelines = models.Pipeline.objects.all()
+        serialized_data = serializers.PipelineSerializer(pipelines, context={'request': request}, many=True).data
```
tooxie, line 27: Is there any possible exception being thrown by this?

aether-ui/aether/ui/api/views.py
```diff
@@ -15,14 +17,49 @@ class PipelineViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.PipelineSerializer
     ordering = ('name',)
 
+    @action(methods=['post'], detail=False)
+    def fetch(self, request):
+        '''
+        This view gets kernel objects, transforms and loads into a pipeline
+        '''
+        ui_utils.kernel_to_pipeline()
```
tooxie, line 25: This can raise [a couple of exceptions](https://github.com/eHealthAfrica/aether/blob/aet-ui-load-mappings-from-kernel/aether-ui/aether/ui/api/utils.py#L128-L136) that are not being handled anywhere.

aether-ui/aether/ui/api/views.py
```diff
@@ -15,14 +17,49 @@ class PipelineViewSet(viewsets.ModelViewSet):
...
+        overwrite = False
+        if 'overwrite' in request.data:
+            overwrite = True
+        outcome = {
+            'successful': [],
+            'error': [],
+            'exists': []
+        }
+        try:
+            pipeline = get_object_or_404(models.Pipeline, pk=pk)
```
tooxie, line 46: Looks like you are ignoring the `_or_404` part of the shortcut. Why don't you use `models.Pipeline.get()` instead which will raise a `DoesNotExist` exception if the object is not found?

aether-ui/aether/ui/api/views.py
```diff
@@ -15,14 +17,49 @@ class PipelineViewSet(viewsets.ModelViewSet):
...
+        except Exception as e:
+            outcome['error'].append(str(e))
+            return Response(outcome, status=HTTPStatus.BAD_REQUEST)
+        outcome = ui_utils.publish_preflight(pipeline, project_name, outcome)
+        if len(outcome['exists']):
```
tooxie: An empty list evals as `False` in python, you don't need to call `len()`.

aether-ui/aether/ui/api/views.py
```diff
@@ -15,14 +17,49 @@ class PipelineViewSet(viewsets.ModelViewSet):
...
+            if overwrite:
```
tooxie: What is being overwritten here?

aether-ui/aether/ui/api/views.py
```diff
@@ -15,14 +17,49 @@ class PipelineViewSet(viewsets.ModelViewSet):
...
+                outcome = ui_utils.publish_pipeline(pipeline, project_name, True)
+            else:
+                return Response(outcome, status=HTTPStatus.BAD_REQUEST)
+        else:
+            outcome = ui_utils.publish_pipeline(pipeline, project_name)
```
tooxie, line 57: This call is almost identical to line 53's. Would it be possible to refactor this to prevent this code duplication?

aether-ui/aether/ui/api/views.py
```diff
@@ -15,14 +17,49 @@ class PipelineViewSet(viewsets.ModelViewSet):
...
+
     @action(methods=['post'], detail=True)
     def publish(self, request, pk=None):
         '''
         This view transform the supplied pipeline to kernal models,
         publish and update the pipeline with related kernel model ids.
         '''
         project_name = request.data['project_name'] if 'project_name' in request.data else 'Aux'
-        return ui_utils.publish_pipeline(pk, project_name)
+        overwrite = False
```
tooxie, line 37: TBH I find this code very hard to follow. What I think could be changed:
* There are 4 return calls in 15 lines (3 of those are `return Response(outcome, status=HTTPStatus.BAD_REQUEST)`)
* `outcome['error']` gets written and deleted at the end. Maybe not add it unless necessary?
* too many `if/else` branches
* almost no error handling

I think it would help if we could flatten the code.

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -50,6 +50,7 @@ class Pipeline extends Component {
   }
 
   publish () {
+    // todo: check if and overwrite is required
```
tooxie, line 53: How do we make sure this doesn't get overlooked and forgotten? Would it be possible to create a jira ticket instead?

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -20,6 +20,7 @@ class PipelineList extends Component {
     if (!this.props.pipelineList.length) {
       this.props.getPipelines()
     }
+    this.props.fetchPipelines()
```
tooxie, line 23: No semicolons? 🤔

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
```diff
@@ -74,6 +76,12 @@ export const getPipelines = () => ({
   )
 })
 
+export const fetchPipelines = () => ({
+  types: ['', types.GET_FROM_KERNEL, types.GET_FROM_KERNEL_ERROR],
+  promise: client => client.post(
+    `${urls.PIPELINES_URL}fetch/?limit=5000`,
```
tooxie, line 82: Why 5000?

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
```diff
@@ -140,11 +148,15 @@ const reducer = (state = INITIAL_PIPELINE, action) => {
     }
 
     case types.PIPELINE_PUBLISH_SUCCESS: {
-      return { ...state, publishSuccess: action.payload, publishError: null }
+      return { ...state, publishSuccess: action.payload.successful }
     }
 
     case types.PIPELINE_PUBLISH_ERROR: {
-      return { ...state, publishSuccess: null, publishError: action.error }
+      return { ...state, publishError: action.error.error }
```
tooxie, line 155: Why did `action.error` became `action.error.error`? I don't see that being used anywhere else.

aether-ui/aether/ui/assets/apps/pipeline/sections/Output.jsx
```diff
@@ -3,10 +3,61 @@ import { FormattedMessage } from 'react-intl'
...
+      Object.keys(exists).forEach(exist => {
+        errorList.push(<li key={exist}>
+          <FormattedMessage id={`publish.exists.${exist}`} defaultMessage={exists[exist]} />
+        </li>)
+      })
+    })
+    return <ul>{errorList}</ul>
+  }
+
+  buildPublishSuccess (success) {
```
tooxie: Just `success` is a misleading name for a list, sounds like a boolean. What about `publishedPipelines` for example?

aether-ui/aether/ui/assets/apps/pipeline/sections/Output.jsx
```diff
@@ -3,10 +3,61 @@ import { FormattedMessage } from 'react-intl'
...
+    const successList = []
+    success.forEach(passed => {
+      successList.push(<li key={passed}>
+        <FormattedMessage id={`publish.success.${passed}`} defaultMessage={passed} />
+      </li>)
+    })
+    return <ul>{successList}</ul>
```
tooxie: IMHO it would be more idiomatic to reduce the list instead:
```js
buildPublishSuccess (publishedPipelines) {
  const formattedMessages = publishedPipelines.reduce((formattedList, pipe) => {
    return (
      <li key=pipe>
        <FormattedMessage id={`publish.success.${pipe}`} defaultMessage={pipe} />
      </li>
    );
  });

  return <ul>{formattedMessages}</ul>;
}
```

And you don't really need the `formattedMessages` throwaway list actually.

aether-ui/aether/ui/assets/apps/pipeline/sections/Output.jsx
```diff
@@ -3,10 +3,61 @@ import { FormattedMessage } from 'react-intl'
...
+  }
+
   render () {
     return (
       <div className='section-body'>
-        <div className='pipeline-errors'>
+        {
+          this.props.publishError && (
```
tooxie, line 36: IMHO this is a misuse of a binary operator, and makes the code harder to read. I think that you could move that to a function instead:
```js
const getErrors = (errors) => {
  if (errors) {
    return <div className='pipeline-errors'> ... </div>
  }
}

render() {
  return (
    <div className='section-body'>
      {
        getErrors(this.props.publishError);
        getFormattedMessages(this.props.publishSuccess);
      }
    </div>
  );
}
```

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
lordmallam: This is implemented [here](https://github.com/eHealthAfrica/aether/blob/aet-ui-publish-actions/aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx#L130) comment to be removed. PR coming soon

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
lordmallam: Nope.. 😉

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
lordmallam: Just to make sure all records are returned.. pagination not fully in place

aether-ui/aether/ui/assets/apps/pipeline/sections/Output.jsx
lordmallam: `short-circuit operators` `&&` doesn’t evaluate the right-hand side expression, if just by evaluating the left-hand expression can decide the final result.

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
lordmallam: Because of the `error` object from [here](https://github.com/eHealthAfrica/aether/pull/167#discussion_r190297524)

aether-ui/aether/ui/api/views.py
lordmallam: same function is called with a different arg. What refactor suggestions do you have?

aether-ui/aether/ui/api/views.py
lordmallam: Objects in the kernel db

aether-ui/aether/ui/api/views.py
lordmallam: `_or_404` raises a `Http404` exception which is noted in `L48`. I agree a simpler message can suffice

aether-ui/aether/ui/api/views.py
lordmallam: the logic implemented covers the possible outcome from a pipeline publish. All errors/exceptions and successes are logged in the `outcome` dict and handled on a conditional bases.

Torn between maintaining the `outcome['error']` list in one function or having to check if the list is exists or not before declaration (not to overwrite existing entries),  and adding to it in multiple places. I went for managing it in one place.

quesebifurcan (approved):

quesebifurcan:
Looks good :+1: let's get @tooxie 's suggestions in here as well and then I think it'll be good to go.

aether-ui/aether/ui/assets/apps/pipeline/sections/Output.jsx
tooxie: Now that I think about it, we are going through all published pipelines, so a `map` is better suited for this.

aether-ui/aether/ui/assets/apps/pipeline/sections/Output.jsx
tooxie: You are right, and it works, but my concern is more about readability of the code.

aether-ui/aether/ui/api/views.py
tooxie: It's not about simplicity imo, but about using `get_object_or_404` just to test existence of an object. The function is not returning 404 (as I expected when I saw the `get_objects_or_404` call) but 400 instead. So I think it's more suitable to test explicitly for the existence of the object because it explains the purpose of the code.

aether-ui/aether/ui/api/views.py
tooxie: This would be my suggestion:
```py
class BadRequestResponse(Response):
    def __init__(self, error=None, outcome=None):
        if not outcome:
            outcome = {
                'successful': [],
                'error': [],
                'exists': [],
            }

        if error:
            outcome['error'].append(error)

        super(BadRequestResponse, self).__init(outcome, status=HTTPStatus.BAD_REQUEST)


        '''
        This view transform the supplied pipeline to kernel models,
        '''
        project_name = request.data.get('project_name', 'Aux')
        overwrite = request.data.get('overwrite', False)
        outcome = {}

        try:
            pipeline = models.Pipeline.get(pk=pk)
        except DoesNotExist as error:
            return BadRequestResponse(error=error)

        outcome = ui_utils.publish_preflight(pipeline, project_name, outcome)
        if outcome['exists'] and not overwrite:
            return BadRequestResponse(outcome=outcome)
        else:
            outcome = ui_utils.publish_pipeline(pipeline, project_name, overwrite)

        if outcome['error']:
            return BadRequestResponse(outcome=outcome)

        return Response(outcome, status=HTTPStatus.OK)
```

I'm not sure this code works, I haven't tried it, but it's just to give you an idea. I abstracted all the responses into its own class to make it clear which return statements are errors and which are `OK`.

It also flattens the couple of nested ifs by combining them into `if outcome['exists'] and not overwrite`. Please triple check that this expression is correct, I'm terrible at boolean logic :)

And I left the final `return` at the end, outside of any if branch, to make it visually clear that this function cannot return `undefined`.

Again, this is just an idea, treat it as pseudo code :)

aether-ui/aether/ui/api/views.py
lordmallam: I understand. Here's the discussion on how we got here https://github.com/eHealthAfrica/aether/pull/151/files/c903d82cad16ef6b913b1cf96cbbf059c6b1c7b0#r187016024

aether-ui/aether/ui/api/views.py
lordmallam: Nice. I like the `BadRequestResponse` abstraction. Due to current usage of the `outcome` variable in especially the `/utils` file, i will look into the refactor in another PR

aether-ui/aether/ui/api/views.py
tooxie: I see. The difference, if I understand it correctly, is that in the case you mention the function actually returns a 404 if the object doesn't exist. In this case however, we return 400. Is that correct?

PR #170, opened 2018-05-17
feat(odk): parse xForm into to AVRO schema

obdulia-losantos:
https://jira.ehealthafrica.org/browse/AET-246

- Reorganized code.
- Use `pyxform` methods when possible instead of handmade ones.
- Include more use cases in the samples, like empty geopoints, empty arrays, nested arrays (Thanks to @quesebifurcan ), default values and multilanguage.
- Improve the xForm validation method.
- Generate AVRO schema from xForm definition and save in the database.

Assumptions:
- `pyxform` methods work as expected (like `XFormInstanceParser`).
- The instance skeleton is included in the xForm definition (the only `instance` tag without id). The survey fields appear in order.
- ODK collect will send back the whole skeleton with null values if needed. This means that only leafs can be null.


quesebifurcan:
Looks great! As mentioned, I ran into some issues with nested repeats. Pushed an example here: https://github.com/eHealthAfrica/aether/commit/f11b4b8834c022710d22b94e4eacff01ac975e50

Current output of nested repeats:
```
...
{
      "name": "Repeat_1",
      "doc": null,
      "type": {
        "type": "array",
        "items": {
          "name": "Repeat_1",
          "doc": null,
          "type": "record",
          "fields": [
            {
              "name": "name",
              "type": [
                "null",
                "string"
              ],
              "doc": "name_label"
            },
            {
              "name": "Repeat_2",
              "doc": null,
              "type": "record",
              "fields": [
                {
                  "name": "name2",
                  "type": [
                    "null",
                    "string"
                  ],
                  "doc": "name_label"
                }
              ]
            }
          ]
        }
      }
    },
...
```

obdulia-losantos:
Looking into it, thanks for finding this!

obdulia-losantos:
@quesebifurcan better?

```json
{
  "name": "NestedRepeatsTest",
  "namespace": "aether.odk.xforms",
  "doc": "nested repeats test (id: nested_repeats_test, version: 0)",
  "type": "record",
  "fields": [
    {
      "name": "_id",
      "doc": "xForm ID",
      "type": "string",
      "default": "nested_repeats_test"
    },
    {
      "name": "_version",
      "doc": "xForm version",
      "type": "string",
      "default": "0"
    },
    {
      "name": "Repeat_1",
      "type": {
        "type": "array",
        "items": {
          "name": "Repeat_1",
          "doc": null,
          "type": "record",
          "fields": [
            {
              "name": "name",
              "type": [
                "null",
                "string"
              ],
              "doc": "name_label"
            },
            {
              "name": "Repeat_2",
              "type": {
                "type": "array",
                "items": {
                  "name": "Repeat_2",
                  "doc": null,
                  "type": "record",
                  "fields": [
                    {
                      "name": "name2",
                      "type": [
                        "null",
                        "string"
                      ],
                      "doc": "name_label"
                    }
                  ]
                }
              }
            }
          ]
        }
      }
    },
    {
      "name": "meta",
      "type": {
        "name": "meta",
        "doc": "/meta",
        "type": "record",
        "fields": [
          {
            "name": "instanceID",
            "type": [
              "null",
              "string"
            ],
            "doc": "/meta/instanceID"
          }
        ]
      }
    }
  ]
}
```

quesebifurcan:
@obdulia-losantos looks good!

I tried generating a value for your new schema, the output makes sense to me.
```
{
  "_id": "rhEZDCKCxjRhdUnWCZkOFt",
  "_version": "GJR",
  "Repeat_1": [
    {
      "name": "DEzwKXHrOktBXcMEzzmtMyPDy",
      "Repeat_2": [
        {
          "name2": "BqIfWMwDzrsAhhSNBJfQdZLPcYLrzp"
        },
        {
          "name2": null
        },
        {
          "name2": null
        }
      ]
    }
  ],
  "meta": {
    "instanceID": "BWXbtZwZqWahghCUWyj"
  }
}
```

quesebifurcan:
@obdulia-losantos Looking at groups, the following xform doesn't result in a valid avro schema, as far as I can tell:

```
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:odk="http://www.opendatakit.org/xforms" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <h:head>
    <h:title>cascading select test</h:title>
    <model>
      <instance>
        <groups id="cascading_select_test">
          <group1>
            <name/>
          </group1>
          <meta>
            <instanceID/>
          </meta>
        </groups>
      </instance>
      <bind nodeset="/groups/group1/name" type="string"/>
      <bind calculate="concat('uuid:', uuid())" nodeset="/groups/meta/instanceID" readonly="true()" type="string"/>
    </model>
  </h:head>
  <h:body class="pages">
    <group ref="/groups/group1">
      <label>group1_label</label>
      <input ref="/groups/group1/name">
        <label>name_label</label>
      </input>
    </group>
  </h:body>
</h:html>
```
Output from `parse_xform_to_avro_schema`:
```
var schema = {
  "name": "CascadingSelectTest",
  "doc": "cascading select test (id: cascading_select_test, version: 0)",
  "type": "record",
  "fields": [
    {
      "name": "_id",
      "doc": "xForm ID",
      "type": "string",
      "default": "cascading_select_test"
    },
    {
      "name": "_version",
      "doc": "xForm version",
      "type": "string",
      "default": "0"
    },
    {
      "name": "group1",
      "doc": "group1_label",
      "type": "record",
      "fields": [
        {
          "name": "name",
          "type": [
            "null",
            "string"
          ],
          "doc": "name_label"
        }
      ]
    },
    {
      "name": "meta",
      "doc": "/meta",
      "type": "record",
      "fields": [
        {
          "name": "instanceID",
          "type": [
            "null",
            "string"
          ],
          "doc": "/meta/instanceID"
        }
      ]
    }
  ]
}
```
I think the issue is that the field "group_1" has the type "record", whereas it should be `"type": {"type": "record"...etc.`

UPDATE: fixed in https://github.com/eHealthAfrica/aether/pull/170/commits/58dc4acbeab10e14f7c26bd6f353c60aa4599f1c

obdulia-losantos:
@quesebifurcan I think that I concluded with the changes.

Validation is included but because we know that is not 100% reliable in `spavro` I created our own `validate names` method in the meantime.

shawnsarwar (dismissed):
Looks great! My only suggestion is that we find a way to package the XForm -> Avro functionality and package it for PyPi as I think it's generally useful.

aether-odk-module/aether/odk/api/xform_utils.py
```diff
@@ -16,265 +16,772 @@
...
+                return instance
+
+            # assumption: there is only one child (key)
+            key = list(instance.keys())[0]
+            return instance[key]
+
+    except Exception:
+        pass
+
+    raise TypeError('missing instance definition')
```
quesebifurcan: Maybe this would be a good place for an application-specific error? TypeError is very general, I'd expect to see that raised when python-related issues happen, such as `int({})`. And this is more related to the shape/content of the xml, if I'm not mistaken. 

Also, could we raise the new error inside the `Exception` block?
```
except KeyError:
    raise MissingInstanceError('...')
```

aether-odk-module/aether/odk/api/xform_utils.py
```diff
@@ -16,265 +16,772 @@
...
-                # transform it back into a list
-                obj[k] = [v]
+        # convert all text entries in a dict wich key is the text id
+        texts = __wrap_as_list(translation['text'])
+        return {text_entry['@id']: text_entry['value'] for text_entry in texts}
 
-            if isinstance(v, dict):
-                walk(v, keys, coerce_dict)
+    except Exception:
+        pass
```
quesebifurcan: Same thing here -- `except Exception: pass` is bit dangerous -- if any programming errors or typos sneak in, the result will always be an empty dict.

aether-odk-module/aether/odk/api/xform_utils.py
```diff
@@ -16,265 +16,772 @@
...
+            label_tag = tag['label']
+            if isinstance(label_tag, dict):
+                # there are more than one language defined in the form
+                #   <label ref="jr:itext('{xpath}:label')"/>
+                label_id = label_tag['@ref'][10:-2]  # f'{xpath}:label'
+                label = texts[label_id]
+            elif label_tag:
+                label = label_tag
+    except Exception:
+        pass
```
quesebifurcan: And here :-)

quesebifurcan (changes requested):

obdulia-losantos:
@quesebifurcan fixed! :+1:

aether-odk-module/aether/odk/api/models.py
```diff
@@ -16,18 +16,48 @@
...
+                           |   | form_id          |   +--<| xform            |
+                           |   | version          |       +------------------+
+                           |   | avro_schema      |
+                           |   | md5sum           |
+                           |   +::::::::::::::::::+
+                           +--<| mapping          |
+                               | surveyors (User) |
+                               +------------------+
+
+'''
```
quesebifurcan, line 60: Beautiful!

quesebifurcan (approved):
Great work!

PR #177, opened 2018-05-25
feat(odk): improve data model

obdulia-losantos:
Changed:

In ODK module:
- renamed model `Mapping` to `Project`.
- added new field `kernel_id` in `XForm` model to keep track of the linked instances in Kernel.
- added replication methods from ODK to Kernel with default values.
- in view submission indicate missing instance ID ASAP, try to avoid useless db searching if possible.
- PATCH endpoint `{model}/{pk}/propagates` for projects and xforms.

Chore:
- adapted tests scripts, odk module does not need previous data in kernel.

https://docs.google.com/document/d/1wXVFfpWtDWeWSJiygx4GGdv6tOEX5wRTQMNdefnL6H0/edit


obdulia-losantos:
Open questions: 
- should we try to replicate every time a project or xform is created/updated?
- should we delete the kernel copies if the instance is deleted in odk?

aether-odk-module/aether/odk/api/kernel_replication.py
```diff
@@ -0,0 +1,185 @@
...
+import requests
+
+from aether.common.kernel.utils import get_auth_header, get_kernel_server_url
+
+
+class KernelReplicationError(Exception):
+    pass
+
+
+def replicate_project(project):
```
admbtlr: I'm not sure that `replicate` is the right term here. It's a little confusing, because it implies an exact copy, which is not really true. Maybe `create_kernel_project`?

aether-odk-module/aether/odk/api/kernel_replication.py
```diff
@@ -0,0 +1,185 @@
...
+        },
+        # no need to update any property there
+        item_update=None,
+    )
+
+    # indicates that the item is in Kernel
+    return True
+
+
+def replicate_xform(xform):
```
admbtlr: `create_kernel_artefacts`?

aether-odk-module/aether/odk/api/kernel_replication.py
```diff
@@ -0,0 +1,185 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+import json
+import requests
+
+from aether.common.kernel.utils import get_auth_header, get_kernel_server_url
+
+
+class KernelReplicationError(Exception):
```
admbtlr: `KernelPropagationError`?

admbtlr (changes requested):

aether-odk-module/aether/odk/api/kernel_replication.py
```diff
@@ -0,0 +1,185 @@
...
+            'project': project_id,
+            'schema': item_id,
+        },
+    )
+
+    # indicates that the items are in Kernel
+    return True
+
+
+def __upsert_item(item_model, item_id, item_new, item_update=None):
```
quesebifurcan: This method would be very useful for other modules (for example, I saw a lot of similar things in the `aether-ui` module recently). What do you think about adding this method to `kernel` instead? That would allow us to have the actual upsert logic run closer to the database.The odk-specific error handling would stay here, of course.

aether-odk-module/aether/odk/api/views.py
```diff
@@ -89,12 +104,33 @@ class XFormViewSet(viewsets.ModelViewSet):
...
+
+        Reachable at ``.../xforms/{pk}/replicates/``
+        '''
+
+        xform = get_object_or_404(XForm, pk=pk)
+        xform.save()  # creates avro schema if missing
+
+        try:
+            replicate_xform(xform)
+        except KernelReplicationError as kre:
```
quesebifurcan: Just some thoughts: what happens if one of the kernel requests fails but the others succeed? Is there a risk that we end up with e.g. `mapping` and `projectschema` but no `schema`? If so, how do we recover from such errors?

An alternative would be to add a post view in kernel which attempts to create/update a project from request data like:
```
{"mapping": {...}, "schema": {...}, "projectschema":{...}}
```
...and wrap that view in a transaction, so the entire thing would roll back if e.g. the schema wasn't valid AVRO.

quesebifurcan (changes requested):

aether-odk-module/aether/odk/api/kernel_replication.py
obdulia-losantos: :thinking:

aether-odk-module/aether/odk/api/views.py
obdulia-losantos: I see this method being used by the UI publish feature :stuck_out_tongue: 

So, should we create an standard one that can be used by both systems? later even more apps ;-)

quesebifurcan: Precisely :smile: I think it would be helpful for every non-trivial aether module.

obdulia-losantos: If you don't mind I would rather prefer to do it in another PR and afterwards use that in this one... it's going to be clearer for everybody at the end. What do you think? :thinking:

quesebifurcan: Sure, that's fine with me :-) Maybe create a ticket so we remember to do it?

obdulia-losantos: Done in #180

obdulia-losantos:
@quesebifurcan now looks much simpler :wink:

admbtlr (approved):

quesebifurcan (approved):

PR #180, opened 2018-05-29
feat(kernel): generate artefacts endpoint

obdulia-losantos:
Finally the endpoint that @quesebifurcan always wanted to have :wink: 

Reachable at `/projects/{pk}/artefacts/` 

Open question: should we use the :gb: British or the :us: American version of the word?

This is a little bit tricky because if there is no project with that primary key, the endpoint will create it, and maybe this is a way to create random projects or to collapse the app. :thinking: 

This is pretty simple to use, the endpoint allows only two methods `GET` and `PATCH`. (Why not `POST` or `PUT`? Because in the RESTful world, `POST` means create and `PUT` update and we are doing more than that, yes, `PATCH` means also partial update, nobody is perfect)

The `GET` one will return the list of **all** artefact IDs related to the project.

The `PATCH` one will create the project if it doesn't exist but not update and will create/update the indicated artefacts and return the list of **affected** artefacts (if the project has more, they are not going to be included in the response). Given and `id` in any of the entries doesn't mean that the instance must exists, but in case of not, a new one with that id will be created.

The expected payload would be something like this:

```javascript
{
  // this is optional, if missing the method will assign a random name
  "name": "project name (optional but unique)",

  // this is optional, for each entry the method will create/update a schema and also link it to the project (projectschema entry)
  "schemas": [
    {
      "id": "schema id (optional)",
      "name": "schema name, optional but unique",
      "definition": {
        // the avro schema, I guess
      },
      "type": "record" // there are no docs about this value... 
    },
    ...
  ],

  // also optional
  "mappings": [
    {
      "id": "mapping id (optional)",
      "name": "mapping name, optional but unique",
      // optional
      "definition": {
        "mapping": [
          // the mapping rules
        ]
      }
    },
    // ...
  ]
}
```


aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -0,0 +1,163 @@
...
+        results['schemas'].add(str(project_schema.schema.pk))
+        results['project_schemas'].add(str(project_schema.pk))
+
+    for mapping in Mapping.objects.filter(project=project):
+        results['mappings'].add(str(mapping.pk))
+
+    return results
+
+
+@transaction.atomic
```
quesebifurcan, line 50: :+1:

aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -0,0 +1,163 @@
...
+    results = {
+        'project': None,
+        'schemas': set(),
+        'project_schemas': set(),
+        'mappings': set(),
+    }
+
+    # 1. create/update the project
+    project = __upsert_instance(
+        Model=Project,
```
quesebifurcan: minor thing: I'm used to function parameters being lowercase, even if the type is a class. Example: https://docs.djangoproject.com/en/2.0/ref/models/querysets/#django.db.models.query.QuerySet

aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -0,0 +1,163 @@
...
+
+    is_new = True
+    if not pk:
+        item = Model()
+    else:
+        try:
+            item = Model.objects.get(pk=pk)
+            is_new = False
+        except ObjectDoesNotExist:
+            item = Model(pk=pk)
```
quesebifurcan: Is there a specific reason we're not using `update_or_create` here? Seems like it could be a good fit, but maybe there's an edge case or two I'm missing.

quesebifurcan (changes requested):
Looks great, nice that you included `IntegrityError` in the tests. Just a couple of comments.

aether-kernel/aether/kernel/api/project_artefacts.py
obdulia-losantos: In this case where we need to control the "ignore in case of update" fields will not work.

https://docs.djangoproject.com/en/2.0/ref/models/querysets/#update-or-create

```python

    def test__update_or_create(self):

        project_id = uuid.uuid4()

        project_1, created_1 = Project.objects.update_or_create(
            pk=project_id,
            defaults={'name': 'Project'},
        )
        self.assertTrue(created_1)
        self.assertEqual(project_1.pk, project_id)
        self.assertEqual(project_1.name, 'Project')

        project_2, created_2 = Project.objects.update_or_create(
            pk=project_id,
            defaults={'name': 'Project 2'},  # it's going to be updated
        )
        self.assertFalse(created_2)
        self.assertEqual(project_2.pk, project_id)
        self.assertEqual(project_2.name, 'Project 2')  # but we don't want to change the name

```

quesebifurcan (approved):

PR #181, opened 2018-05-29
fix: do not overwrite auto-generated avro record names.

quesebifurcan:
Do not overwrite auto-generated avro record names.
    
The function AvroSchemaViewer mutates its input. Clone
the schema before it enters the function.
    
Resolves https://jira.ehealthafrica.org/browse/AET-275


obdulia-losantos (approved):
_simplicity_

PR #185, opened 2018-05-30
Aet ui setup

lordmallam:
UI ==> Develop

PR #186, opened 2018-05-31
fix(ui): spring cleaning before merging to develop

obdulia-losantos:
If you don't mind I'd love to merge this before #185 

Actions:

- Removed `setup_projects` in `entrypoint.sh`
- Removed `UserTokens` and related: decorator, view...
- Joined test files by concept due to the above cleaning.
- Squashed migration files into one, maybe it's better to delete the others and rename it to `0001_initial.py` to have a start cleaning. What do you think?
- Changed the API url from `/api/ui/` to simple `/api/` (it made sense with the tokens feature)


lordmallam:
If i remember correctly the `ui` was setup with the intention of making it a "boiler-plate" app for integration to Aether; Thus, the `proxyViews` and url structure. Has that changed?

obdulia-losantos:
The `ProxyView` pretends to "proxy" the frontend request (in the react app) from the current server to the real server, without interacting with the call, just include the needed authorization header. So far, the react app only uses its own API...

The `UserTokens` is something that is used in gather to keep track of who made which change and when (handled in kernel by the `reversion` module). But we are using the ADMIN token to "publish" the pipeline and that makes the `UserTokens` useless.

If this is going to be a boiler-plate app, let's make it as simple as possible! Please, do not introduce useless features that complicate the code and make it difficult to understand.

Does it make sense for you?

lordmallam:
Ok, ill leave it to @quesebifurcan then, he has more insight on the `Proxy` functions as he wrote them.

obdulia-losantos:
:joy: No, I wrote them for Gather. But we can wait for him, no problem at all.

lordmallam (approved):
👍 . So we consider extracting proxy functions to a separate module

PR #188, opened 2018-06-04
fix: remove mutable default arguments

quesebifurcan:
Remove mutable default arguments from all aether modules and change camelcased function parameters and variable names to snake case.


obdulia-losantos (approved):

shawnsarwar (approved):
Good catch.

PR #189, opened 2018-06-04
feat: "identity mapping"

quesebifurcan:
This PR adds an "Add Identity Mapping" button, which, when clicked, generates a set of default entity types and mappings.

How to test: 
- Open the ui, paste some json into the "JSON data" section
- Click "Derive Schema"
- Click "Add Identity Mapping"
- Check that the created entity types and mappings were created and that they conform to your schema.


aether-ui/aether/ui/assets/apps/components/Modal.jsx
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react'
 import { injectIntl } from 'react-intl'
 import Portal from './Portal'
 
-class Modal extends Component {
+export class Modal extends Component {
```
obdulia-losantos: :thinking: 

Maybe we should delete the useless last line `export default injectIntl(Modal)` and change this one into `export default class...`

What do you think?

aether-ui/package-lock.json
```diff
@@ -0,0 +1,15168 @@
+{
```
obdulia-losantos: Is this file really needed?  If not, maybe we can add the rule in `.gitignore` to skip adding everytime someone uses `git add .`

aether-ui/package.json
```diff
@@ -57,6 +57,7 @@
     "redux-devtools-extension": "~2.13.0",
     "sass-lint": "~1.12.0",
     "sass-loader": "~7.0.0",
+    "sinon": "^5.0.10",
```
obdulia-losantos: As I use to mention, the dependencies uses the `~` flag and not the `^` one that brings to the image the patches and the minor changes. We faced in the past annoying errors when some of the libraries didn't follow the semantic release conventions and include breaking changes as minor releases...

aether-ui/aether/ui/assets/apps/utils/index.jsx
```diff
@@ -92,3 +92,9 @@ export const generateSchemaName = (prefix) => {
     }
   }
 }
+
+/* Locate a DOM element by its data-qa attribute */
```
obdulia-losantos: If this is only used in tests maybe you could include it in the test environment package :grimacing:

shawnsarwar (changes requested):
Looking great! A couple of things I noticed:
Auto_0 is automatically used for the entity name. If we could configure that, it would be nice.
The identity mapping should include a id field that's automatically mapped from #!uuid.
The identity schema marks all fields as required which could cause issues. It might be best to have all generated types except id be null optional to keep things flexible since it's not really a proper schema.

aether-ui/aether/ui/assets/apps/pipeline/sections/Input.spec.jsx
```diff
@@ -0,0 +1,101 @@
...
+      expect(mappingRule.destination).toEqual(result[i].destination)
+    })
+  })
+})
+
+describe('<IdentityMapping />', () => {
+  it('opens modal', () => {
+    const component = mountWithIntl(<IdentityMapping />)
+    expect(component.find(Modal).length).toEqual(0)
+    const button = component.find('[data-qa="input.identityMapping.btn-apply"]')
```
obdulia-losantos: couldn't this use your new `findByDataQa` function?

obdulia-losantos (changes requested):
not yet tested locally... but a couple of comments.

aether-ui/package-lock.json
quesebifurcan: npm recommends using it -- it ensures that we get the exact same set of dependencies (and dependencies of depencies) that we tested the app with.

aether-ui/package.json
quesebifurcan: Ah, I see. I just kept whatever npm gave me on install. package-lock.json might help, perhaps (see above)? Will change to `~` though :smile:

aether-ui/package-lock.json
obdulia-losantos: Curious, because in the `aet-ui-setup` you can see that the last version of this file (before taking it out) only included one library...

https://github.com/eHealthAfrica/aether/blob/7683c67bc3c084203989007352b4626203866514/aether-ui/package-lock.json

That's why we removed it.

aether-ui/package.json
obdulia-losantos: And I have the concern about `package-lock.json` usefulness... :wink:

aether-ui/package-lock.json
quesebifurcan: Ah, ok, I guess I was the first one using it, then (switching to the avsc library a couple of weeks back). It'll lead to a more reproducible build, as far as I can tell, but I'm really not the expert here.

obdulia-losantos (approved):

shawnsarwar (approved):

PR #195, opened 2018-06-05
 fix(odk): better error messages

obdulia-losantos:
- validate file content in serializer and sent possible errors linked to that field (not to xml_data)
- Indicate the missing required tags while validating xml
- bind entries are not mandatory, but nice to have
- change location schema (do not use geoJSON specification)
- use the open rosa response specification on ODK submissions


admbtlr (approved):
Bravo!

PR #196, opened 2018-06-06
chore: clean up setup

obdulia-losantos:


obdulia-losantos:
@shawnsarwar I saw a lot of legacy docker setup, entrypoints, requirements... that are not really used in the rest of `aether-utils` modules. I tried to clean them up, but I just ended up with more confusion and chaos than expected.

I would appreciate if you can take a look and refactor those modules at some point. :disappointed_relieved:

tooxie (approved):
🎉

PR #200, opened 2018-06-11
Removed `manager.stop()` call from `except` block

tooxie:
We are already doing it in the `finally` block, which will get executed
whether an exception occurs or not, so this call was redundant.


obdulia-losantos (approved):

shawnsarwar:
Included in Producer rewrite. See #209

PR #201, opened 2018-06-11
Limit the `count` query to 1

tooxie:
We don't need to know how many new items we have, we just need to know
whether we have new items. Besides, we were unwrapping them and
iterating over them in memory.

The parameter `topic` was removed because it was not being used. And the
ordering of elements was also removed because we don't need elements in
any particular order when we are just counting them, that only slows
down the query.

### DIFF
--- a/aether-producer/producer/aether_producer.py
+++ b/aether-producer/producer/aether_producer.py
@@ -29,7 +29,6 @@ from avro.datafile import DataFileWriter
 from avro.io import DatumWriter
 from avro.io import Validate
 
-
 from kafka import KafkaProducer, KafkaConsumer
 from aether.client import KernelClient
 from psycopg2.extras import DictCursor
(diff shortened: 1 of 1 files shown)


shawnsarwar:
Yeah, I had a much more efficient query for count, but it randomly stopped working and we couldn't get it back. Limit 1 is a great idea. I'm planning to rework this whole thing this week to operate and maintain state on a per-topic basis. Since you're AFAIK the only other person to really look at this (it was written as a POC and got merged because it was blocking other things) I'll definitely look to you for a review.

aether-producer/producer/aether_producer.py
```diff
@@ -126,14 +123,10 @@ def count_since(offset=None, topic=None):
...
-            count_str += '''AND ps.name = '%s'
-            ORDER BY e.modified ASC;''' % topic
-        else:
-            count_str += '''ORDER BY e.modified ASC;'''
-        cursor.execute(count_str);
-        return sum([1 for row in cursor])
+            LIMIT 1;
+        ''' % (offset));
+
+        return bool(cursor.fetchall())
```
obdulia-losantos, line 129: Following your thoughts could the SQL query be as simple as this one?

```sql
SELECT true FROM kernel_entity WHERE modified > '%s' LIMIT 1;
```

tooxie: I didn't go that far because I don't understand the schema that well to know if we can just drop the join altogether. But if it's possible I'd love to!

shawnsarwar: Yes indeed. Once we start maintaining individual offsets, we'll probably need a set of the existing entity types and a second parameter for type in this query. It's a much more efficient approach than what exists now.

shawnsarwar:
Idea used in producer rewrite.
Superceded by: https://github.com/eHealthAfrica/aether/pull/209

PR #203, opened 2018-06-11
feat: adding caching of expensive jsonpath.parse calls

shawnsarwar:
jsonpath_ng creates a new lexer instance each time .parse() is called. This is extremely slow. This pr adds caching of the result of the parse call, as it's always the same for the same input and can be reused for the subsequent .find() calls. We've replaced the jsonpath_ng.parse(jsonpath).find(obj) call with a single call to the static CachedParser of CachedParser.find(jsonpath, obj).

As a secondary note, this exposed an unwanted Django behavior. In debug mode, all db calls are cached causing unbounded memory use expansion. We now detect debug mode and on a random basis, there's a 1% chance the db cache will be cleared when performing entity extraction. This keeps memory stable while running in debug mode.


aether-kernel/aether/kernel/api/utils.py
```diff
@@ -35,6 +37,37 @@
 from . import models, constants
 
 
+# We need to detect instance running in debug mode. See CachedParser docstring
+DEBUG = os.environ['DEBUG'] == 'true'
+if DEBUG:
+    from random import randint
+    from django import db
```
obdulia-losantos: this debug value is already in django settings

```python

from django.conf import settings

if settings.DEBUG:
    # ...
```

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -35,6 +37,37 @@
...
+
+
+class CachedParser(object):
+    # jsonpath_ng.parse is a very time/compute expensive operation. The output is always
+    # the same for a given path. To reduce the number of times parse() is called, we cache
+    # all calls to jsonpath_ng here. This greatly improves extraction performance. However,
+    # when run in an instance of Django with debug=True, Django's db call caching causes massive
+    # memory usage. In debug, we must periodically call db.reset_queries() periodically to avoid
```
obdulia-losantos: we must **periodically** call db.reset_queries() **periodically** to... :thinking:

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -131,15 +164,17 @@ def find_by_jsonpath(obj, path):
 
         # Perform an standard jsonpath search.
         result = []
-        for datum in jsonpath_ng.parse(standard_jsonpath).find(obj):
+        matches = CachedParser.find(standard_jsonpath, obj)
+        for datum in matches:
```
obdulia-losantos: _datum_  / _data_
_forum_ / _fora_
...

back to Latin classes :nerd_face:

obdulia-losantos (changes requested):

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -542,6 +577,13 @@ def run_entity_extraction(submission):
             submission=submission,
         )
         entity_instance.save()
+    # If we're in debug, we need to periodically clear Django's caching of DB calls.
+    # This doesn't have to happen often so we do it randomly
+    if DEBUG:
+        wipe = randint(0, 100) == 100
+        if wipe:
+            db.reset_queries()
+
```
quesebifurcan: An alternative solution would be to just start the local server with `DEBUG="false"` (set this in `docker-compose.yml`) whenever we need to test code which generates a lot of debug-related artifacts. It would be nice to keep environment-specific details out of the code (if possible).

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -35,6 +37,37 @@
...
+    # memory usage. In debug, we must periodically call db.reset_queries() periodically to avoid
+    # unbounded memory requirements.
+
+    cache = {}
+
+    @staticmethod
+    def parse(path):
+        # we never need to call parse directly; use find()
+        if path not in CachedParser.cache.keys():
+            CachedParser.cache[path] = deepcopy(jsonpath_ng.parse(path))
```
quesebifurcan: Why do we need `deepcopy` here? Is the result of `jsonpath_ng.parse` mutated somewhere?

quesebifurcan (changes requested):

aether-kernel/aether/kernel/api/utils.py
shawnsarwar: Ah, forgot that was still there. I grew deeply paranoid at a number of times trying to track this down.

aether-kernel/aether/kernel/api/utils.py
shawnsarwar: I'm open to either; I thought adding the option to have it run in debug would be less problematic than forcing people to run in production mode or have a giant memory leak.

shawnsarwar: @obdulia-losantos ; which do you prefer? Check if debug and flush or ignore the leaky behavior in debug?

obdulia-losantos:
Stupid question, did you take into consideration the Django cache framework? I'm not really sure if this only applies to urls or to any piece of data you want to :thinking: 

https://docs.djangoproject.com/en/2.0/topics/cache/

aether-kernel/aether/kernel/api/utils.py
tooxie: Oh, latin! I thought it was german.

aether-kernel/aether/kernel/api/utils.py
obdulia-losantos: Note: take in mind that the tests are always run with `DEBUG=False`. This is due to the ### behavior of the debug toolbar in tests :unamused:

I agree with @quesebifurcan about keeping the code "agnostic" of this kind of issues. But also see your problem.

Maybe you can use this option locally to activate a cron job that reset the db every X minutes:
https://github.com/eHealthAfrica/aether/blob/develop/aether-common-module/aether/common/conf/settings.py#L252-L259

But maybe this is too much effort for a simple problem. :see_no_evil:

shawnsarwar:
@obdulia-losantos I took a look at it, but decided not to primarily as I didn't want to tie the functionality to Django since we'll probably want to externalize it into a consumer in the near future.

aether-kernel/aether/kernel/api/utils.py
shawnsarwar: I decided to leave the leak in DEBUG=true and the doc on CacheParser talking about the error.

quesebifurcan (dismissed):

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -19,6 +19,7 @@
 import collections
 import json
 import re
+import os
```
obdulia-losantos: is this `os` still in use?

obdulia-losantos (approved):
:see_no_evil:

PR #205, opened 2018-06-12
feat: (ui) mapping errors indicators

lordmallam:
Indicates errors on mapping sections

#### Related JIRA ticket [AET-273](https://jira.ehealthafrica.org/browse/AET-273)

### DIFF
--- a/aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
+++ b/aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
@@ -128,6 +128,7 @@ class Pipeline extends Component {
               className='pipeline-nav-item__output'
               onClick={() => this.toggleOutput()}>
               <div className='badge badge-small' />
+              <span className={`status ${this.props.selectedPipeline.mapping_errors.length > 0 ? 'red' : 'green'}`} />
               <FormattedMessage
                 id='pipeline.navbar.output'
                 defaultMessage='Output'
--- a/aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
+++ b/aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
@@ -37,7 +37,6 @@ class Pipeline extends Component {
   }
 
   componentWillReceiveProps (nextProps) {
-    console.log(nextProps.selectedPipeline)
     if (this.props.match.params.id !== nextProps.match.params.id) {
       this.props.getPipelineById(nextProps.match.params.id)
     }
--- a/aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
+++ b/aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
@@ -37,6 +37,7 @@ class Pipeline extends Component {
   }
 
   componentWillReceiveProps (nextProps) {
+    console.log(nextProps.selectedPipeline)
     if (this.props.match.params.id !== nextProps.match.params.id) {
       this.props.getPipelineById(nextProps.match.params.id)
     }
@@ -128,7 +129,9 @@ class Pipeline extends Component {
               className='pipeline-nav-item__output'
               onClick={() => this.toggleOutput()}>
               <div className='badge badge-small' />
-              <span className={`status ${this.props.selectedPipeline.mapping_errors.length > 0 ? 'red' : 'green'}`} />
+              { this.props.selectedPipeline.mapping.length > 0 &&
+                <span className={`status ${this.props.selectedPipeline.mapping_errors.length > 0
+                  ? 'red' : 'green'}`} /> }
               <FormattedMessage
                 id='pipeline.navbar.output'
                 defaultMessage='Output'
--- a/aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
+++ b/aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
@@ -133,11 +133,9 @@ class Mapping extends Component {
       })
     }
 
-    const hasError = (source, destination = null) => {
-      return this.props.selectedPipeline.mapping_errors.filter(
-        error => (error.path && (error.path === source || error.path === destination))
-      ).length > 0
-    }
+    const hasError = (source, destination = null) => (Boolean(this.props.selectedPipeline.mapping_errors.find(
+      error => (error.path && (error.path === source || error.path === destination))
+    )))
 
     return (
       <div key={rule.id} className={`${hasError(rule.source, rule.destination) && 'error'} rule`}>
--- a/aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
+++ b/aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
@@ -133,15 +133,21 @@ class Mapping extends Component {
       })
     }
 
+    const hasError = (source, destination = null) => {
+      return this.props.selectedPipeline.mapping_errors.filter(
+        error => (error.path && (error.path === source || error.path === destination))
+      ).length > 0
+    }
+
     return (
-      <div key={rule.id} className='rule'>
+      <div key={rule.id} className={`${hasError(rule.source, rule.destination) && 'error'} rule`}>
         <div className='rule-input source'>
           <FormattedMessage id='mapping.rule.source.placeholder' defaultMessage='define source'>
             {message => (
               <input
                 type='text'
                 required
-                className='input-d'
+                className={`${hasError(rule.source) && 'error'} input-d`}
                 name='source'
                 value={rule.source}
                 onChange={onChangeInput}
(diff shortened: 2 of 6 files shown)


sabeen:
one small thing: Can we not show the green dot in the output heading, if there are no mappings yet?

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -37,6 +37,7 @@ class Pipeline extends Component {
   }
 
   componentWillReceiveProps (nextProps) {
+    console.log(nextProps.selectedPipeline)
```
tooxie: Could we remove logging statements from production?

aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
```diff
@@ -133,15 +133,21 @@ class Mapping extends Component {
       })
     }
 
+    const hasError = (source, destination = null) => {
+      return this.props.selectedPipeline.mapping_errors.filter(
+        error => (error.path && (error.path === source || error.path === destination))
+      ).length > 0
```
tooxie: What about [find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead of `filter()`?
```js
const hasError = (source, destination = null) => {
  return Boolean(this.props.selectedPipeline.mapping_errors.find(
    (error) => (error.path && (error.path === source || error.path === destination))
  ));
};
```

aether-ui/aether/ui/assets/css/_pipeline.scss
```diff
@@ -177,6 +177,24 @@ div.pipeline-nav-item__output {
 
     transition: right $transition-speed;
   }
+
+  .status {
+    @include shadow-cutout-dark;
+    border-radius: 50%;
+    width: .8em;
+    height: .8em;
```
tooxie, line 185: I think we should stick to either `em`s or `rem`s, but not mix them both.

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -128,6 +129,9 @@ class Pipeline extends Component {
               className='pipeline-nav-item__output'
               onClick={() => this.toggleOutput()}>
               <div className='badge badge-small' />
+              { this.props.selectedPipeline.mapping.length > 0 &&
+                <span className={`status ${this.props.selectedPipeline.mapping_errors.length > 0
+                  ? 'red' : 'green'}`} /> }
```
tooxie, line 133: You could reuse `hasError` here.

tooxie (changes requested):

aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
obdulia-losantos: Regarding `find` first we need to decide if we want to support IE :see_no_evil: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
lordmallam: I rather not loop through the entire error list to determine if theres an error.

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
lordmallam: My bad...

aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
admbtlr: It would be nice to support IE if we can, but I don't think it's worth stressing over. Has anyone tested the UI in IE?

Maybe it would be good to add an explanatory comment like `Using filter rather than filter for IE support`

tooxie: Arrow functions, which we use extensively, are not supported in IE. Are we transpiling the code?

obdulia-losantos: yes, with babel.

aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
```diff
@@ -133,15 +133,21 @@ class Mapping extends Component {
...
+    }
+
     return (
-      <div key={rule.id} className='rule'>
+      <div key={rule.id} className={`${hasError(rule.source, rule.destination) && 'error'} rule`}>
```
admbtlr: There needs to be a moratorium on using $. My first thought was "what's jquery doing here?", then "why is this a Sass variable?", and let's not even mention PHP...

aether-ui/aether/ui/assets/css/_pipeline.scss
sabeen: Sometimes we want to have rem to have some kind of "fixed" unit for margins and paddings to better align items. And in some cases we use em to make it relative to the parents font-size, f.e. in buttons, where you want to adjust the padding according to the size of the button, or within text. It is not always obvious when to use what, but in this case it is related to the heading of the section, and theoretically if you decided to make that heading relative bigger or smaller, that circle indicator would grow or shrink according to the headings font size. Which is the intended behavior.

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
tooxie: Why is it ok to loop through the entire error list in `hasError` but not here? Why is here `mapping_errors.length` enough, but in `hasError` we check what the error actually has?

aether-ui/aether/ui/assets/css/_pipeline.scss
admbtlr: They mean different things. `em` is kinda locally scoped, while `rem` is global, and each one makes more sense in certain contexts.

admbtlr: OTOH when you're down to like 0.8 of something, it's probably pretty arbitrary... what do you think @sabeen ?

admbtlr: Ah, OK, you beat me to it :)

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
lordmallam: Because `hasError` is for a specific error to display an icon against the problematic mapping. In this case, i only need to know if theres an error or not (holistically)

aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
tooxie: Then we can polyfill `Array.prototype.find` as well.

aether-ui/aether/ui/assets/css/_pipeline.scss
sabeen: why, because it is below 0? It is ok, you have to make fine adjustments, and using a relation to the font size is good.

aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
obdulia-losantos: :+1:

aether-ui/aether/ui/assets/css/_pipeline.scss
admbtlr: Yes, you're right, I take back my comment about it being arbitrary :)

aether-ui/aether/ui/assets/apps/pipeline/sections/Mapping.jsx
lordmallam: Hehehehe... that means we shouldn't use template strings?

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
tooxie: Alright. You can drop `> 0` from both conditions btw.

lordmallam: If i drop it `0` gets rendered `if this.props.selectedPipeline.mapping.length === 0`

tooxie: In my opinion that's an effect of misusing a binary operator as a conditional statement (but that's something we discussed already 😄 )

tooxie (approved):

PR #208, opened 2018-06-14
test: (ui) return kernel url to ui

lordmallam:
Displays published pipeline metadata (submission url and sample data)

#### Related JIRA ticket [AET-287](https://jira.ehealthafrica.org/browse/AET-287)

### DIFF
--- a/aether-ui/aether/ui/api/tests/test_views.py
+++ b/aether-ui/aether/ui/api/tests/test_views.py
@@ -1,5 +1,6 @@
 import json
 import mock
+import os
 
 from django.contrib.auth import get_user_model
 from django.urls import reverse
@@ -206,3 +207,8 @@ class ViewsTest(TestCase):
         response = self.client.post(url, content_type='application/json')
         response_data = json.loads(response.content)
         self.assertEqual(len(response_data), 2)
+
+    def test_view_get_kernel_url(self):
+        url = reverse('kernel-url')
+        response = self.client.get(url)
+        self.assertEqual(json.loads(response.content), os.environ.get('AETHER_KERNEL_URL_TEST'))
--- a/aether-ui/aether/ui/api/urls.py
+++ b/aether-ui/aether/ui/api/urls.py
@@ -1,9 +1,13 @@
 from rest_framework import routers
+from django.conf.urls import url
 
 from . import views
-
 
 router = routers.DefaultRouter()
 router.register('pipelines', views.PipelineViewSet)
 
-urlpatterns = router.urls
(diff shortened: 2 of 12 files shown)


obdulia-losantos:
This is the way it's done in gather

https://github.com/eHealthAfrica/aether/blob/develop/aether-ui/aether/ui/webpack.common.js#L65-L70
`webpack.common.js`
```javascript
    // new plugin
    new webpack.EnvironmentPlugin({
      AETHER_KERNEL_URL: null, // this takes the env var
      // ...
    }),
```

`assets/utils/env.jsx`
```javascript

export const AETHER_KERNEL_URL = process.env.AETHER_KERNEL_URL

```

`assets/utils/env.spec.jsx`
```javascript

import assert from 'assert'

const TEST_ENV = process.env

describe('environment constants', () => {
  beforeEach(() => {
    // necessary to remove `require` cache
    jest.resetModules()

    // assign default environment variables
    process.env = Object.assign(
      {},
      process.env,
      {
        AETHER_KERNEL_URL: 'http://kernel-test:9001'
      }
    )
  })

  afterEach(() => {
    process.env = TEST_ENV
  })

  it('should take environment variables', () => {
    const envVars = require('./env')

    assert.equal(envVars.AETHER_KERNEL_URL, 'http://kernel-test:9001')
  })
})
```

With this you don't need to create a view to return the kernel server url

shawnsarwar (dismissed):
Looks good from a requirements standpoint.

lordmallam:
@obdulia-losantos I thought about that approach and decided to go with using `views` because it utilises the `get_kernel_server_url` function from the `common` module which is widely used across modules. Using the webpack method will add an extra layer of complexity in terms of code maintenance. In light of the microservices architecture, I think its best we have the business logic for determining the kernel url in one place.

quesebifurcan (changes requested):
@lordmallam Would it be possible to reuse the kernel mapping url that you get from kernel whenever you successfully post a mapping? The response will always contain the same url we are rebuilding here https://github.com/eHealthAfrica/aether/blob/d0932403fe4d93cf9f53c5c2b016df5cd76b5823/aether-ui/aether/ui/assets/apps/pipeline/InfoButton.jsx#L26. And this could be stored in the ui database, along with the other kernel references.

lordmallam:
@quesebifurcan The mapping id is already stored along with refs to other kernel artefacts. I don't think it's necessary storing the url in the db if it can be determined on the fly. Also, incase the host changes.

quesebifurcan (approved):

PR #209, opened 2018-06-14
Feat/producer health and status

shawnsarwar:
I've rewritten the Aether Producer.

1. Added /status and /healthcheck http endpoints
2. Offset is now kept on a per entity type basis
3. Entities are collected in a batch instead of individually and serialized as a single package before being sent to Kafka (This is much more efficient.)
4. Logging is much improved and generally follows DEBUG / INFO / ERROR
5. /status allows other applications to access current producer state on each topic
6. /status allows integration test to wait until kafka is up and production is done to assert final values
7. Used gevent to allow threadlike design without significant overhead (30% CPU / 50mb RAM at full load)
8. Offsets and now kept in a sqlite db at a URL defining in the settings file instead of flat file.


aether-producer/producer/aether_producer.py
```diff
@@ -16,132 +16,263 @@
...
+    def register(self, route_name, fn):
+        self.app.add_url_rule('/%s' % route_name, route_name, view_func=fn)
+
+    def serve(self):
+        self.app = Flask(__name__)  # pylint: disable=invalid-name
+        self.logger = self.app.logger
+        log_level = logging.getLevelName(
+            self.settings.get('log_level', 'DEBUG'))
+        self.logger.setLevel(log_level)
+        self.app.debug = True
```
tooxie: This looks like leftover.

aether-producer/producer/db.py
```diff
@@ -0,0 +1,119 @@
...
+
+def make_uuid():
+    return str(uuid.uuid4())
+
+
+class Offset(Base):
+    __tablename__ = 'offset'
+
+    schema_name = Column(String, primary_key=True)
+    offset_value = Column(String, nullable=False)
```
tooxie, line 69: Why do we need a db for this? We are not keeping any history, nor is the data being queried (besides just fetching the only record in the DB).

What happens when this server crashes and we have to recreate the instance? Will it re-send the complete DB to kafka again?

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +288,255 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
+            else:
+                raise ValueError('No entry for %s' % self.name)
+        except Exception as err:
+            self.logger.error('Could not get offset for %s it is a new type' % (self.name))
+            return "" # No valid offset so return empty string which is < any value
+
+    def set_offset(self, offset):
+        # Set a new offset in the database
+        new_offset = Offset.update(self.name, offset)
+        if not new_offset:
```
tooxie: If I understand that correctly, `update()` will create the offset if it doesn't find it. So this condition will never be true.

shawnsarwar:
It might be helpful to read about gevent before reviewing the PR as I used it's patterns pretty extensively.

http://sdiehl.github.io/gevent-tutorial/

aether-producer/producer/aether_producer.py
```diff
@@ -16,132 +16,263 @@
...
+        self.logger.debug('No longer attempting to connect to Aether')
+
+    # Main Kafka / Schema Loop
+    def check_schemas(self):
+        # Checks for schemas in Kernel
+        # Creates SchemaHandlers for found schemas.
+        # Updates SchemaHandler.schema on schema change
+        while not self.killed:
+            if not self.kernel:
+                sleep(1)
```
tooxie: If `self.kernel` is `None`, it will sleep for 1 second and go back into the `while not self.killed` loop, and so on forever. When will `self.kernel` get set? What am I missing here? Is this being set in the background by some other concurrent thread?

aether-producer/producer/aether_producer.py
```diff
@@ -16,132 +16,263 @@
...
+class KernelHandler(object):
+    # Context for Kernel, which nulls kernel for reconnection if an error occurs
+    # Errors of type ignored_exceptions are logged but don't null kernel
+    def __init__(self, handler, ignored_exceptions=None):
+        self.handler = handler
+        self.log = self.handler.logger
+        if ignored_exceptions is not None and isinstance(ignored_exceptions, list) is not True:
+            self.ignored_exceptions = [ignored_exceptions]
+        elif ignored_exceptions is not None:
+            self.ignored_exceptions = ignored_exceptions
```
tooxie: This case is not used, I would simplify and assume it will get a list.

aether-producer/producer/aether_producer.py
```diff
@@ -16,132 +16,263 @@
...
+        # Updates SchemaHandler.schema on schema change
+        while not self.killed:
+            if not self.kernel:
+                sleep(1)
+            else:
+                schemas = []
+                with KernelHandler(self):
+                    self.kernel.refresh()
+                    schemas = [
+                        schema for schema in self.kernel.Resource.Schema]
```
tooxie, line 206: What type is `self.kernel.Resource.Schema`? Is it [DottedDict](https://github.com/eHealthAfrica/aether/blob/315f95170f4a8ed3ff912d53f13a65a3740d9711/aether-utils/aether-client/aether/client.py#L47)? If so you could just call `self.kernel.Resource.Schema.keys()`.

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +288,255 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
+
+                self.logger.debug('All changes saved ok.')
+                break
+
+            # Remove successful and errored changes
+            for _id, err in self.failed_changes.items():
+                try:
+                    del self.change_set[_id]
+                except KeyError as ke:
+                    pass  # could have been removed on previous iter
```
tooxie, line 525: How is this possible? `self.failed_changes` is a dictionary, it can't contain duplicated items.

aether-producer/producer/aether_producer.py
```diff
@@ -16,132 +16,263 @@
...
+    def add_endpoints(self):
+        # URLS configured here
+        self.register('healthcheck', self.request_healthcheck)
+        self.register('status', self.request_status)
+
+    def register(self, route_name, fn):
+        self.app.add_url_rule('/%s' % route_name, route_name, view_func=fn)
+
+    def serve(self):
+        self.app = Flask(__name__)  # pylint: disable=invalid-name
```
tooxie, line 237: De we really need Flask at all? This not really a web app, more than a job that queries the DB and posts to kafka.

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +288,255 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
-    finally:
-        if manager:
-            manager.stop()
-
-
-if __name__ == "__main__":
-    main_loop()
+def main(test=False):
+    settings = Settings(test=test)
+    handler = ServerHandler(settings)
```
tooxie: This doesn't look like a class, in the sense that once you instantiate it, it just does everything. There is no extensibility or a clear API that encapsulates the logic away.

aether-producer/producer/aether_producer.py
```diff
@@ -16,132 +16,263 @@
...
+                for schema in schemas:
+                    if not schema.name in self.schema_handlers.keys():
+                        self.logger.info(
+                            "New topic connected: %s" % schema.name)
+                        self.schema_handlers[schema.name] = SchemaHandler(
+                            self, schema)
+                    else:
+                        res = self.schema_handlers[schema.name]
+                        if res.schema_obj != res.parse_schema(schema):
+                            res.update_schema(schema)
```
tooxie: Since `update_schema` calls `parse_schema` it would simplify the code (and be more performant) to just do the update every time, right?

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +288,255 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
+        self.failed_changes = {}
+        self.pg_creds = self.context.settings.get('postgres_connection_info')
+        self.kafka_failure_wait_time = self.context.settings.get('kafka_failure_wait_time', 10)
+        try:
+            self.topic_name = self.context.settings \
+                .get('topic_settings', {}) \
+                .get('name_modifier', "%s") \
+                % self.name
+        except Exception as err:  # Bad Name
+            self.topic_name = self.name
```
tooxie: It would be worth notifying that the expected name is not valid.

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +288,255 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
-    def stop(self):
-        for name, stream in self.streams.items():
-            stream.stop()
-            print ("released connection to topic: %s" % name)
-        self.streams = {}
-        print ("manager stopped")
+            except Exception as ke:
+                self.logger.error('error in Kafka save: %s' % ke)
+                self.logger.error(traceback.format_exc())
+                sleep(1)
```
tooxie: As I commented in another line, if we swallow the exception and silently log the error we will only realize that something is not working once we notice that the data is too old.

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +288,255 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
-            bytes_writer = io.BytesIO()
-            writer = DataFileWriter(bytes_writer, DatumWriter(), self.schema, codec='deflate')
-            writer.append(msg)
-            writer.flush()
-            raw_bytes = bytes_writer.getvalue()
-            writer.close()
-            future = self.producer.send(self.topic, key=str(msg.get("id")), value=raw_bytes)
-            #block until it actually sends. We don't want offsets getting out of sync
+            obj = io.BytesIO()
+            if err:
```
tooxie: It seems to me that these 2 branches are identical and they could be unified:
```py
def kafka_callback(self, err=None, msg=None, **kwargs):
    if err:
        self.logger.error('ERROR %s', [err, msg, kwargs])

    with io.BytesIO() as obj:
        obj.write(msg.value())
        reader = DataFileReader(obj, DatumReader())

    for x, message in enumerate(reader):
        _id = message.get("id")
        if err:
            self.logger.debug("NO-SAVE: %s in topic %s" % (_id, self.topic_name))
            self.failed_changes[_id] = err
        else:
            self.logger.debug("SAVE: %s in topic %s" % (_id, self.topic_name))
            self.successful_changes.append(_id)
```

Note:
* Dropped `_` parameter (the docs only show the first 2 args, I couldn't find what it was and we don't use it)
* Introduced `with` statement
* Removed `try/except` block
* Turned topic logs into `debug`

These are just some ideas, specially the `try` block removal, it's up for discussion. My concern is that we don't really know what we are catching there, which means it could be hiding important errors.

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +288,255 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
-        valid = Validate(self.schema, msg)
-        if not valid:
-            raise ValueError("message doesn't adhere to schema \n%s\n%s" % (json.dumps(self.schema), json.dumps(msg)))
+        '''
+
+    def __init__(self, server_handler, schema):
+        self.context = server_handler
+        self.logger = self.context.logger
+        self.name = schema.name
+        self.modified = ""
```
tooxie: I'd rename this to `offset` for consistency.

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +288,255 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
+        try:
+            offset = Offset.get_offset(self.name)
+            if offset:
+                self.logger.debug("Got offset for %s | %s" % (self.name, offset.offset_value))
+                return offset.offset_value
+            else:
+                raise ValueError('No entry for %s' % self.name)
+        except Exception as err:
+            self.logger.error('Could not get offset for %s it is a new type' % (self.name))
+            return "" # No valid offset so return empty string which is < any value
```
tooxie: It took me a lot of digging into the code to figure out what happened in the first run, when there was no data in the database. This comment was extremely helpful :) but I think we should make this more explicit in the code, because just the comment it's not enough to understand what it means.

An idea:
```py
# pseudocode
offset = get_offset()

if offset:
    QUERY += 'e.modified > %s' % offset
```

aether-producer/producer/aether_producer.py
```diff
@@ -16,132 +16,263 @@
...
+        self.threads = []
+        self.threads.append(gevent.spawn(self.keep_alive_loop))
+        self.threads.append(gevent.spawn(self.connect_aether))
+        self.threads.append(gevent.spawn(self.check_schemas))
+        gevent.joinall(self.threads)
+
+    def update_schema(self, name):
+        pass
 
+    def kill(self, *args, **kwargs):
```
tooxie: Who calls this?

tooxie (changes requested):
This is cool, I'm excited to see async execution being introduced here. It's something that we could use in some other parts of the codebase eventually. As usual, concurrency comes at a cost: readability. I find it quite difficult to follow the execution path and to know what runs when, and who modifies what (see the comment about `self.kernel` in line 194 for example). I would question if it's necessary to run everything concurrently (or even anything) since this service doesn't have to answer other requests, just process one thing (or 1 batch) at a time. Besides, if the request takes long it will simply batch all the entries since `offset`, so nothing gets lost.

On top of that extra complexity inherent to concurrency, business logic and retry logic seems to be intertwined:

```py
# pseudocode
while not self.killed:
    # retry logic
    if not self.available:
        sleep(1)
        continue

    # business logic here
```

This is a recurrent pattern. If we want to keep the concurrent model, then I would prefer that we have a setup phase where we check that we have all the available resources to work (are all the services reachable?) and then have something like an operator that spins off the threads. That way the business logic behaves as single threaded/synchronous, and we keep the concurrency logic out, in a separate module.

Something else that I noticed is that try/except blocks are quite big, they catch every Exception, and just log and continue. This is very dangerous because in production we won't log anything under `error` level, and we won't notice that data is not being sent to kafka. Speaking of which, we should have a clear guidance on when a log message is *warn*, *debug*, *info*, etc...

One last question: Have you considered having this functionality as part of `aether-kernel`? Problems like the cold start would disappear if we kept the offset in the kernel's database.

Nice PR! :)

aether-producer/producer/aether_producer.py
shawnsarwar: app.debug is left over. Setting log level via the config is something we want to keep.

aether-producer/producer/db.py
shawnsarwar: Generally the does the following.
a) look at Kernel for a list of current schemas.
b) connect to a topic on Kafka for each schema type.
c) Access the kernel postgres and send data of each type to the right topic.

The producer needs to keep track of which data has been saved successfully in Kafka. We get data from PostgreSQL in the order in which it was saved via the modified value. For this to work efficiently, we pull a batch of changes of size n from PG.  It's ordered and changes[n-1].modified < changes[n].modified. So if everything in the batch is saved ( or there are errors < n ), the new query should start have a value > previous changes[n].modified.

In the producer we call the .modified value an offset, matching the Kafka term for a similar mechanism. We expect he producer will at some point go down while data is still being saved in Kernel. If that happens, we need to be able to pick up from the last saved offset in Kafka, ( which isn't a value that is saved in Kafka because it describes when the record was inserted into PG, not anything inherent about the data itself.). We used to save that information in a flat file. However, we want this to be something we can easily migrate and upgrade, so I wanted to move all persistence out of the application container. Hence SQLite doing a job that could be accomplished by a 0.1kb text file.

aether-producer/producer/aether_producer.py
shawnsarwar: Good catch, that changed mid write.

aether-producer/producer/aether_producer.py
shawnsarwar: Whenever we access kernel, we wrap it in a contextmanger created by KernelHandler (KH)
https://github.com/eHealthAfrica/aether/blob/a335d8f9178a6ba539c6029453e5d16b9955d1e8/aether-producer/producer/aether_producer.py#L78

It's the job of KH to catch exceptions in the Kernel operation and if they're unexpected to kill the kernel by setting kernel = None. So any failure of the kernel connection will see it nulled.

We then have a greenlet constantly running on ServerHandler.connect_aether()
https://github.com/eHealthAfrica/aether/blob/a335d8f9178a6ba539c6029453e5d16b9955d1e8/aether-producer/producer/aether_producer.py#L166
connect_aether is constantly looking at the kernel connection to see if it's down. If it is, then it tries to connect to it.

Other things that need the kernel need to do two things, check to see if it's up so as to not waste their time, and use KH as a context manager for their connections to it. This allows kernel start order to have no effect on the producer. It can be up or down or somewhere in between and the producer acts properly.

aether-producer/producer/aether_producer.py
shawnsarwar: yeah, Schema.keys() would be the same and more idiomatic.

aether-producer/producer/aether_producer.py
shawnsarwar: Yeah, that was a bug somewhere else. I forgot to clean this up.

aether-producer/producer/aether_producer.py
shawnsarwar: Oh, it's also a web app. We can provide current status of the producer to other applications via the /status and /healthcheck endpoints. I hope to get this information into the UI eventually so people can watch their data flow.
I also already added /status usage to the integration tests to fix a race condition where the tests were expecting Kafka to be up and it wasn't ready yet.

aether-producer/producer/aether_producer.py
shawnsarwar: Settings could be something more simple. It's really just a dictionary that know something about how to preload itself.

aether-producer/producer/aether_producer.py
shawnsarwar: I don't think there would be any side effects so maybe. Parse schema only does have the job required though. Update schema also turns schema into a Spavro object so I think this is still slightly more performant but not in any important way.

aether-producer/producer/aether_producer.py
shawnsarwar: Good point.

aether-producer/producer/aether_producer.py
shawnsarwar: Ah, that's a nice refactor. I think the enumeration of reader needs to happen inside the context but I'm not positive.

aether-producer/producer/aether_producer.py
shawnsarwar: Agreed.

aether-producer/producer/aether_producer.py
shawnsarwar: Good idea.

aether-producer/producer/aether_producer.py
shawnsarwar: This is registered to signal.SIGTERM and SIGINT. Anything with system process control can call this with a kill command (including docker)

shawnsarwar:
Very nice code review @tooxie 

I'm pretty married to the concurrent approach as we may have a large number of topics to monitor on a given instance, but points well taken.

aether-producer/producer/aether_producer.py
shawnsarwar: Wait, that's not a bug. If not all changes are processed in the first run, then on a subsequent iteration successful changes will be longer but still contain the first set of successful changes and we'll get a key error when we try to remove it again from change_set as we did that on the first iteration.

aether-producer/producer/aether_producer.py
shawnsarwar: This works now so I'm not going to address it as it's fairly minor and is only slightly gross.

aether-producer/producer/aether_producer.py
shawnsarwar: Something going wrong in this try block is likely due to systemic failure. I don't know how to scream more loudly than logger.error(). It would also be apparent from the /status endpoint that data is no longer being saved to Kafka. The other options would be killing the producer, but if it's a data issue causing this operation to fail, that would take down other topics for no good reason.

aether-producer/producer/aether_producer.py
tooxie: Yes, sorry, I meant the `debug` part only.

aether-producer/producer/db.py
tooxie: So, if I understand your answer correctly, we want to turn this into its own DB? That will incur in costs (both in infrastructure and in maintenance) for a full blown DB containing a handful of registers in one single table. This looks more like a cache than a DB to be honest.

shawnsarwar: I'm flexible on the what but I do think we need a resource external to the application container. This could be a shared resource as I assume other applications that need to keep minimal state might also benefit. Not really anything on the consumer side as Kafka does that for you, but other producers / stream processors might need to save some information and Kafka may not always be ideal.
Alpine Sqlite looks pretty lightweight.
https://github.com/jansanchez/sqlite-alpine

aether-producer/producer/aether_producer.py
tooxie: How is that possible? We get the new entries that need to be published:

https://github.com/eHealthAfrica/aether/blob/f3c7e296f62ace74e55abc4573a9887505417887/aether-producer/producer/aether_producer.py#L381

Then we add them to `change_set`:

https://github.com/eHealthAfrica/aether/blob/f3c7e296f62ace74e55abc4573a9887505417887/aether-producer/producer/aether_producer.py#L388-L391

And finally we send them to kafka and wait for the reply:

https://github.com/eHealthAfrica/aether/blob/f3c7e296f62ace74e55abc4573a9887505417887/aether-producer/producer/aether_producer.py#L398-L404

Why do subsequent runs get mixed with previous updates? Why do we keep `change_set` at all? (It looks like the only thing we do with it is count it)

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +283,248 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
+    def wait_for_kafka(self, end_offset, timeout=10, iters_per_sec=10, failure_wait_time=10):
+        # Waits for confirmation of message receipt from Kafka before moving to next changeset
+        # Logs errors and status to log and to web interface
+
+        sleep_time = timeout / (timeout * iters_per_sec)
+        change_set_size = len(self.change_set)
+        errors = {}
+        for i in range(timeout * iters_per_sec):
+
+            # whole changeset failed; systemic failure likely; sleep it off and try again
```
tooxie, line 503: I don't think that we should *sleep a systemic failure off*, this looks like something that we should be waking people up to look into.

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +283,248 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
+        # Timeout reached or all messages returned ( and not all failed )
+
+        self.status["last_changeset_status"] = {
+                    "changes": change_set_size,
+                    "failed": len(self.failed_changes),
+                    "succeeded": len(self.successful_changes),
+                    "timestamp": datetime.now().isoformat()
+        }
+        if errors:
+            self.handle_kafka_errors(change_set_size, all_failed=False)
```
tooxie: If I understand this correctly, when something was not processed we just log it in move on, writing a new offset. Does that mean that failed entries will not be re-sent? How do we know that one (or more) entries are failing once and again?

aether-producer/producer/aether_producer.py
```diff
@@ -157,135 +283,248 @@ def get_entities(offset = None, max_size=1000):  # TODO implement batch pull by
...
+        self.clear_changeset()
+        # Once we're satisfied, we set the new offset past the processed messages
+        self.context.kafka = True
```
tooxie, line 550: The function `handle_kafka_errors` called a couple of lines above sets `self.context.kafka` to `False` in case everything failed, but immediately here we set it back to `True`. What's the point of setting it to `False` then if we don't do anything with it?

tooxie (dismissed):

aether-producer/producer/aether_producer.py
shawnsarwar: Generally the systemic failure is just the Kafka container not being ready. Unfortunately within the Confluent Kafka library there's no good way to tell if the brokers are available. It just fails message submission with ERROR_LOCAL_TIMEOUT. This could be a temporary or permanent state, but it's apparent both in the log as an error and in the /status endpoint for anyone watching. If it resolves then the newest changeset timestamp will be past the failure timestamp.

aether-producer/producer/aether_producer.py
shawnsarwar: Messages as a rule shouldn't fail (outside of Kafka not being ready), so anytime one does it's a big deal. It would likely be a schema validation error where a schema has been updated but the producer wasn't running at the time and has old messages waiting submission that adhere to and old schema. There's not a good way to handle it yet because we don't have schema revisions in Kernel yet. If a message fails for that reason there's likely nothing we can do here except log it.

That's why we move on to the next offset if any of the messages get through (telling us that Kafka is working properly)

aether-producer/producer/aether_producer.py
shawnsarwar: context.Kafka is purely a status flag for reporting in the /status endpoint so other systems know if Kafka should be available since this interacts with it so often.

aether-producer/producer/aether_producer.py
shawnsarwar: Right now we just keep the changeset to make sure that kafka has acknowledged receipt one way or another; if it hasn't we loop until we hit the timeout. If you up the postgres_pull_limit you can get cases where messages are still enqueued in the local producer object for Kafka but haven't been submitted before the timeout hits. That's why we have the outer timeout loop within wait_for_kafka(). If messages are getting stuck in flight ( in change_set ) for long periods, you probably need to tune your kafka settings.

aether-producer/producer/db.py
tooxie: Yes, we agree that it has to go out of the app container. My question is why is this a completely separate module.

This is why I asked if we have considered integrating this with the main django app. We could reuse the DB (which is coupled anyway, we query it through SQL, i.e. we know the schema, and if that changes this module breaks). Django provides mechanisms to modularize and decouple components like these (middlewares, backends, etc)

How do you picture *sqlite-alpine* being used? If we keep them in the same instance then there is no difference from having a sqlite file in the filesystem. If we deploy it to its own instance then we just moved the problem to a different host but we didn't solve it.

aether-producer/producer/aether_producer.py
tooxie: I like the idea of showing this info in the UI, but we don't do it now. I would vote for removing it and adding it back in the future. More code means more code to maintain and more potential for bugs.

Besides, why flask? Right now it's just a few lines of code, but what happens when we want to add some kind of rest api instead of just a dummy endpoint? Will we have to maintain 2 different rest frameworks simultaneously?

Don't get me wrong, I love flask, I would drop django tomorrow and rewrite everything in flask :) but I don't think we should introduce a new framework without discussing it with the team first. After all, we all have to maintain it afterwards.

aether-producer/producer/aether_producer.py
tooxie: At least we should be able to discriminate between brokers not being ready (all good, just retry) and nothing works (call the firemen). Otherwise how can we add alarms to notify of problems? I don't think that someone should be watching in order to find problems, there are too many parts in the system, and nobody will be monitoring kafka logs.

aether-producer/producer/aether_producer.py
tooxie: My point is that it looks to me that it will always be `True` (for practical purposes). Yes, it gets set to false by `handle_kafka_errors`, but it immediately returns, calls `clear_changeset()`, and it gets set back to `True` in `wait_for_kafka`:

So what happens is:
```py
if all_failed:
    self.context.kafka = False

self.failed_changes = {}
self.successful_changes = []
self.change_set = {}

self.context.kafka = True
```

Once `self.context.kafka` is set to `True`, the endpoint will never show it as `False` again, even if it keeps failing. Right? Am I missing something?

shawnsarwar: Ah, I understand. 

Everytime a batch comes back with all messages having failed we enter this block.
https://github.com/eHealthAfrica/aether/blob/019bb0dfb90553f8aa02c43439e373430abbc1c2/aether-producer/producer/aether_producer.py#L421

It then triggers  self.handle_kafka_errors(all_failed=True) which on all_failed sets context.kafka to false. 
The failed block then returns wait_for_kafka without exiting the timeout loop (line 428)
`                return  # all failed; ignore changeset`
since all messages were resolved as failed and we don't want to update the offset and kafka =True is never called.

The the only place that context.kafka = False happens is when a whole changeset fails or on startup. context.kafka becomes true when any part of a changeset is successfully sent to kafka.

aether-producer/producer/aether_producer.py
shawnsarwar: That's a great question. We could have some data loss threshold ( possibly < 0 ) where we send a high level alert. That would coincide with this line having an output:
https://github.com/eHealthAfrica/aether/blob/019bb0dfb90553f8aa02c43439e373430abbc1c2/aether-producer/producer/aether_producer.py#L487

This is a case where Kafka is handling some messages so we need to increment the offset, but something special went wrong with that particular message. This is a noteworthy error that wouldn't be detected somewhere else. ( A consumer, or in Kafka directly )

What's the best way to express the direness of this issue without dropping other errors to below `logLevel.ERROR`?

shawnsarwar: We're using logLevel.critical for anything that requires attention, such as dropped messages that couldn't be recovered.

PR #216, opened 2018-06-18
WIP --do not merge | Feat/elasticsearch consumer

shawnsarwar:
I've modified the toy elasticsearch consumer to work with our current setup. The consumer needs some refactor, but it's now based on the consumer-sdk. If you need ES, this should work well for now. Take a look at the sample index in the /index folder for supported syntax. Currently it handles parent child relationships and creation of a location geojson object from existing fields. (It MUST be named location.)


shawnsarwar:
Before running create an ES index in the /index sub folder for the module. If you create a geolocation in the index, the consumer will attempt to find matching lat and long fields in the data. The names of the types in the index correspond to kafka topics.

Once you're setup, run the docker-compose-es.yml file from the base Aether folder. This expects docker-compose.yml and docker-compose-connect.yml to already be running.

consumer-elasticsearch/aether_elasticsearch/elasticsearch_consumer.py
```diff
@@ -0,0 +1,380 @@
...
+    connect_es()
+
+
+def connect_es():
+    for x in range(CONN_RETRY):
+        try:
+            global es
+            # default connection on localhost
+            es = Elasticsearch(
+                ["localhost", "elasticsearch"], sniff_on_start=False)
```
quesebifurcan, line 40: Would it make sense to be able to set these values from a config file/environment variables? Thinking of Kubernetes, where this consumer might run in a different pod than the es instance itself (and would be referred to by its hostname).

consumer-elasticsearch/aether_elasticsearch/elasticsearch_consumer.py
```diff
@@ -0,0 +1,380 @@
...
+            print (es.info())
+            return
+        except TransportError as ese:
+            print("Could not connect to Elasticsearch Instance")
```
quesebifurcan, line 44: Many print statements in this file could be turned into logging statements. Stdout wouldn't be so cluttered and devops can filter messages by severity.

consumer-elasticsearch/aether_elasticsearch/elasticsearch_consumer.py
```diff
@@ -0,0 +1,380 @@
...
+            index_files = os.listdir(index_path)
+            for index_file in index_files:
+                self.register_index(index_path, index_file)
+
+    def register_index(self, index_path, index_file):
+        index_name = index_file.split(".")[0]
+        data = None
+        path = "%s/%s" % (index_path, index_file)
+        with open(path) as f:
+            data = json.load(f)
```
quesebifurcan, line 90: When testing different indices, I got a lot of errors from here. The reason was always a json error in my hand-edited file, but it wasn't so obvious that that was actually the error. Something about "Malformed config file" would be great to have.

consumer-elasticsearch/aether_elasticsearch/elasticsearch_consumer.py
```diff
@@ -0,0 +1,380 @@
...
+        return res
+
+
+def main_loop():
+    connect()
+    manager = ESConsumerManager()
+    print ("Started!")
+    while True:
+        try:
+            pass
```
quesebifurcan, line 367: `pass` is probably a leftover from a previous version?

consumer-elasticsearch/aether_elasticsearch/elasticsearch_consumer.py
```diff
@@ -0,0 +1,380 @@
...
+                return doc[name]
+            else:
+                # Looks for Nested Keys
+                for key in doc.keys():
+                    val = doc.get(key)
+                    try:
+                        if name in val.keys():
+                            return val.get(name)
+                    except Exception as err:
+                        pass
```
quesebifurcan, line 327: It looks like this try/except block could be replaced by `val.get(name, None)` or `val.get(name)`?

quesebifurcan (changes requested):
Nice! I tested a couple of entities/indices and it seems to work well. Nice to see more specific exceptions in the code.

Some minor changes suggested below.

shawnsarwar:
This has been moved out of aether into aether-elastic-consumer

PR #217, opened 2018-06-19
feat: avro validation with error accumulation

quesebifurcan:
Improve validation of avro schemas -- accumulate a list of errors
encountered during validation instead of just returning a boolean.

Background: when attempting to validate a complex form in the
aether ui, any mapping which produces an entity with type errors
will result in a single generic error. This becomes a problem when
a form has more than one field -- the user does not get any hints as 
to what went wrong (what type was expected? what was the actual 
value? which part of the schema did not validate?).

The implementation of AvroValidator is based on the official
avro.io.validate() function, which can be found here:
https://github.com/apache/avro/blob/b71dcf24f252da5858bd95bfd4bd56402e3a458c/lang/py3/avro/io.py#L96-L142.

avro.io.validate() returns a boolean which
indicates the outcome of the validation. No error details
are returned. In contrast, AvroValidator will accumulate a
list of all errors encountered during validation. The type
of AvroValidator.errors is a list of AvroValidationErrors.

Performance: running AvroValidator.validate() will be about
twice as slow as avro.io.validate().
avro.io.validate() will check 10000 datums against a
moderately complex schema in ~0.5 seconds. AvroValidator
will finish in ~1 second. If this is not acceptable, we could
add a full_validation=<boolean> flag in kernel, the value of
which would only be true when called from the ui (where
0.0001 seconds definitely would not make a difference).

Note: this test https://github.com/eHealthAfrica/aether/blob/3302b3a150afa168d051b2396d43b642c56a648d/aether-kernel/aether/kernel/api/tests/test_avro_tools.py#L464-L535 gives a good overview of most features and is probably a good place to start a review.


aether-kernel/aether/kernel/api/avro_tools.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+class AvroValidationException(Exception):
+    pass
+
+
+# This namedtuple represents an avro validation error.
+# Fields:
+#     - expected: the expected type.
+#     - datum: the actual type.
```
admbtlr: You mean "actual _value_", right?

aether-kernel/aether/kernel/api/tests/test_avro_tools.py
```diff
@@ -0,0 +1,561 @@
...
+            )
+        ])
+
+    def test_validate_long(self):
+        self.run_validation_tests([
+            AvroValidatorTest(
+                fields=[{'name': 'test', 'type': 'long'}],
+                datum={'test': 999999999999999999},
+                expected_errors=[],
+            )
```
admbtlr: Where's `not-a-long` gone? 😢

admbtlr (changes requested):
Nice!

aether-kernel/aether/kernel/api/tests/test_avro_tools.py
quesebifurcan: ah, good catch!

admbtlr (approved):

PR #218, opened 2018-06-20
feat: validation and error handling

quesebifurcan:
Most of these changes are intended to improve schema and
mapping validation and make some of the implicit
requirements for data types in json fields more explicit.

Decouple Entity.id from Entity.payload.id -- up until now,
it was an implicit requirement that they should match. We
should not let user-submitted ids act as primary keys for
our tables.

Add validation for schema and mapping definitions.

Assert that Entity.payload.id is a uuid and that it does not
equal Entity.id.

Add a validators.py module containing serializer validators
and a jsonschema for the mapping definition.

When deriving entity types from sample json data, do not
make the top-level id field optional.

Use `TestCase` instead of `TransactionTestCase` to speed up
filter tests.

Depends on: https://github.com/eHealthAfrica/aether/pull/217 (which needs to be merged first)
Resolves https://jira.ehealthafrica.org/browse/AET-276
Discussion: https://confluence.ehealthafrica.org/pages/viewpage.action?pageId=17072168


admbtlr:
@quesebifurcan now that #217 is merged, can you merge develop into this so that it's easier to review?

lordmallam (dismissed):
LGTM

aether-ui/aether/ui/api/utils.py
```diff
@@ -119,27 +122,36 @@ def validate_pipeline(pipeline):
...
-            [{'description': f'It was not possible to validate the pipeline: {str(e)}'}],
-            []
-        )
+    kernel_url = utils.get_kernel_server_url()
+    resp = requests.post(
+        url=f'{kernel_url}/validate-mappings/',
+        json=json.loads(json.dumps(payload)),
+        headers=utils.get_auth_header(),
+    )
+    # If everything went well, map 'mapping_errors' to 'errors' and 'entities'
```
admbtlr: This is a pretty redundant comment - I think comments like this tend to be counterproductive, since they make the reader wonder whether she's missing something that would render the comment important

aether-ui/aether/ui/assets/apps/pipeline/sections/Input.jsx
```diff
@@ -51,6 +51,9 @@ export const makeOptionalType = (type) => {
 }
 
 export const makeOptionalField = (field) => {
+  // The top-level "id" field is reserved for unique ids; do not make it
```
admbtlr, line 54: This is much better - "why" comments are always preferable to "what", unless the code is severely obfuscated

admbtlr (changes requested):

admbtlr (approved):

PR #221, opened 2018-06-21
Add timeout parameter to client

tooxie:
It's important that no calls to any service leave the system
unresponsive, consuming resources for a long period. Timeouts by default
are counted in minutes, which is ridiculous for any web app.

If any call took longer a few seconds then it must be looked into, with
long timeouts we won't even know that there is a problem.

**Note:** This PR adds no tests. I tried but since these are not unit but integration tests, I couldn't get them to run locally in a short period of time, so I will wait for someone to help me set it up.


aether-utils/aether-client/aether/client.py
```diff
@@ -62,7 +58,7 @@ def delete(self, url):
         }
 
     def get(self, url):
-        req = requests.get(url, auth=self.auth)
+        req = requests.get(url, auth=self.auth, timeout=getattr(self, 'timeout'))
```
quesebifurcan, line 61: Just a small detail -- why `getattr(self, 'timeout')` instead of `self.timeout`?

aether-utils/aether-client/aether/client.py
```diff
@@ -31,16 +31,12 @@ class GenericClient(object):
     _holds_resources = {}
     _data_types = {}
 
-    def __init__(self, url=None, **credentials):
+    def __init__(self, url=None, username=None, password=None, timeout=3):
```
quesebifurcan, line 34: Good catch! How about including timeouts for `requests.post` as well?

quesebifurcan (changes requested):

aether-utils/aether-client/aether/client.py
tooxie: Because this class is mostly used as base class for other to extend from, which means that `__init__()` can be overwritten and `self.timeout` never defined. With `getattr` we get `None` instead of an `AttributeError` exception.

quesebifurcan (approved):

PR #222, opened 2018-06-21
AUX - Kong Integration -- Do not merge -- Kong integration to kubernetes pending

lordmallam:
Kong ==> Develop

I think we need to review the requirements of Kong, because the more i think of it, the less i think we actually need kong


shawnsarwar:
I'm closing this as it doesn't align with the current MT strategy.

PR #225, opened 2018-06-24
Delete *.whl files

tooxie:
These are autogenerated binary files that should not be committed in the
repository. They get recreated at build time, no need to have them here.

Wheel files are the #1 source of merge/rebase conflicts, and since they
are binary it's impossible to know which one is the most up-to-date (if
any, most likely they will both contain necessary changes, so they have
to be recreated once again after the conflict is resolved).

There is only 1 module left as whl files, and it's the
`django_ums_client`. This is because this module is in an external
private repository, and getting pip to fetch it from travis is not
trivial. However, because it's in a different repository, this file
doesn't suffer from the same problems as all the others, so it's fine to
keep it here.


lordmallam (dismissed):
👍 Much needed

quesebifurcan:
Good idea, agree that it would be very nice to have less conflicts.
The `docker-compose up` and `docker-compose build` commands that we use for local development don't work any more:
```
  Could not find a version that satisfies the requirement aether.common==0.0.0 (from -r /code/conf/pip/requirements.txt (line 12)) (from versions: )
No matching distribution found for aether.common==0.0.0 (from -r /code/conf/pip/requirements.txt (line 12))
```
Could you update the README to include the extra commands we need prior to `docker-compose up`?
Alternatively, we could consider uncommenting these blocks https://github.com/eHealthAfrica/aether/blob/develop/docker-compose.yml#L53-L55 in `docker-compose.yml`.

tooxie:
@quesebifurcan good point. I added it to the README file's installation instructions.

quesebifurcan (approved):

PR #227, opened 2018-06-25
fix(k8s): slightly safer defaults for values.yaml files

quesebifurcan:
Change default entrypoint command for kernel- and odk-
containers from `start_dev` to `start` to avoid accidentally
running a development server in production.

Add `start_dev` to override files used for local development.


admbtlr (approved):
I know this is a WIP, but it LGTM anyway :)

PR #229, opened 2018-06-27
Aet ui paste mapping

lordmallam:
Allows user to paste mapping json

Styling can be enhanced 😬 

Related JIRA ticket: [AET-251](https://jira.ehealthafrica.org/browse/AET-251)


shawnsarwar (approved):
Looking good.

PR #231, opened 2018-06-27
fix: improve test cleanup

quesebifurcan:
Do not delete all kernel mappings before every test.
Move cleanup to `tearDown()` and only delete the mapping
which was created by the test.


shawnsarwar (dismissed):
Smart change. Do we need to worry about failure during or before teardown?

quesebifurcan:
Yes, good point -- the tests might fail during cleanup/teardown and leave a kernel mapping behind. This change just decreases the likelihood that user data is affected.

obdulia-losantos (dismissed):

obdulia-losantos (dismissed):

obdulia-losantos (approved):

admbtlr (approved):

PR #232, opened 2018-06-27
Fix/validate ids in union schemas

shawnsarwar:
Added the ability for the _id validator to handle union avro types.
Added a test reflecting this ability.

### DIFF
--- a/aether-kernel/aether/kernel/api/validators.py
+++ b/aether-kernel/aether/kernel/api/validators.py
@@ -95,6 +95,7 @@ def _has_valid_id_field(schema):
     else:
         schemas = schema
     for schema in schemas:
+        id_field_type = None
         for field in schema.get('fields', []):
             if field.get('name', None) == 'id':
                 id_field_type = field.get('type', None)
@@ -102,7 +103,7 @@ def _has_valid_id_field(schema):
 
         schema_statuses.append(id_field_type and id_field_type == 'string')
     # True if All of the statuses were True
-    return True in schema_statuses and False not in schema_statuses
+    return all(schema_statuses)
 
 
 def validate_id_field(schema):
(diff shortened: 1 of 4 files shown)


aether-kernel/aether/kernel/api/validators.py
```diff
@@ -86,13 +86,23 @@ def validate_avro_schema(value):
...
+        schemas = [schema]
+    else:
+        schemas = schema
+    for schema in schemas:
+        for field in schema.get('fields', []):
+            if field.get('name', None) == 'id':
+                id_field_type = field.get('type', None)
+                break
+
+        schema_statuses.append(id_field_type and id_field_type == 'string')
```
quesebifurcan, line 107: Looks like `id_field_type` needs to be reset for every schema, otherwise a successful validation of the first subschema will render the rest valid, even though they aren't:
```
>>> from aether.kernel.api.validators import _has_valid_id_field
>>> _has_valid_id_field([{'type': 'record', 'name': 'Test', 'fields': [{'name': 'id', 'type': 'string'}]}, {}, {}, {}])
True
```

aether-kernel/aether/kernel/api/validators.py
```diff
@@ -86,13 +86,23 @@ def validate_avro_schema(value):
...
+    # True if All of the statuses were True
+    return True in schema_statuses and False not in schema_statuses
```
quesebifurcan: We could use the `all` built-in here: `all(schema_statuses)`

quesebifurcan (dismissed):

aether-kernel/aether/kernel/api/validators.py
shawnsarwar: Done and done. Good catch.

quesebifurcan:
Is this intended? (on develop, after merge)
```
>>> _has_valid_id_field({})
True
```
This will most likely affect the UI, since `aetherBaseSchema` isn't used at all over there. 
@shawnsarwar

shawnsarwar:
I think we need to add aetherBaseSchema as a default element that's generated for each schema by the UI

PR #233, opened 2018-06-28
feat: allow gather play with entities

obdulia-losantos:
Needed for: https://jira.ehealthafrica.org/browse/AET-248

In KERNEL
- Include entities count in stats views
- Include attachments list within entity serialized data
- filter schema by linked project (through projectschema)
- filter project by linked schema (through projectschema)
- filter entity by linked schema (through projectschema)

In ODK
- Create and propagate identity mapping
- Fix docs in AVRO schema, labels in xForm can contain links to images (found in Birds form)


aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -89,6 +89,8 @@ def upsert_project_artefacts(project_id=None, project_name=None, schemas=[], map
             project_schema = ProjectSchema.objects.get(project=project, schema=schema)
         except ObjectDoesNotExist:
             project_schema = ProjectSchema.objects.create(
+                # this enforces the schema to be used only once by a project
+                id=schema.pk,  # sharing schema id ???
```
obdulia-losantos, line 93: @shawnsarwar @quesebifurcan 
I don't like so much this "workaround", do you have a better idea to solve this issue?

My task:
- Create the identity mapping using the AVRO schema derived from the xForm.

The problem:
- the mapping definition requires the list of used "projectschemas" in the rules, but in ODK only the future "schema" is defined.

```python
    # in aether/odk/api/kernel_utils.py (__xform_to_artefacts)
    mapping = {
        'id': <<xform.kernel_id>>,
        'name': "random name",
        'definition': {
            'entities': {
                <<xform id>>: <<xform.kernel_id>>  # this is the schema id, ODK ignores the existence of any projectschema
            },
            'mapping': <<rules using the derived schema>>,
        }
    }
```

shawnsarwar: I think this is a reasonable approach since an identity mapping and its related entity definition and project schema ONLY describe one version of an XForm.

shawnsarwar (approved):
Look great!

PR #235, opened 2018-06-29
WIP -- do not merge /Proof of concept nifi example

shawnsarwar:
Needs to be updated to more modern kernel but don't want to lose the NiFi example.


shawnsarwar:
Closing until relevant again.

PR #236, opened 2018-06-29
fix: (ui) check for recursive schemas

lordmallam:
Temporary mitigation when recursive schemas are used.

- Prevents `Input` component from crashing
- Displays `You have provided a recursive schema. Currently not supported` message 

Related JIRA ticket: [AET-290](https://jira.ehealthafrica.org/browse/AET-290)

Recursive schema = `{
  "name": "recursive",
  "type": "record",
  "fields": [
    {
      "name": "recursive",
      "type": "recursive"
    }
  ]
}`


shawnsarwar (dismissed):

obdulia-losantos (changes requested):
The way of indicating the "translatable" message is not correct

aether-ui/aether/ui/assets/apps/pipeline/sections/Input.jsx
```diff
@@ -103,16 +103,34 @@ class SchemaInput extends Component {
...
       // generate a new input sample
-      const input = type.random()
+      let input = {}
+      try {
+        input = type.random()
+      } catch (error) {
+        if (error.message && error.message.startsWith('Maximum call stack size exceeded')) {
+          this.setState({
+            error: error.message,
+            errorHead: 'You have provided a recursive schema. Currently not supported'
```
obdulia-losantos: this must use the `formatMessage(MESSAGES.recursiveError)` option.

Take a look at https://github.com/eHealthAfrica/aether/blob/1ac4f475e5625389c9527f08e6902f5f2db8b3e5/aether-ui/aether/ui/assets/apps/components/NavBar.jsx

lordmallam: This is setting the state value, formatting happens [here](https://github.com/eHealthAfrica/aether/pull/236/files/4ab7f7a290f317d7549329717836da0cd87dc9a9#diff-9e33e8d32267278e6482688e5c123949R155)

obdulia-losantos: But it's wrong! You are assigning the same id to two different messages, and react-int is going to complain! or it should.

There are many ways of doing this, you can create the MESSAGE dictionary and set the header using the message id and call it in the render process... many, many ways.

```javascript
this.setState({
  error: ...,
  errorHead: MESSAGE.mySpecificError
})

// later in the render step
```javascript 
const {formatMessage} = this props.intl

// ...
 <h4 className='hint-title'>
  { formatMessage(this.state.errorHead) }
</h4>
{this.state.error}

File: aether-ui/aether/ui/assets/apps/pipeline/sections/Input.jsx [hunk 2]
```diff
@@ -144,18 +144,26 @@ class SchemaInput extends Component {
   }
 
   render () {
+    const {formatMessage} = this.props.intl
+    const MESSAGES = defineMessages({
+      errorHead: {
+        defaultMessage: this.state.errorHead,
```
obdulia-losantos: this is the text not the state :wink:

aether-ui/aether/ui/assets/apps/pipeline/sections/Input.jsx
```diff
@@ -103,16 +103,45 @@ class SchemaInput extends Component {
 
   notifyChange (event) {
     event.preventDefault()
-
+    const {formatMessage} = this.props.intl
+    const MESSAGES = defineMessages({
```
obdulia-losantos: This can be declared outside the class. But I'm not going to complain if you keep it here ;-)

obdulia-losantos (dismissed):

obdulia-losantos (approved):
:joy:

quesebifurcan:
The difference between Firefox and Chrome (didn't test with other browsers yet) is that Firefox bails out earlier, throwing `InternalError` with the message "too much recursion". This is not caught by the change in this PR.

Chrome happily continues a bit longer and throws a `RangeError`. This PR addresses that.

Could we try to make the recursion detection work for Firefox as well?

obdulia-losantos:
Good catch!

obdulia-losantos:
What happens with IE, Edge, Safari and... :see_no_evil:

PR #237, opened 2018-06-29
Move Aether Consumer SDK to new repo

shawnsarwar:
The consumer SDK now lives here : https://github.com/eHealthAfrica/aether-consumer-sdk
That repo will be used to build and publish to pypi where it is available as 
`aet.consumer`


admbtlr (dismissed):

scripts/test_all.sh
```diff
@@ -98,7 +98,4 @@ kill_all
 # execute INTEGRATION TEST
 ./scripts/test_integration.sh
 
-# Testing Consumer Library
-./scripts/test_consumer_lib.sh
```
obdulia-losantos, line 102: Is this file `./scripts/test_consumer_lib.sh` still there?

obdulia-losantos: same question for the docker-compose files...

shawnsarwar: Good catch on the shell script. The compose file was in the local directory so it's already been moved.

obdulia-losantos (approved):

PR #241, opened 2018-07-02
chore: added PRODUCER_SETTINGS_FILE to env and updated references

shawnsarwar:
As requested by @tooxie, we've changed to test/ no test toggle to a settings file path, which is read from an environment variable. PRODUCER_SETTINGS_FILE. This should make it simpler for templated configuration files to be injected to change producer behavior.

### DIFF
--- a/aether-producer/manage.py
+++ b/aether-producer/manage.py
@@ -26,7 +26,7 @@ if __name__ == '__main__':
 
     try:
         settings_path = os.environ['PRODUCER_SETTINGS_FILE']
-    except Exception:
+    except KeyError:
         print('PRODUCER_SETTINGS_FILE not set in environment.')
         sys.exit(1)
     from producer import aether_producer
--- a/aether-producer/manage.py
+++ b/aether-producer/manage.py
@@ -24,17 +24,11 @@ from time import sleep
 
 if __name__ == '__main__':
 
-    if len(sys.argv) > 1:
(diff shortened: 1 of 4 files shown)


aether-producer/manage.py
```diff
@@ -24,17 +24,11 @@
...
-            print ("invalid argument: %s" % sys.argv[1])
-    else:
-        print("starting producer")
-        print("waiting for requisites")
-        from producer import aether_producer
-        aether_producer.main(test=False)
-        print("Started Producer")
+    try:
+        settings_path = os.environ['PRODUCER_SETTINGS_FILE']
+    except Exception:
```
tooxie: This should be `KeyError`, because if the error is actually something different the log message will be misleading.

tooxie: Another approach:
```py
settings_path = os.environ.get('PRODUCER_SETTINGS_FILE')
if not settings_path:
    print('...')
    sys.exit(1)
```

tooxie (changes requested):

shawnsarwar:
@tooxie We should be able to put as many DBs into a single RDS Postgres instance as the instance can handle: https://dba.stackexchange.com/questions/194940/multiple-databases-in-same-aws-rds-postgresql-instance

tooxie:
@shawnsarwar good! That should be quite straightforward then.

tooxie (approved):

PR #242, opened 2018-07-02
chore/release 0.7.2

shawnsarwar:
We need to release the Aether Producer to dockerhub. Here we add it to the current release build process.


scripts/release.sh
```diff
@@ -26,24 +26,37 @@ set -Eeuo pipefail
 
 # Build docker images
 IMAGE_REPO='ehealthafrica'
-APPS=( kernel odk couchdb-sync ui )
+CORE_APPS=( kernel odk couchdb-sync ui )
+CORE_COMPOSE='docker-compose.yml'
+CONNECT_APPS=( aether-producer )
+CONNECT_COMPOSE='docker-compose-connect.yml'
```
obdulia-losantos, line 32: I'm wondering if we should better create a `docker-compose-travis/release.yml` with the containers we want to release

shawnsarwar: We may also want to release "core" consumers using the same process so I'm not positive what would be best. I made a ticket in Jira: https://jira.ehealthafrica.org/browse/AET-303

shawnsarwar: For now can we live with this method? It's blocking and shouldn't necessarily effect how we do things in the future if we change our minds.

obdulia-losantos: :+1:

obdulia-losantos (approved):

PR #253, opened 2018-07-03
wip: release-0.7.3

shawnsarwar:
Fixed build process for aether producer and RCs.


scripts/release.sh
```diff
@@ -26,24 +26,37 @@ set -Eeuo pipefail
...
 
-if [ -z "$TRAVIS_TAG" ];
-then
-  VERSION=${VERSION}-rc
-fi
-
-for APP in "${APPS[@]}"
-do
-  AETHER_APP="aether-${APP}"
+release_app () {  # ( name of app -> $1, compose_path -> $2 )
```
tooxie, line 35: You could assign these 2 to a properly named variable instead of having a comment. Will make the code where it's used easier to read too.

docker-compose-connect.yml
```diff
@@ -31,7 +31,7 @@ services:
   # Aether Kafka Producer
   # ---------------------------------
 
-  aether-producer:
+  producer:
```
tooxie, line 34: Good! This mismatch was quite annoying :)

helm/kernel/templates/volume_claim.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: v1
 metadata:
   name: {{ .Chart.Name }}
   annotations:
-    volume.beta.kubernetes.io/storage-class: "aws-efs-gp"
+    volume.beta.kubernetes.io/storage-class: "aws-efs"
```
tooxie, line 7: Does this come from a different commit?

.travis.yml
```diff
@@ -17,12 +17,14 @@ script:
...
 - "./scripts/kill_all.sh"
+
 deploy:
 - provider: script
   script: "./scripts/release.sh"
   skip_cleanup: true
   on:
-    branch: ^release-\d+\.\d+(\.\d+)?(-\S*)?$
+    all_branches: true
+    condition: $TRAVIS_BRANCH =~ release\-[0-9]+\.[0-9]+[\.0-9]*
```
tooxie, line 27: Wouldn't it be more explicit to have the branch named literally with the `-rc` suffix? That way we know whether is final or not.

Besides, right now a release consists of a branch name, a tag, and a `VERSION` file. Which means that all 3 could differ (and it will eventually happen). Why don't we just use tags for RCs?

tooxie (changes requested):

.travis.yml
admbtlr: The idea is that we create a version branch when the feature(s) are ready, and do final work on it, releasing `-rc` images as we go. Once the features are ready to release, we apply a tag. Bug fix releases occur on the same branch, by applying tags.

So e.g. we create a `0.8` branch, and set `VERSION` to `0.8`. Every time something is merged to this branch, an `0.8-rc` image is built automatically. When we're ready to release, we apply a `0.8` tag, and `0.8` images are built and pushed.

We then set the `VERSION` file to 0.8.1, since that will be the next release on this branch (if there is one). Any bugfix merges will automatically build `0.8.1-rc` images. Once we're ready to release, we apply the `0.8.1` tag; `0.8.1` is built and pushed. We then set the `VERSION` file to `0.8.2`.

etc.

It's not perfect (I'd rather not need a `VERSION` file) but it works.

shawnsarwar:
This PR ended up containing too much extra fluf from the 0.7x branch. Closing and taking the new script to a new PR:
https://github.com/eHealthAfrica/aether/pull/258

.travis.yml
tooxie: I don't aim for *perfect*, but for *simple*. My question is since we already use tags to release, what's the difference between a major release, a bugfix release, and a release candidate? I don't understand why only 1 of these gets special treatment.

Meanwhile, the **actual** version of [all](https://github.com/eHealthAfrica/aether/blob/develop/aether-common-module/setup.py#L27) [our](https://github.com/eHealthAfrica/aether/blob/develop/aether-couchdb-sync-module/setup.py#L27) [packages](https://github.com/eHealthAfrica/aether/blob/develop/aether-kernel/setup.py#L27) [is](https://github.com/eHealthAfrica/aether/blob/develop/aether-odk-module/setup.py#L27) [*0.0.0*](https://github.com/eHealthAfrica/aether/blob/develop/aether-ui/setup.py#L27).

PR #254, opened 2018-07-04
fix: updated saladbar pipenv so we can run it stand-alone

shawnsarwar:
Moved pipfile and made sure the schemas folder isn't deleted by adding a .gitignore to it.


admbtlr (approved):

shawnsarwar:
This is breaking integration tests for some unknown reason so I'm going to close it for now.

PR #257, opened 2018-07-04
docs: include UI specs in README

obdulia-losantos:
:books: :books: :books: :books:

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@
     - [Aether Kernel](#aether-kernel)
     - [Aether ODK Module](#aether-odk-module)
     - [Aether CouchDB Sync Module](#aether-couchdb-sync-module)
+    - [Aether UI Module](#aether-ui-module)
 - [Usage](#usage)
   - [Users & Authentication](#users--authentication)
     - [UMS settings for local development](#ums-settings-for-local-development)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,6 +100,7 @@ of the most common ones with non default values. For more info take a look at th
 - `AETHER_KERNEL_URL_TEST`: `http://kernel-test:9000` Aether Kernel Testing Server url.
 - `AETHER_ODK_TOKEN`: `d5184a044bb5acff89a76ec4e67d0fcddd5cd3a1`
   to connect to it from other modules. It's used within the start up scripts.
+- `ADMIN_PASSWORD`: `adminadmin`. It's used within the start up scripts.
```
shawnsarwar, line 103: @quesebifurcan ; I thought we were changing this?

shawnsarwar (commented):
I'm not sure where we landed with default passwords. I think Fredrik was working on something?

README.md
obdulia-losantos: Yeah! He was taking out the `/conf/extras/initial.json` files, but the start up script will use these env vars. 

We are not using it locally but we need to inform the user or the container will never start :thinking: 

https://github.com/eHealthAfrica/aether/pull/252

quesebifurcan: What Obdulia said -- some modules still have hardcoded credentials which would need more refactoring than we have time for before Friday. #252 will make sure that no development passwords will be included in the images.

obdulia-losantos: Another mandatory variable is `DJANGO_SECRET_KEY`, without it the Django app does not start.

PR #260, opened 2018-07-04
WIP fix: remove custom "modified" field on Entity model

quesebifurcan:
The previous version of `aether.kernel.api.models.Entity`
appended the unique id of the instance
to the "modified" field. After recent updates to the
producer module, this overloaded behavior might no longer
be necessary. Inherit "created" and "modified" fields from
`TimeStampedModel` instead.

This PR is more of a suggestion -- I'd like to see us not conflating 
timestamps and ids in the same field but am not 100% sure
that other modules are _not_ depending on this custom format
already. I couldn't find anything, can you think of something?


aether-kernel/aether/kernel/api/models.py
```diff
@@ -429,7 +428,7 @@ class Meta:
         ordering = ('-modified',)
 
 
-class Entity(models.Model):
+class Entity(TimeStampedModel):
```
obdulia-losantos, line 431: Sounds good to me, but could you also change the docs regarding data model (the table definition below and the table schemas above)?

shawnsarwar:
The custom model turned out not be be a factor in the race condition. The reason for the TS+uuid as a modified value was to enforce a deterministic order for entities in the unlikely case that entities from many Aether instances were ever put into the same system and some had the same timestamp. It's probably not a very important constraint. Changing it would require a change to the producer, as it currently operates on a string comparison and not a SQL datetime comparison.

obdulia-losantos:
My question with this field it that it looks like `modified` field was taking/stealing the `revision` file role. :thinking:

shawnsarwar:
I believe this change is no longer required so I'm closing the PR.

PR #261, opened 2018-07-05
chore: release 0.8 addenda

admbtlr:


obdulia-losantos (approved):

aether-ui/aether/ui/settings.py
```diff
@@ -44,10 +44,8 @@
     },
 }
 
-INSTALLED_APPS += [
-    'webpack_loader',
-    'aether.ui',
-]
+# aether.ui must be first because of template overrides
+INSTALLED_APPS = ['aether.ui'] + INSTALLED_APPS + ['webpack_loader']
```
tooxie: A more pythonic way of doing this:
```py
INSTALLED_APPS = [
    'aether.ui',
    *INSTALLED_APPS,
    'webpack_loader',
]
```

This will prevent the creation of 2 lists that get immediately discarded.

aether-ui/aether/ui/templates/registration/logged_out.html
```diff
@@ -0,0 +1,80 @@
...
+    .top-nav-logo span {
+      display: inline-block;
+      font-weight: 300;
+      margin: 0 60px;
+      height: 24px;
+      color: white;
+    }
+    .btn {
+      background: rgba(255, 255, 255, 0.15);
+      box-shadow: 1px 1px 0 rgba(255, 255, 255, 0.1) inset, -1px -1px 0 rgba(45, 49, 70, 0.3) inset, 0 0 4px rgba(45, 49, 70, 0.1);
```
tooxie, line 40: Is there a reason why we can't use sass for these styles? Do they have to be inlined here?

tooxie (changes requested):

aether-ui/aether/ui/templates/registration/logged_out.html
admbtlr: Time pressure. This is an isolated case (overriding these two templates) and the tech debt that is introduced is *extremely* unlikely to accumulate over time, so I'm happy to push that work down the line.

tooxie (approved):

PR #263, opened 2018-07-05
chore: internal network name

obdulia-losantos:
Sorry but I had the same problem as Travis locally once I purged my local images :disappointed:


admbtlr (dismissed):

shawnsarwar (changes requested):
It looks like this is available on 3.5 +
https://forums.docker.com/t/set-default-network-name-for-compose/36779/3

shawnsarwar:
@obdulia-losantos I'm going to try to mess with this on the bootstrap as we'll have the same issue there. I'll let you know if I find a fix.

obdulia-losantos:
Let's try then :crossed_fingers:

obdulia-losantos:
But in this case `extends` is not supported :unamused:

obdulia-losantos:
This is becoming a nightmare... I'll try it later

PR #268, opened 2018-07-09
fix: propagate artefacts with create only option

obdulia-losantos:


shawnsarwar (approved):
Thanks for handling this on short notice!

PR #269, opened 2018-07-10
fix: mocker nesting

shawnsarwar:
Made changes to the mocker to handle schema with a union type as their base, and single type schemas. Also added namespace recognition to handle linked types that come in separately.


aether-utils/aether-mock-data/aether/mocker.py
```diff
@@ -314,8 +321,11 @@ def parse(self, schema):
         for field in fields:
             instructions.append(self._comprehend_field(field))
         self.instructions[name] = instructions
+        for i in self.instructions[name]:
+            print(name, i)
```
obdulia-losantos: do you need t print it?

aether-utils/aether-mock-data/aether/mocker.py
```diff
@@ -314,8 +321,11 @@ def parse(self, schema):
...
 
     def _comprehend_field(self, field):
+        print("comprehend", field)
```
obdulia-losantos: same comment

aether-utils/aether-mock-data/aether/mocker.py
shawnsarwar: Erps, frantic debugging. I'll add logging to it. Thanks

obdulia-losantos (approved):

PR #271, opened 2018-07-10
fix: remove hardcoded credentials

quesebifurcan:
This PR removes hardcoded credentials from code and 
config files in several modules.

Remove superuser fixtures from all django applications.
Replace them with the `setup_admin` script from the
`aether-common` module.

Remove superfluous `collectstatic` invocations from
`start_dev` blocks in entrypoint.sh.

Remove hardcoded passwords from salad-bar, producer, client
and mocker.

Create convenience scripts for quickly generating
credentials for local development with docker-compose and
minikube.

Remove no-op blocks from entrypoint files.

Rewrite docker-compose-base.yml to read sensitive data from
an `.env` file at the root of the repository.

Make couchdb-sync fail early when environment variables are
not set.

Add generated files `.env` and `helm/test-secrets.yaml` to
.gitignore.


aether-couchdb-sync-module/entrypoint.sh
```diff
@@ -83,12 +82,7 @@ setup_db() {
...
 
-setup_initial_data() {
-    # create initial superuser
-    ./manage.py loaddata /code/conf/extras/initial.json
-}
-
-setup_prod() {
+setup_admin() {
   # arguments: -u=admin -p=secretsecret -e=admin@aether.org -t=01234656789abcdefghij
   ./manage.py setup_admin -p=$ADMIN_PASSWORD
```
obdulia-losantos: Being :smiling_imp: could we standardize the indentation size? 2 or 4 spaces but the same in the whole file :wink:

scripts/generate-docker-compose-credentials.sh
```diff
@@ -0,0 +1,57 @@
...
+
+ODK_ADMIN_PASSWORD=$(gen_random_string)
+ODK_DJANGO_SECRET_KEY=$(gen_random_string)
+ODK_RDS_PASSWORD=$(gen_random_string)
+
+COUCHDB_SYNC_ADMIN_PASSWORD=$(gen_random_string)
+COUCHDB_SYNC_DJANGO_SECRET_KEY=$(gen_random_string)
+COUCHDB_SYNC_RDS_PASSWORD=$(gen_random_string)
+COUCHDB_SYNC_REDIS_PASSWORD=""
+COUCHDB_SYNC_GOOGLE_CLIENT_ID=
```
obdulia-losantos: :thinking: 

```COUCHDB_SYNC_GOOGLE_CLIENT_ID=${COUCHDB_SYNC_GOOGLE_CLIENT_ID}```

File: test-aether-integration-module/entrypoint.sh [hunk 3]
```diff
@@ -67,11 +67,6 @@ case "$1" in
         pip3 freeze --local | grep -v appdir | tee -a conf/pip/requirements.txt
     ;;
 
-    setuplocaldb )
-        #setup_db
-        #setup_initial_data
-    ;;
-
     setupproddb )
```
obdulia-losantos: Could this entry be renamed to `setup_db` like the others? :grimacing:

obdulia-losantos:
So far, looks good to me!

test-aether-integration-module/entrypoint.sh
quesebifurcan: It seems like it's not used at all, so let's remove it.

shawnsarwar (approved):
This is excellent. It's going to break all sorts of things in bootstrap, so we'll need to check that after the new images get built for v0.8.0-rc

scripts/generate-docker-compose-credentials.sh
```diff
@@ -0,0 +1,57 @@
...
+source ./scripts/random_string.sh
+
+cat << EOF
+COUCHDB_USER=admin
+COUCHDB_PASSWORD=$(gen_random_string)
+
+AETHER_KERNEL_TOKEN=$(gen_random_string)
+AETHER_ODK_TOKEN=$(gen_random_string)
+
+KERNEL_ADMIN_USERNAME=admin
```
obdulia-losantos, line 39: If we have a `KERNEL_ADMIN_USERNAME`... why don't we use it in the `setup_admin` instruction?

```bash
    # now
    ./manage.py setup_admin -p=$ADMIN_PASSWORD

    # but why not...
    ./manage.py setup_admin -p=$ADMIN_PASSWORD -u=$ADMIN_USERNAME -t=$ADMIN_TOKEN
```

Actually, why don't we have that for all containers? kernel, odk, ui...

helm/README.md
```diff
@@ -22,6 +22,16 @@ sudo ./scripts/kubernetes/install_helm.sh --version v2.8.1
 
 ## Running a local kubernetes cluster
 
+### Secrets
+
+For local development with Kubernetes and Minikube, we need to create some secrets.
+
+```bash
+POSTGRES_PASSWORD=<postgres-password> ./scripts/generate-kubernetes-credentials.sh > helm/test-secrets.yaml
```
tooxie, line 30: Since this file (`helm/test-secrets.yaml`) *has* to have that name, I think that we should hardcode it in the `generate-kubernetes-credentials.sh` script. Nothing would work if it were given a different name anyway.

scripts/generate-docker-compose-credentials.sh
quesebifurcan: Good point, explicit is better. Tried to change as little application code as possible in this PR, that's why. Will fix in a follow-up PR.

PR #276, opened 2018-07-11
fix(kernel): append suffix to artefact name if necessary

obdulia-losantos:


aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -182,3 +187,10 @@ def __random_name():
     # Names are unique so we try to avoid annoying errors with duplicated names.
     alphanum = string.ascii_letters + string.digits
     return ''.join([random.choice(alphanum) for x in range(50)])
+
+
+def __right_pad(value):
+    '''
+    Creates a random string of length 50 and appends it to the given value
```
admbtlr: Can we not just add an integer and increment it if necessary? So that if my project is called `my_project`, and I try to add another project with the same name, it will become `my_project_1`, and then the next one will become `my_project_2`, etc.

And if I'm crazy enough to try and create a project called `my_project_1`, then it will end up as `my_project_1_1`, since `my_project_1` is already taken.

admbtlr (dismissed):

aether-kernel/aether/kernel/api/project_artefacts.py
obdulia-losantos: :thinking:

obdulia-losantos:
@adamvert changed, it will first append `_1`,  `_2`, ...,  `_999999`. But in case this is not possible (all numbers were already in use, a really edge case) it will append the random string or fail.

admbtlr:
OK, that's an edge case I can live with :)

obdulia-losantos:
You can test them with these images:
- `ehealthafrica/aether-testing:aether-kernel-artefacts`
- `ehealthafrica/aether-testing:aether-odk-artefacts`

In `aether-boostrap` or `gather-deploy`

```yml
  # ---------------------------------
  # Aether Kernel
  # ---------------------------------

  kernel-base:
    image: ehealthafrica/aether-testing:aether-kernel-artefacts
    environment:
      ADMIN_USERNAME: ${KERNEL_ADMIN_USERNAME}
      ADMIN_PASSWORD: ${KERNEL_ADMIN_PASSWORD}
      ADMIN_TOKEN: ${KERNEL_ADMIN_TOKEN}

      CSRF_COOKIE_DOMAIN: .aether.local
      HOSTNAME: kernel.aether.local

      DJANGO_SECRET_KEY: ${KERNEL_DJANGO_SECRET_KEY}

      DB_NAME: aether
      PGHOST: db
      PGPASSWORD: ${KERNEL_DB_PASSWORD}
      PGPORT: 5432
      PGUSER: postgres

      WEB_SERVER_PORT: 8000
    volumes:
      - ./.persistent_data/kernel/media:/media
      - ./.persistent_data/kernel/static:/var/www/static
    ports:
      - "8000:8000"
    command: start

  # ---------------------------------
  # ODK module
  # ---------------------------------

  odk-base:
    image: ehealthafrica/aether-testing:aether-odk-artefacts
    environment:
      ADMIN_USERNAME: ${ODK_ADMIN_USERNAME}
      ADMIN_PASSWORD: ${ODK_ADMIN_PASSWORD}
      ADMIN_TOKEN: ${ODK_ADMIN_TOKEN}

      AETHER_KERNEL_TOKEN: ${KERNEL_ADMIN_TOKEN}
      AETHER_KERNEL_URL: http://kernel:8000
      AETHER_KERNEL_URL_TEST: http://kernel-test:9000

      CSRF_COOKIE_DOMAIN: .aether.local
      HOSTNAME: odk.aether.local

      DJANGO_SECRET_KEY: ${ODK_DJANGO_SECRET_KEY}

      DB_NAME: odk
      PGHOST: db
      PGPASSWORD: ${ODK_DB_PASSWORD}
      PGPORT: 5432
      PGUSER: postgres

      # behind ngnix
      WEB_SERVER_PORT: 8002
    volumes:
      - ./.persistent_data/odk/media:/media
      - ./.persistent_data/odk/static:/var/www/static
    ports:
      - "8002:8002"
    command: start
```

aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -182,3 +187,20 @@ def __random_name():
...
+    '''
+    Creates a numeric or a random string suffix for the given value
+    '''
+    numeric_suffix = 0
+    new_value = value
+
+    while model.objects.filter(name=new_value[0:50]).count() > 0:
+        if len(new_value) > 50:
+            # in case of name size overflow
+            return (f'{value} - {__random_name()}')[0:50]
```
tooxie: There's no need to check for the length, if you request the first 50 chars it won't fail if it's shorter:
```py
new_name = value

# just get the first one, we don't care if there are more than 1
while model.objects.filter(name=new_name).exists():
    new_name = f'{value}_{__random_name()}'[:50]

return new_name
```

obdulia-losantos: Django ORM complains with this `model.objects.filter(name=new_value)` if `new_name` is longer that `name` size. :unamused:

tooxie: strip it:
```py
new_name = value[:50]
# ... the rest stays the same
```

aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -182,3 +187,20 @@ def __random_name():
...
+
+def __right_pad(model, value):
+    '''
+    Creates a numeric or a random string suffix for the given value
+    '''
+    numeric_suffix = 0
+    new_value = value
+
+    while model.objects.filter(name=new_value[:50]).exists():
+        if len(new_value) > 50:
```
tooxie, line 200: My point was that there is no need for this check, if you get the first 50 characters, it will never be longer than that, and this check becomes irrelevant.

quesebifurcan:
Looks good :+1: though I'd love for us to re-remove the unique constraint on the name fields eventually. 

Question: can a project upsert (triggered by e.g. a gather request) result in something like:
- Project name: A
- Schema name: A_2
- Projectschema name: A_5

That would be a bit confusing and possibly break things for downstream consumers relying on names to match. They could of course then lookup the relationships by id, but why then require unique names in the first place?

aether-kernel/aether/kernel/api/project_artefacts.py
obdulia-losantos: As we discussed before, the case to avoid is this one:

`Project_zzzz` (48 chars and in use)
`Project_zzzz_1`  (50 chars and in use)
`Project_zzzz_2` (50 chars and in use)
...
`Project_zzzz_9` (50 chars and in use)

`Project_zzzz_10` (51 chars) if we strip it to `new_value = f'{value}_{numeric_suffix}'[:50]` it turns into `Project_zzzz_1` and we will never escape the `while` :scream:

tooxie (approved):

PR #277, opened 2018-07-11
chore: unify ports with start and start_dev commands

obdulia-losantos:
This will simplify the `aether-bootstrap` setup (all the custom `uwsgi.ini` files could be removed)

- Dynamic ports with `start` command using `WEB_SERVER_PORT` env var.
- ODK Module port is `8001` (`9001` in tests) to use ODK Collect run it behind nginx.
- CouchDB-Sync port is `8006` (`9006` in tests).
- Unify indentation to 4 spaces in scripts.
- Updated README


aether-couchdb-sync-module/entrypoint.sh
```diff
@@ -147,16 +147,16 @@ case "$1" in
         chown aether: /media
 
         # create static assets
-        ./manage.py collectstatic --noinput
+        ./manage.py collectstatic --noinput --noinput --clear --verbosity 0
```
tooxie: I would set the verbosity level based on the `$DEBUG` env var.

aether-couchdb-sync-module/entrypoint.sh
```diff
@@ -147,16 +147,16 @@ case "$1" in
...
-        cp /code/REVISION /var/www/REVISION
+        # expose version number (if exists)
+        cp ./VERSION /var/www/static/VERSION 2>/dev/null || :
+        # add git revision (if exists)
+        cp ./REVISION /var/www/static/REVISION 2>/dev/null || :
 
-        [ -z "$DEBUG" ] && DISABLE_LOGGING="true" || DISABLE_LOGGING="false"
-        /usr/local/bin/uwsgi --ini /code/conf/uwsgi.ini --disable-logging=$DISABLE_LOGGING
+        [ -z "$DEBUG" ] && LOGGING="--disable-logging" || LOGGING=""
+        /usr/local/bin/uwsgi --ini /code/conf/uwsgi.ini --http 0.0.0.0:$WEB_SERVER_PORT $LOGGING
```
tooxie: What about:
```sh
/usr/local/bin/uwsgi \
    --ini /code/conf/uwsgi.ini \
    --http 0.0.0.0:$WEB_SERVER_PORT \
    $DISABLE_LOGGING
```
Just a nitpick, up to you.

tooxie (dismissed):

aether-couchdb-sync-module/entrypoint.sh
obdulia-losantos: :+1:

aether-couchdb-sync-module/entrypoint.sh
```diff
@@ -147,16 +147,19 @@ case "$1" in
...
-        ./manage.py collectstatic --noinput
+        ./manage.py collectstatic --noinput --clear --verbosity 0
         chmod -R 755 /var/www/static
 
-        # expose version number
-        cp /code/VERSION /var/www/VERSION
-        # add git revision
-        cp /code/REVISION /var/www/REVISION
+        # expose version number (if exists)
+        cp ./VERSION /var/www/static/VERSION 2>/dev/null || :
```
quesebifurcan, line 154: Would it be possible to always include the version here (and exit with error if it's not present)? Just so it's not accidentally left out when building an image. Are there any situations in which we don't want to pass in a version?

quesebifurcan (changes requested):

aether-couchdb-sync-module/entrypoint.sh
obdulia-losantos: :thinking:

admbtlr: We want to ditch the version file and rely on `VERSION` and `REVISION`/`COMMIT` env vars being set on all the images at build time.

But this should probably happen in a separate PR that also changes `release.sh`, so maybe it's best to leave this as is for now?

quesebifurcan (approved):

aether-couchdb-sync-module/entrypoint.sh
quesebifurcan: Alright, that makes sense. See https://jira.ehealthafrica.org/browse/AET-317 for details.

PR #284, opened 2018-07-12
Add helm package for UI

tooxie:


helm/ui/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Kubernetes
```
MrPink: Can we make this more descriptive, this is used when you search Helm packages on a repository

helm/ui/templates/deployment.yaml
```diff
@@ -0,0 +1,121 @@
...
+        - name: ui
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.pullPolicy }}
+          args:
+          - {{ .Values.app.command }}
+          ports:
+            - name: http
+              containerPort: 8000
+              protocol: TCP
+          livenessProbe:
```
MrPink: I wonder if it makes sense to add an HTTP header to make clear what is hitting the `/health` endpoint, see https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/ under `httpHeaders`

helm/ui/templates/deployment.yaml
```diff
@@ -0,0 +1,121 @@
...
+            httpGet:
+              path: /health
+              port: http
+          readinessProbe:
```
MrPink, line 81: same here

helm/ui/templates/ingress.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: Ingress
+metadata:
+  name: {{ .Release.Name }}
+  labels:
+    app: {{ .Release.Name }}
+    chart: {{ template "ui.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+  annotations:
+    alb.ingress.kubernetes.io/scheme: "internet-facing"
```
MrPink: did we want AWS specific things in the ingress

helm/ui/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+service:
+  type: NodePort
+  externalPort: 80
+  internalPort: 80
+
+ingress:
+  enabled: true
+  path: /
+  annotations:
+    alb.ingress.kubernetes.io/security-groups: "alb-ingress-controller.dev.k8s.eha.im"
```
MrPink: probably makes sense to consolidate the annotations for `alb` in one file

helm/ui/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+    alb.ingress.kubernetes.io/certificate-arn: "arn:aws:acm:eu-west-1:387526361725:certificate/dcd71c49-1b5b-4b19-bd2d-6de98a3f6477"
+
+app:
+  port: 8000
+  command: start
+  secrets: ui-secrets
+  rdsDbName: aether_ui
+  rdsDbUser: aether_ui
+  rdsDbHost: dev-k8s-eha-im.cpqsfknb8sgc.eu-west-1.rds.amazonaws.com
```
MrPink, line 33: We can think about  `ExternalName service` for this perhaps - https://kubernetes.io/docs/concepts/services-networking/service/ but this doesn't have to be immediate.

aether-ui/entrypoint.sh
```diff
@@ -55,20 +55,16 @@ pip_freeze() {
 }
 
 setup_db() {
-  export PGPASSWORD=$RDS_PASSWORD
```
obdulia-losantos: Why are your removing these lines? They are necessary to connect to postgres :thinking:

tooxie: Because the deployment generates the correct environment variables, there is no need to rename them now. See:

https://github.com/eHealthAfrica/aether/blob/aa23af18eea4d7d549ba6ac2e3d76123ed9f0bd0/helm/ui/templates/deployment.yaml#L96-L108

obdulia-losantos: What happens locally? This is also used without helm :flushed:

tooxie: I checked the `aether-ui` source code and I couldn't find any mentions to these variables, but I see that `docker-compose-base.yml` sets them too. I'll issue a fix in a new PR. Is there any other place that you know these names should be updated?

obdulia-losantos: it's not used by django it's used by the psql commands `pg_isready`, `psql`, `createdb`...

obdulia-losantos: https://www.postgresql.org/docs/9.6/static/libpq-envars.html

tooxie: Yes, and they are being set with the correct name from the start. `RDS_PASSWORD` doesn't exist any more, it's now called `PGPASSWORD`, so there is no need to rename it.

tooxie: https://github.com/eHealthAfrica/aether/pull/285

MrPink (changes requested):

MrPink:
@tooxie can you also write how this should be tested in terms of installation

helm/ui/templates/deployment.yaml
tooxie: That information does not get logged, where does it show up? How would we use it?

helm/ui/templates/ingress.yaml
tooxie: I didn't know that was AWS-specific. Will move it back to the values file.

helm/ui/values.yaml
tooxie: I like the idea. But yeah, I also agree that we better think that through and introduce it for all charts at once.

tooxie:
@MrPink what do you mean by "tested in terms of installation"? Installation instructions? Automated tests? Manual tests? All of the above?

helm/ui/templates/deployment.yaml
MrPink: does it not get logged on debug?

MrPink:
as in the `helm install` + `override`

helm/ui/templates/deployment.yaml
tooxie: I'll double check, but I don't remember it being logged.

tooxie:
@MrPink there is no special instalation instructions for this package.

helm/ui/templates/nginx-config-map.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+      if ($http_x_forwarded_proto != 'https') {
+       return 301 https://$host$request_uri;
+      }
+
+      location /static/ {
+        alias /var/www/static/;
+      }
+
+      # Protected media
```
obdulia-losantos: There is no media content in `ui`, you can remove this if you want.

helm/ui/templates/init-config-map.yaml
```diff
@@ -0,0 +1,16 @@
...
+  init.sh: |
+    #!/bin/bash
+    set -u
+
+    psql -c "CREATE DATABASE $DB_NAME;"
+    psql -c "CREATE USER $DB_USERNAME WITH PASSWORD '$DB_PASSWORD';"
+    psql -c "GRANT ALL PRIVILEGES ON DATABASE $DB_NAME TO $DB_USERNAME;"
+
+    # Fix EFS mount permissions
+    chmod -R 775 /media
```
obdulia-losantos: Nice to have but not needed. :wink:

helm/ui/templates/deployment.yaml
```diff
@@ -0,0 +1,121 @@
...
+      - name: {{ .Chart.Name }}
+      - name: static-assets-volume
+        emptyDir: {}
+      - name: nginx-config
+        configMap:
+          name: {{ .Release.Name }}-nginx-config
+      - name: init-config
+        configMap:
+          name: {{ .Release.Name }}-init-config
+          defaultMode: 0744
```
obdulia-losantos, line 123: I missed this block

```
        # `WEB_SERVER_PORT` is kept around for compatibility with the old tests,
        # which are using different ports in different environments. Once all
        # tests have been updated to run inside the cluster, `WEB_SERVER_PORT`
        # can be removed.
        - name: WEB_SERVER_PORT
          value: {{ quote .Values.webserverPort }}
```

obdulia-losantos (changes requested):
`WEB_SERVER_PORT`is now required to start the container, even with the `start` command.

helm/ui/templates/deployment.yaml
tooxie: Added.

obdulia-losantos (approved):

helm/ui/templates/deployment.yaml
tooxie: Checked, they don't show up anywhere.

MrPink (approved):

PR #289, opened 2018-07-16
chore: simplify python containers setup

obdulia-losantos:


aether-common-module/setup.py
```diff
@@ -19,51 +19,28 @@
 import os
 from setuptools import find_packages, setup
 
-# allow setup.py to be run from any path
 os.chdir(os.path.normpath(os.path.join(os.path.abspath(__file__), os.pardir)))
 
 setup(
     name='aether.common',
     version='0.0.0',
```
tooxie, line 26: We should start using this version instead, and generate the `VERSION` file at build time. I think this is a change for a new PR, though.

tooxie (dismissed):

aether-common-module/setup.py
obdulia-losantos: yeah! it applies to all `setup.py` files :scream:

tooxie (approved):

PR #291, opened 2018-07-16
Set the common flags for our scripts

tooxie:


scripts/kubernetes/run_tests.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -ex
+set -Eeuox pipefail
```
obdulia-losantos, line 3: with `x`?

tooxie: Yes, it was there already, so I left it because there is no sensitive information being used here. Do you think it should be removed?

obdulia-losantos: Not really, but just for coherence with all script files :thinking:

tooxie: The thing is that it's also useful, for those scripts that don't log anything, to know what's going on.

scripts/kubernetes/start_cluster.sh
```diff
@@ -1,12 +1,11 @@
 #!/usr/bin/env bash
 
-set -u
+set -Eeuox pipefail
 
 # $VALUES_DIR should point to one of the directories in `helm/overrides`. Each
 # subdirectory (e.g. "local", "test") represents an environment and contains one
 # file for each module with settings for that environment.
 VALUES_DIR=$1
-eval $(minikube docker-env)
```
tooxie, line 9: @quesebifurcan I hope you don't mind, I had to remove this. Is there somewhere else we could add it, that doesn't get executed by travis?

obdulia-losantos (approved):

scripts/kubernetes/start_cluster.sh
quesebifurcan: None that I can think of now. But since it's related to local development with minikube, it could go in the docs perhaps?

tooxie: I will bring it back in this form:
```sh
if [ -z $TRAVIS ]; do
    eval $(minikube docker-env)
done
```

In a followup PR.

quesebifurcan (approved):

PR #292, opened 2018-07-16
Set minikube env vars only when not in Travis

tooxie:


scripts/kubernetes/start_cluster.sh
```diff
@@ -7,6 +7,10 @@ set -Eeuox pipefail
 # file for each module with settings for that environment.
 VALUES_DIR=$1
 
+if [ -z $TRAVIS ]; do
+    eval $(minikube docker-env)
+done
+
```
quesebifurcan: $TRAVIS will most likely not be set when developing locally, so it'll fail (due to the flags). You could move this statement up above the `set -Eeuox pipefail` line.
I'm getting a bash syntax error here as well -- did you mean `if [...] then`?

quesebifurcan (changes requested):

quesebifurcan (approved):

PR #280, opened 2018-07-12
Update kernel helm chart

tooxie:
Introduce the improvements that we already have for the UI chart.

PR #293, opened 2018-07-16
fix: create readonly user for non-core aether modules

quesebifurcan:
Create readonly user in kernel database. This role can be used
by non-core Aether modules such as `aether-producer`.

The user is granted read permissions for the following tables:
- kernel_entity
- kernel_mapping
- kernel_projectschema
- kernel_schema

NOTE:
The main script is in python, not pure SQL. I tried the latter, but
there were simply too many variables to interpolate; quoting gets 
tricky when the same variable has to serve as both identifier and 
literal.

Partly resolves https://jira.ehealthafrica.org/browse/AET-315


helm/kernel/templates/deployment.yaml
```diff
@@ -135,6 +135,16 @@ spec:
             secretKeyRef:
               name: {{ .Values.application.secret }}
               key: kernel-database-name
+        - name: KERNEL_READONLY_DB_USERNAME
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.application.secret }}
+              key: kernel-readonly-db-username
```
tooxie: I know this is consistent with the current state of the chart, but in #280 we are moving all non-password values from the secret to the values file.

tooxie (changes requested):
Yayyy for the principle of least privilege!! \o/ 😄 🎉

helm/kernel/templates/deployment.yaml
quesebifurcan: ah, that makes sense. Never seemed like a secret to me, this one.

shawnsarwar (dismissed):

tooxie (approved):

aether-kernel/sql/create_readonly_user.py
```diff
@@ -0,0 +1,82 @@
...
+
+def main():
+    dbname = os.environ['DB_NAME']
+    host = os.environ['PGHOST']
+    user = os.environ['PGUSER']
+
+    postgres_credentials = {
+        'dbname': dbname,
+        'host': host,
+        'port': 5432,
```
tooxie: Please get this from the environment as well.

tooxie (changes requested):

aether-kernel/sql/create_readonly_user.py
```diff
@@ -0,0 +1,83 @@
...
+BEGIN
+  IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = {rolename})
+  THEN
+      CREATE ROLE {role} WITH LOGIN ENCRYPTED PASSWORD {password}
+      INHERIT NOSUPERUSER NOCREATEDB NOCREATEROLE NOREPLICATION;
+  END IF;
+END
+$$ LANGUAGE plpgsql;
+
+GRANT CONNECT ON DATABASE {database} TO {role};
```
obdulia-losantos: Could you include this instruction before the GRANT ones? Just to remove previous privileges added manually.

```SQL
REVOKE ALL PRIVILEGES ON ALL TABLES IN SCHEMA public FROM {role};
```

obdulia-losantos: Better 

```sql
REVOKE  ALL PRIVILEGES ON DATABASE {database} FROM {role} CASCADE;
```

https://www.postgresql.org/docs/9.6/static/sql-revoke.html

aether-kernel/sql/create_readonly_user.py
```diff
@@ -44,6 +44,8 @@
 END
 $$ LANGUAGE plpgsql;
 
+REVOKE ALL PRIVILEGES ON DATABASE {database} FROM {role} CASCADE;
```
obdulia-losantos: :+1:

obdulia-losantos (approved):

tooxie (approved):

PR #295, opened 2018-07-17
fix: ensure env vars before start container

obdulia-losantos:
Changing from `os.environ.get('DJANGO_SECRET_KEY')` to `os.environ['DJANGO_SECRET_KEY']` makes that the app will raise a `KeyError` if the env var is missing.

`DJANGO_SECRET_KEY`, `DB_NAME` and `PGPASSWORD` are now required.
Renamed a couple of env vars :vulcan_salute: 
Added `check_vars.sh` script in container setup with friendly messages. We could also add `DJANGO_SECRET_KEY`, `DB_NAME` and `PGPASSWORD` there. :thinking: 
The eternal cleaning and refactoring :see_no_evil:


aether-common-module/aether/common/conf/settings.py
```diff
@@ -141,8 +141,8 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': os.environ.get('DB_NAME'),
-        'PASSWORD': os.environ.get('PGPASSWORD', ''),
+        'NAME': os.environ['DB_NAME'],
+        'PASSWORD': os.environ['PGPASSWORD'],
         'USER': os.environ.get('PGUSER', 'postgres'),
         'HOST': os.environ.get('PGHOST', 'db'),
         'PORT': os.environ.get('PGPORT', '5432'),
```
tooxie: How could we force this to fail in production if these variables are not set, while using sane defaults in dev? An `if DEBUG` condition, maybe?

aether-common-module/aether/common/conf/settings.py
```diff
@@ -183,34 +183,34 @@
...
         'raven.contrib.django.raven_compat',
     ]
     MIDDLEWARE = [
         'raven.contrib.django.raven_compat.middleware.SentryResponseErrorIdMiddleware',
     ] + MIDDLEWARE
+
+    SENTRY_CLIENT = os.environ.get(
+        'DJANGO_SENTRY_CLIENT',
+        'raven.contrib.django.raven_compat.DjangoClient'
+    )
```
tooxie: Why is this set in an environment variable? What possible values can it hold? Do we ever set it to something different?

aether-couchdb-sync-module/conf/check_vars.sh
```diff
@@ -0,0 +1,39 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+set -Eeuo pipefail
+
+check_variable() {
+    if [ -n "$1" ];
+    then
+        echo "$2 set in Aether CouchDB-Sync!"
+    else
```
tooxie: I would just `echo` and `exit` on error, and assume that no news is good news.

scripts/generate-kubernetes-credentials.sh
```diff
@@ -36,12 +36,14 @@ metadata:
...
 type: Opaque
 stringData:
+  kernel-admin-username: admin
   kernel-admin-password: $(gen_random_string)
   kernel-database-user: postgres
   kernel-database-password: $POSTGRES_PASSWORD
   kernel-database-name: aether
   kernel-django-secret-key: $(gen_random_string)
   kernel-token: $(gen_random_string)
+  odk-admin-username: admin
```
tooxie: Why is this a secret? I would remove all users from this file tbh.

tooxie (changes requested):

aether-common-module/aether/common/conf/settings.py
obdulia-losantos: never ever

scripts/generate-kubernetes-credentials.sh
obdulia-losantos: I don't find this a secret, I included here with the rest of env vars but I can move it to the `Values.yml` file if you prefer :+1:

aether-common-module/aether/common/conf/settings.py
obdulia-losantos: What do you mean? Do you prefer to remove the default values?

tooxie: If by some reason these variables don't get set in production we will see a connection error, which could be misleading. It would save us a lot of time to see the error that the variables are not set. I'm thinking:
```py
db_user = 'postgres' if DEBUG else os.environ['PGUSER']
# the same for host and port
```

scripts/generate-kubernetes-credentials.sh
tooxie: I would prefer that, yes, thanks. In the new kernel chart (#280) we moved this data to the `values.yaml` file and left only the password as a secret:

https://github.com/eHealthAfrica/aether/blob/32da5316ef6b2a22e38cacf0d0b19326a277f51d/helm/kernel/values.yaml#L25-L30

This is what a secret would look like:

https://github.com/eHealthAfrica/aether/blob/30c15f39201ecaedf94132005c1cfb03f9d0146b/helm/ui/secrets.tpl.yaml#L11-L19

aether-common-module/aether/common/conf/settings.py
tooxie: I don't like my approach now that I look at it 😅 We want to set `DEBUG=True` in the development cluster, and take those values from the environment at the same time.

scripts/generate-kubernetes-credentials.sh
obdulia-losantos: The point is, should I change them here or change them in your PR?

obdulia-losantos: your PR is huge and I wait for a lot of conflicts :scream:

aether-couchdb-sync-module/conf/check_vars.sh
```diff
@@ -0,0 +1,37 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+set -Eeuo pipefail
+
+check_variable() {
+    if [ -z "$1" ];
+    then
+        echo "Missing $2 in Aether CouchDB-Sync!"
+        exit -1
```
tooxie: Why a negative number?

aether-common-module/aether/common/conf/settings.py
```diff
@@ -141,10 +141,10 @@
...
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': os.environ.get('DB_NAME'),
-        'PASSWORD': os.environ.get('PGPASSWORD', ''),
-        'USER': os.environ.get('PGUSER', 'postgres'),
-        'HOST': os.environ.get('PGHOST', 'db'),
+        'NAME': os.environ['DB_NAME'],
+        'PASSWORD': os.environ['PGPASSWORD'],
+        'USER': os.environ['PGUSER'],
+        'HOST': os.environ['PGHOST'],
         'PORT': os.environ.get('PGPORT', '5432'),
```
tooxie: Why do we leave the port out?

docker-compose-common.yml
```diff
@@ -9,9 +9,15 @@ services:
   common:
     build: ./aether-common-module
     environment:
-      AETHER_KERNEL_URL_TEST: "http://kernel-test"
-      DJANGO_SECRET_KEY: "secret"
       TESTING: "true"
+      # required environment variables
+      DJANGO_SECRET_KEY: secret
+      PGPASSWORD: secretsecret
```
tooxie, line 15: Why are these not picked up from the `.env` file?

tooxie (changes requested):

docker-compose-common.yml
obdulia-losantos: Because this is the common module with just dummy values for the tests, no need of using the `.env` file here.

aether-couchdb-sync-module/conf/check_vars.sh
obdulia-losantos: :woman_shrugging:

aether-common-module/aether/common/conf/settings.py
obdulia-losantos: Because most of the times the databases use the default port... do we need to do this stricter?

tooxie: Yes, I would keep all together. That said, I'm ok with adding this in a new PR.

tooxie (approved):
Could we make a release with all these changes? We need them in the kubernetes setup.

PR #296, opened 2018-07-17
Add random string generator for mac

tooxie:


scripts/random_string.sh
```diff
@@ -1,3 +1,10 @@
 gen_random_string () {
-    cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1
+    OPENSSL=`which openssl`
+    if [ -z "$OPENSSL" ]; then
+        # In linux we use /dev/urandom
+        cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1
+    else
+        # In mac we use `openssl`
+        openssl rand -hex 16 | tr -d "\n"
```
quesebifurcan: What happens if openssl is not installed on a mac -- that would result in the same "Illegal byte sequence" error, wouldn't it? Can we perhaps add a note about openssl in the README?

obdulia-losantos:
I would let mac people review this :grimacing:

quesebifurcan (changes requested):

obdulia-losantos:
In ubuntu with openssl

```bash
aether (gen-random-string-onmac) $ which openssl
/usr/bin/openssl
```

obdulia-losantos (approved):

quesebifurcan (approved):

PR #303, opened 2018-07-19
feat: Add DB port to values file

tooxie:


helm/ui/templates/deployment.yaml
```diff
@@ -72,7 +72,7 @@ spec:
           - {{ .Values.app.command }}
           ports:
             - name: http
-              containerPort: {{ quote .Values.app.port }}
+              containerPort: {{ .Values.app.port }}
```
obdulia-losantos, line 75: Why without quotes?

tooxie: Because, if I understood correctly, the quotes are only needed for environment variables. Kubernetes can handle ports as integers.

obdulia-losantos (approved):

PR #307, opened 2018-07-23
chore: add link to CLA

quesebifurcan:


tooxie (approved):
I think we should be a little more detailed on how to contribute :) but it's a good start.

obdulia-losantos (approved):

PR #313, opened 2018-07-25
refactor the config handling and add env variable lookup

tremolo:
change the way config variables are looked up:
 - add automatic env variable lookup by uppercasing the key name, env variables have precedence over the config file 
- flatten the config file a bit to make this work properly
- add special case for the kafka config that is passed on to the confluent_kafka module


shawnsarwar (dismissed):
Changes look good, just needs to pass integration tests. You can run then locally with the integration test script from /scripts.

shawnsarwar (approved):

PR #315, opened 2018-07-27
fix(kernel): indices and filters

obdulia-losantos:
~Any hint on why `test_entity_filter__by_project` is failing?~ 

- Included check by uuid or name in `project` and `schema` filters.
  Now it's possible to filter by `/entitites/project=My_Project`  or `/entitites/project=1234-564-263`.

- Included redundant field `project` in models to speed up lookups.

- Included indices using this new field.
  Gather fetches the data using the `project` filter: `SELECT * FROM kernel_entity WHERE project={project_id}`, in this case the expected behavior is also to return them ordered by the `modified` field in descendant order; the easiest way to achieve this is creating an index with this condition: `project`, `-modified`.
Obviously, later we must re-examine this index and check if it's really useful or conversely it decreases the performance being a stopper during insertion and updating tasks. But this cannot be done without more real data and reading the db logs or more recently the prometheus metrics.

https://www.postgresql.org/docs/9.6/static/indexes-examine.html


quesebifurcan:
@obdulia-losantos could it be that you're running into a pagination issue? That is, the entities are correctly filtered but only the first 30 (?) are returned in the response. The filter logic itself looks ok to me.

obdulia-losantos:
:woman_facepalming:

aether-kernel/aether/kernel/api/models.py
```diff
@@ -473,7 +280,19 @@ class Entity(ExportModelOperationsMixin('kernel_entity'), models.Model):
     projectschema = models.ForeignKey(to=ProjectSchema, on_delete=models.SET_NULL, null=True)
     submission = models.ForeignKey(to=Submission, on_delete=models.SET_NULL, blank=True, null=True)
 
+    # redundant but speed up queries
+    project = models.ForeignKey(to=Project, on_delete=models.SET_NULL, blank=True, null=True)
```
quesebifurcan, line 284: Should it be possible to create an entity which does not belong to a Project? When would that be useful? Asking since it's usually helpful to have as few nullable fields as possible.

aether-kernel/aether/kernel/api/models.py
```diff
@@ -51,62 +53,31 @@
...
-            REFERENCES kernel_project(id)
-            DEFERRABLE INITIALLY DEFERRED
-
-        TABLE "kernel_projectschema"
-            CONSTRAINT "kernel_projectschema_project_id_###_fk_kernel_project_id"
-            FOREIGN KEY (project_id)
-            REFERENCES kernel_project(id)
-            DEFERRABLE INITIALLY DEFERRED
-
-    '''
```
quesebifurcan, line 109: +1 on removing these docstrings; the models contain this information anyway.

aether-kernel/aether/kernel/api/models.py
```diff
@@ -122,43 +93,13 @@ def __str__(self):
     class Meta:
         app_label = 'kernel'
         default_related_name = 'projects'
-        ordering = ('-modified',)
+        ordering = ['-modified']
+        indexes = [
+            models.Index(fields=['-modified']),
+        ]
```
quesebifurcan, line 99: Could you add a comment about approximately how much this improved performance for your particular usecase? Either in the PR description or the commit itself. Would be helpful to have for future readers.

quesebifurcan (changes requested):

aether-kernel/aether/kernel/api/models.py
obdulia-losantos: It might not but because `projectschema` and `submission` were already optional... :woman_shrugging:

obdulia-losantos: This is something we need to discuss, the `on_delete=models.SET_NULL` option indicates that if the foreign instance it's deleted the entity will remain in the database.

quesebifurcan: Right, it makes sense that `project` is optional if `projectschema` and `submission` are. As I've understood the system, only `submission` need to be nullable (an entity might be submitted directly, bypassing entity extraction) but `projectschema` is always required (an entity is expected to conform to its `self.projectschema.schema`).

obdulia-losantos: In that case, might it be better to link the entity directly to the schema? or maybe do not allow to delete a project schema, or a project, or a schema if they have at least one linked entity? `on_delete=models.PROTECT` :thinking: 

@shawnsarwar @quesebifurcan

quesebifurcan: Maybe keep it this way for now and address this issue in a separate PR?

quesebifurcan (approved):

PR #317, opened 2018-07-27
feat: use S3/GCS for file storage

lordmallam:
Use S3 as default file storage

Env:

- `DJANGO_REMOTE_STORAGE` (optional): Used to specify a custom [Default file storage system](https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-DEFAULT_FILE_STORAGE). If not provided, defaults to the [Django default storage](https://docs.djangoproject.com/en/2.0/ref/files/storage/#django.core.files.storage.FileSystemStorage) Available options: S3, . More information [here](https://django-storages.readthedocs.io/en/latest/index.html)
- `REMOTE_STATIC_FILES` (optional): Set to "true" to collect static files to the selected `DJANGO_REMOTE_STORAGE`
- `BUCKET_NAME` (optional): Name of the bucket to use for `DJANGO_REMOTE_STORAGE`. Must be unique on the storage provider. If not provided, it defaults to the `HOSTNAME` value. Bucket is automatically created if does not exist.
- `AWS_ACCESS_KEY_ID`: AWS Access Key to your S3 account. Used when `DJANGO_REMOTE_STORAGE=S3`.
- `AWS_SECRET_ACCESS_KEY`: AWS Secret Access Key to your S3 account. Used when `DJANGO_REMOTE_STORAGE=S3`.


aether-common-module/conf/pip/requirements.txt
```diff
@@ -6,10 +6,12 @@
 
 django<2
 django-cors-headers
+django-storages
 djangorestframework
 djangorestframework-csv
 psycopg2-binary
 requests
+boto3
```
obdulia-losantos, line 16: You need to include those two new dependencies in the `setup.py` file, in `install_requires` option, and remove from the `primary-requirements.txtx` files

aether-kernel/entrypoint.sh
```diff
@@ -154,6 +154,10 @@ case "$1" in
 
         # media assets
         chown aether: /media
+        if [  ! -z $DJANGO_REMOTE_STORAGE ] && [ ! -z $REMOTE_STATIC_FILES ] ; then
+            echo "Collecting static files to" $DJANGO_REMOTE_STORAGE
+            ./manage.py collectstatic --noinput
+        fi
```
obdulia-losantos: Do we need this for the static files? :flushed:

README.md
```diff
@@ -134,6 +134,13 @@ of the most common ones with non default values. For more info take a look at th
   Token used to verify the device identity with Google.
   See more in https://developers.google.com/identity/protocols/OAuth2
 
+#### File Storage System (Used on Kernel, ODK and UI Modules)
```
obdulia-losantos: I thought that this was only needed for Kernel and ODK to save the media content (attachments in kernel and xForm media files in ODK) :thinking:

aether-kernel/entrypoint.sh
lordmallam: 🤷‍♂️ , Its optional

README.md
obdulia-losantos: https://docs.djangoproject.com/en/2.0/topics/files/#the-built-in-filesystem-storage-class

lordmallam: UI was mainly included for the static files option

obdulia-losantos: With the same argument, what happens with `couchdb-sync`?

lordmallam: As long as `common` is a dependency on any module. s3 storage feature will be available with the same arguments

obdulia-losantos: But it's not included in the PR :grimacing:

lordmallam: Yea. Changes due to  your 1st comment will address that 😁

README.md
```diff
@@ -134,6 +134,13 @@ of the most common ones with non default values. For more info take a look at th
...
+- `DJANGO_REMOTE_STORAGE` (optional): Used to specify a custom [Default file storage system](https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-DEFAULT_FILE_STORAGE). If not provided, defaults to the [Django default storage](https://docs.djangoproject.com/en/2.0/ref/files/storage/#django.core.files.storage.FileSystemStorage) Available options: S3, . More information [here](https://django-storages.readthedocs.io/en/latest/index.html)
+- `REMOTE_STATIC_FILES` (optional): Set to "true" to collect static files to the selected `DJANGO_REMOTE_STORAGE`
+- `BUCKET_NAME` (optional): Name of the bucket to use for `DJANGO_REMOTE_STORAGE`. Must be unique on the storage provider. If not provided, it defaults to the `HOSTNAME` value. Bucket is automatically created if does not exist.
+- `AWS_ACCESS_KEY_ID`: AWS Access Key to your S3 account. Used when `DJANGO_REMOTE_STORAGE=S3`.
```
MrPink: we should never need to use AWS keys as we use AWS IAM roles for this

MrPink: see https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html

aether-common-module/aether/common/conf/settings.py
```diff
@@ -232,6 +232,26 @@
...
+# ------------------------------------------------------------------------------
+
+CUSTOM_STORAGE = os.environ.get('DJANGO_REMOTE_STORAGE', '')
+
+if CUSTOM_STORAGE:   # pragma: no cover
+    INSTALLED_APPS += ['storages', ]
+    if CUSTOM_STORAGE.lower() == 's3':
+        DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
+        AWS_AUTO_CREATE_BUCKET = True
+        AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID', '')
```
MrPink: you should be able to simply remove `AWS_SECRET_ACCESS_KEY` and `AWS_SECRET_ACCESS_KEY`

MrPink: do you need to set `AWS_DEFAULT_REGION`?

MrPink (changes requested):

aether-common-module/aether/common/conf/settings.py
```diff
@@ -232,6 +232,26 @@
...
+        AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY', '')
+        AWS_STORAGE_BUCKET_NAME = os.environ.get('BUCKET_NAME',
+                                                 os.environ.get('HOSTNAME', 'aether-project'))
```
tooxie: I would suggest that we avoid defaults, because if we fail to set them we won't notice and the system will behave unexpectedly. So simply get them as a dict, so it will fail fast:
```py
AWS_STORAGE_BUCKET_NAME = os.environ['BUCKET_NAME']
```

tooxie (changes requested):

aether-common-module/aether/common/conf/tests/test_settings.py
```diff
@@ -32,3 +33,7 @@ def test_default_variables(self):
         self.assertEqual(settings.SECURE_PROXY_SSL_HEADER, None)
 
         self.assertEqual(settings.ROOT_URLCONF, 'aether.common.urls')
+
+    def test_custom_storage_variables(self):
+        with self.settings(DEFAULT_FILE_STORAGE='storages.backends.s3boto3.S3Boto3Storage'):
+            self.assertEqual(str(default_storage.__class__), "<class 'storages.backends.s3boto3.S3Boto3Storage'>")
```
tooxie: This is not really testing anything but django internal storage machinery. There doesn't seem there's anything to be unit-tested in this PR.

tooxie (changes requested):

aether-common-module/aether/common/conf/settings.py
lordmallam: `AWS_AUTO_CREATE_BUCKET` is set to `True` so the default bucket name will be created automatically.

README.md
lordmallam: @MrPink Will that be the case if aether is hosted outside AWS?

aether-common-module/aether/common/conf/settings.py
lordmallam: `AWS_DEFAULT_REGION` is not required

README.md
MrPink: image

We have options

MrPink: it should use the supported `filesystem-storage-class` for that particular Cloud provider

lordmallam: I understand what you mean. What is implemented allows a developer to choose to host on one platform and store data on another. Even if the host isn't listed above (on-Premise)

aether-common-module/aether/common/conf/tests/test_settings.py
tooxie: So, are we removing these tests?

tooxie (approved):

MrPink (approved):

aether-kernel/conf/pip/requirements.txt
```diff
@@ -31,17 +34,26 @@ django-prometheus==1.0.15
...
 drf-dynamic-fields==0.3.0
 drf-openapi==1.3.0
 flake8==3.5.0
 flake8-quotes==1.0.0
+google-api-core==1.3.0
+google-auth==1.5.0
+google-cloud-core==0.28.1
+google-cloud-storage==1.10.0
+google-resumable-media==0.3.1
+googleapis-common-protos==1.5.3
```
obdulia-losantos, line 52: @lordmallam did you introduce these dependencies manually? They disappear when you try to freeze pip dependencies :disappointed:

PR #320, opened 2018-07-30
fix: check_vars.sh without u flag

obdulia-losantos:


aether-couchdb-sync-module/conf/check_vars.sh
```diff
@@ -38,3 +38,6 @@ check_variable $AETHER_KERNEL_TOKEN "Aether KERNEL token (AETHER_KERNEL_TOKEN)"
 
 # Google ID
 check_variable $GOOGLE_CLIENT_ID    "Google Client ID (GOOGLE_CLIENT_ID)"
+
+# set default value for DEBUG if missing
+DEBUG="$DEBUG"
```
quesebifurcan: What does this line do? As far as I can tell, it wouldn't have any effect. Did you intend to include a default value and export it somehow?

aether-couchdb-sync-module/conf/check_vars.sh
```diff
@@ -38,3 +38,6 @@ check_variable $AETHER_KERNEL_TOKEN "Aether KERNEL token (AETHER_KERNEL_TOKEN)"
 
 # Google ID
 check_variable $GOOGLE_CLIENT_ID    "Google Client ID (GOOGLE_CLIENT_ID)"
```
quesebifurcan: Can we quote these (i.e. `"GOOGLE_CLIENT_ID"`) as well? In my bash version, this script exits with `0` even though none of those env vars are set. Adding quotes fixes that.

quesebifurcan (changes requested):

aether-couchdb-sync-module/conf/check_vars.sh
obdulia-losantos: Oops!

aether-couchdb-sync-module/conf/check_vars.sh
```diff
@@ -29,12 +29,15 @@ check_variable() {
...
-check_variable $AETHER_KERNEL_TOKEN "Aether KERNEL token (AETHER_KERNEL_TOKEN)"
+check_variable "$AETHER_KERNEL_URL"   "Aether KERNEL url (AETHER_KERNEL_URL)"
+check_variable "$AETHER_KERNEL_TOKEN" "Aether KERNEL token (AETHER_KERNEL_TOKEN)"
 
 # Google ID
-check_variable $GOOGLE_CLIENT_ID    "Google Client ID (GOOGLE_CLIENT_ID)"
+check_variable "$GOOGLE_CLIENT_ID"    "Google Client ID (GOOGLE_CLIENT_ID)"
+
+# set default value for DEBUG if missing
+export DEBUG="$DEBUG"
```
quesebifurcan: Looking at how this script is run (inside of e.g. `kernel/entrypoint.sh`), `DEBUG` wouldn't get exported into the `kernel/entrypoint.sh` environment. Parent->child process/shell would work, but not the other way around. Also, I'm not quite sure what this line would be doing anyway -- it seems to just reassign the value of `$DEBUG` to itself, or?

quesebifurcan (changes requested):

aether-couchdb-sync-module/conf/check_vars.sh
tooxie: My suggestion was that we did this before setting the `-u` flag so that it wouldn't fail:
```bash
DEBUG="$DEBUG"
set -u
# the rest of the script
```
But I agree that here it won't work.

scripts/release.sh
```diff
@@ -20,6 +20,13 @@
 #
 set -Eeuo pipefail
 
+# Try to create the common aether network if it doesn't exist.
+{
+    docker network create aether_internal
```
tooxie: We could simply call:
```sh
docker network create aether_internal 2>/dev/null || true
```

tooxie (dismissed):

aether-couchdb-sync-module/conf/check_vars.sh
obdulia-losantos: What about removing the `-u` flag in the `entrypoint.sh` file? We can check all of the required/used env vars in the `check_vars.sh` script :grimacing: 

```bash 
# App requirements
check_variable "$DB_NAME"           "Database name (DB_NAME)"
check_variable "$WEB_SERVER_PORT"   "Web server port (WEB_SERVER_PORT)"
```

quesebifurcan: We could also keep the `-u` flag in the entrypoints and remove the `check_vars` files? :smile: 

To be clear, I'm perfectly ok with `check_vars.sh`, but the argument would go like this: `-u` is a bit more future-proof -- no need to remember adding new vars to the `check_vars.sh` scripts in all modules.  Result would be the same -- `entrypoint.sh` would exit with `1` whenever it encounters an unset variable. Downside: error message not as verbose; they would look like
```
<filename> <linenumber>: <env-var-name> parameter not set
```
I guess there are a couple of environment variables which are not in the entrypoints but still are required. Those could be checked whenever they are retrieved for the first time (most likely in `settings.py`).

quesebifurcan (dismissed):

obdulia-losantos:
Django required env vars:

### Common:
- `DJANGO_SECRET_KEY`
- `DB_NAME`
- `PGPASSWORD`
- `PGUSER`
- `PGHOST`
- `PGPORT`

### Kernel specific:
- `KERNEL_READONLY_DB_USERNAME`
- `KERNEL_READONLY_DB_PASSWORD`

### CouchDB-Sync specific:
- `GOOGLE_CLIENT_ID`
- `COUCHDB_USER`
- `COUCHDB_PASSWORD`
- `REDIS_HOST`
- `REDIS_PORT`
- `REDIS_PASSWORD`

Should we include them in the `check_vars.sh` file or let Django complain?

obdulia-losantos:
@tooxie @quesebifurcan ^^

quesebifurcan:
@obdulia-losantos My preference would be to let django complain, _if_ it does so during its startup stage (when loading `settings.py`). If we for some reason have an environment variable which gets read dynamically while django is running, then we can put it in `check_vars.sh` until it can be refactored away.

quesebifurcan (dismissed):

shawnsarwar (dismissed):

quesebifurcan (approved):

PR #328, opened 2018-08-08
WIP fix: simplify storage setup

quesebifurcan:
Status: Kubernetes/minikube tests are currently failing. We are waiting for an update to the Aether helm charts which will include a new `DJANGO_STORAGE_BACKEND` environment variable.


README.md
```diff
@@ -137,19 +137,18 @@ of the most common ones with non default values. For more info take a look at th
...
-  - `BUCKET_NAME`: Name of the bucket to use on S3. Must be unique on S3. Bucket will be automatically created if the name is available.
-  - `AWS_ACCESS_KEY_ID`: AWS Access Key to your S3 account. Used when `DJANGO_REMOTE_STORAGE=S3`.
-  - `AWS_SECRET_ACCESS_KEY`: AWS Secret Access Key to your S3 account. Used when `DJANGO_REMOTE_STORAGE=S3`.
- 
-  ##### Google Cloud Storage (GCS)
-  - `BUCKET_NAME`: Name of the bucket to use on GCS. Create bucket using [Google Cloud Console](https://console.cloud.google.com/) and set appropriate permissions
-  - `GS_ACCESS_KEY_ID`: Google Cloud Access Key. Used when `DJANGO_REMOTE_STORAGE=GCS`. [How to create Access Keys on Google Cloud Storage](https://cloud.google.com/storage/docs/migrating#keys)
-  - `GS_SECRET_ACCESS_KEY`: Google Cloud Secret Access Key. Used when `DJANGO_REMOTE_STORAGE=GCS`. [How to create Access Keys on Google Cloud Storage](https://cloud.google.com/storage/docs/migrating#keys)
-  
+  - `BUCKET_NAME`: Name of the bucket to use on s3. Must be unique on s3. Bucket will be automatically created if the name is available.
```
tooxie: I would not create it automatically, if there's a typo it will lead to a subtle misbehaviour and then we'll have to reconcile the 2 diverging filesystems.

README.md
```diff
@@ -137,19 +137,18 @@ of the most common ones with non default values. For more info take a look at th
 
 #### File Storage System
 (Used on Kernel, ODK and UI Modules)
-- `DJANGO_REMOTE_STORAGE` (optional): Used to specify a custom [Default file storage system](https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-DEFAULT_FILE_STORAGE). If not provided, defaults to the [Django default storage](https://docs.djangoproject.com/en/2.0/ref/files/storage/#django.core.files.storage.FileSystemStorage) Available options: S3, GCS. More information [here](https://django-storages.readthedocs.io/en/latest/index.html)
-- `REMOTE_STATIC_FILES` (optional): Set to "true" to collect static files to the selected `DJANGO_REMOTE_STORAGE`
+- `DJANGO_STORAGE_BACKEND`: Used to specify a [Default file storage system](https://docs.djangoproject.com/en/1.11/ref/settings/#default-file-storage). Available options: filesystem, s3, gcs. More information [here](https://django-storages.readthedocs.io/en/latest/index.html).
```
tooxie: Is there a default? Or is this mandatory?

aether-common-module/aether/common/conf/settings.py
```diff
@@ -245,27 +245,22 @@
...
-        if STATIC_FILES:
-            STATICFILES_STORAGE = 'storages.backends.gs.GSBotoStorage'
-
-    logger.info('{} file system used!'.format(CUSTOM_STORAGE))
+DJANGO_STORAGE_BACKEND = os.environ['DJANGO_STORAGE_BACKEND']
+
+if DJANGO_STORAGE_BACKEND == 'filesystem':
+    DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
+elif DJANGO_STORAGE_BACKEND == 's3':
+    INSTALLED_APPS += ['storages']
```
tooxie: Minor: I would add it to the `INSTALLED_APPS` list and just keep it there. The overhead is minimum, and it will only be for development.

aether-common-module/aether/common/conf/settings.py
```diff
@@ -245,27 +245,22 @@
...
+    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
+    AWS_STORAGE_BUCKET_NAME = os.environ['BUCKET_NAME']
+elif DJANGO_STORAGE_BACKEND == 'gcs':
+    INSTALLED_APPS += ['storages']
+    DEFAULT_FILE_STORAGE = 'storages.backends.gs.GSBotoStorage'
+    GS_BUCKET_NAME = os.environ['BUCKET_NAME']
 else:
-    logger.info('Default file system used!')
+    msg = 'DJANGO_STORAGE_BACKEND "{}" is currently not supported'
```
tooxie: I guess that this will always be set to a non-empty string?

tooxie (changes requested):

aether-common-module/aether/common/conf/settings.py
quesebifurcan: It might be something else, e.g. a typo like `flesystem` or `s#`.

tooxie: Yes, my question was about the fact that we treat it always as a non-empty string, and if is it safe to assume that we will never get an empty string there.

README.md
quesebifurcan: It's mandatory now -- hence the additions of `filesystem` to all compose files. If you can think of a better way to:
- make sure that we always have either `s3` or `gcs` in production and
- include a default for local development without cluttering the compose files

...I'm all ears.
In other words, the trade-off is: more complex compose configuration, more certainty that prod environments will be configured correctly.

README.md
quesebifurcan: Good point, `AWS_AUTO_CREATE_BUCKET = True` was actually removed in this PR (at Will's request), so will update the README to reflect this.

aether-common-module/aether/common/conf/settings.py
quesebifurcan: ah, I see. Yes, if we get an empty string here, the error message might be confusing. Will fix.

README.md
tooxie: Maybe it's worth mentioning it in the README that it's mandatory? I don't know, up to you.

I think that's it's good that there's no default here 👍

quesebifurcan: Added to the README.

aether-common-module/aether/common/conf/settings.py
quesebifurcan: Agreed, PR updated.

tooxie (approved):

PR #332, opened 2018-08-22
fix: change kafka url / improve status

shawnsarwar:
This PR aligns the kafka url setting from using kafka's bootstrap nomenclature to KAFKA_URL which is used by consumers and preferred by devops. We also added better live checking of Kafka by the producer. By pinging Kafka's port and expecting Kafka to reset the connection when it's available. This should help troubleshoot configuration errors in which Kafka is not at the expected url/port. Lastly, we've cleaned up the wording of the `/status` endpoint to reduce confusion as to the current state. We previously had a binary value for Kafka which would lead you to assume it was unavailable when the status was False. We've replaced that with two status:
```
  "kafka_container_accessible": true, 
  "kafka_submission_status": "SUBMISSION_ONGOING",
```
`kafka_container_accessible` is the result of the aforementioned liveness check. `kafka_submission_status`  will be `SUBMISSION_ONGOING` if the last attempt at sending data to Kafka successed, or `DATA_NOT_BEING_SUBMITTED` if the last submission failed, or we have yet to attempt a submission (because Kernel is unavailable, etc).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ python:
 matrix:
   fast_finish: true
   include:
-    - name: "Kubernetes tests (Kernel and ODK)"
-      env: 'TEST_MODE=kubernetes'
+    # - name: "Kubernetes tests (Kernel and ODK)"
+    #   env: 'TEST_MODE=kubernetes'
     - name: "Integration tests (Kernel with Kakfa/Zookeeper and Producer)"
       env: 'TEST_MODE=integration'
     - name: "Aether Suite tests (Kernel, UI, ODK and CouchDB-Sync)"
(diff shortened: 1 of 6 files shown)


admbtlr:
It seems odd to roll failed submission and untried submission into the same state is it possible to spit them into separate states? And also to maybe align the state names a bit more: `SUBMISSION_SUCCESS`, `SUBMISSION_FAILURE` and `SUBMISSION_UNTRIED` maybe?

quesebifurcan (dismissed):

shawnsarwar:
@adamvert. Changed possible Kafka statuses to the following.
```
class KafkaStatus(enum.Enum):
    SUBMISSION_PENDING = 1
    SUBMISSION_FAILURE = 2
    SUBMISSION_SUCCESS = 3
```

admbtlr (dismissed):

shawnsarwar:
Removed K8s tests.
@MrPink / @quesebifurcan please reinstate after helm fix is applied..

admbtlr (approved):

quesebifurcan (approved):

PR #335, opened 2018-08-27
fix: remote storage, file urls

quesebifurcan:
Fix file urls in kernel and odk.

* Add function `resolve_file_url()` to common module. 

For local development, the environment variable DJANGO_REMOTE_STORAGE is set to "filesystem" and files (attachments, media files etc.) are stored on the file system and served via nginx. Example URL: `http://odk.aether.local/media/<path-to-file>`. The hostname of the docker container is prepended to the media file url so that the resource can be served locally.
       
When the environment variable DJANGO_REMOTE_STORAGE is set to "s3" or "gcs", all file urls will be absolute. Example: `https://abcd.s3.amazonaws.com/<file-name>?AWSAccessKeyId=ABC&Signature=ABC%3D&Expires=1534775613`. This url is not modified.

* Adjust ODK manifest file creation so that it works with remote storage.

* Fix broken attachment links in kernel. On current develop, the attachment urls include port numbers (generated via HyperLinkedIdentityField), which doesn't play well with our nginx setup.

### Testing this pull request

This PR is unfortunately a bit tricky to test, since it involves ODK Collect, Gather, aether-kernel and aether-odk-module. We are working on an integration test for this part. Steps to take:
- Check out `fix/remote-storage/file-urls` in this repo
- Check out `develop` on https://github.com/eHealthAfrica/gather
- Modify both kernel and odk services in `<aether>/docker-compose-base.yml` to include the following details:
```
DJANGO_STORAGE_BACKEND: s3
BUCKET_NAME: <bucket-name>
AWS_ACCESS_KEY_ID: <access-key-id>
AWS_SECRET_ACCESS_KEY:  <secret-access-key>
```
- Go to the gather repo and run `docker-compose -f docker-compose-local.yml up`. This will bring up the most recent version of gather with aether checked out at `fix/remote-storage/file-urls`.
- Open `gather.local` in a browser and upload a form. Make sure that the form contains media files so we can check that file urls in gather and aether are working as expected.
- Download ODK Collect, point it at your local instance of aether-odk-module and download the new form. If everything works as expected, the manifest file was generated correctly.
- Fill out and submit the form. Please include at least one image/audio/video file.

At this point, we can verify that:
- The attachment links in Gather are working (click on each attachment link in the table)
- The attachment links in Aether are working (click on each attachment link in the response from the `/attachments` endpoint)
- The nested resources in Aether are working (click on each nested attachment link in the response from the `/entities` endpoint)


aether-odk-module/aether/odk/templatetags/odk_tags.py
```diff
@@ -0,0 +1,42 @@
...
+    return f'{scheme}://{host}{media_url}{file_path}'
+
+
+@register.simple_tag(takes_context=True)
+def get_file_url(context, media_file):
+    if settings.DJANGO_STORAGE_BACKEND == 'filesystem':
+        request = context['request']
+        return get_absolute_url(
+            scheme=request.scheme,
+            host=request.get_host(),
```
obdulia-losantos, line 35: what happens when you have a port?

obdulia-losantos:
Don't you want to implement the same solution for this method too?

https://github.com/eHealthAfrica/aether/pull/335/files#diff-255e5cdc695b537c2146d7b378c97eb5R273

aether-odk-module/aether/odk/templatetags/odk_tags.py
quesebifurcan: Good question -- the port is included. 
```
(Pdb) request.get_host()
'odk.aether.local:8002'
```

quesebifurcan:
Is there anything in `/formList` which refers to media files directly or indirectly? If so, then yes, we definitely need to modify it. To me, it looks like everything in there just refers to objects in the database. @obdulia-losantos

obdulia-losantos:
Sorry, you are right, I misunderstood what you were doing :woman_facepalming:

aether-common-module/aether/common/utils.py
```diff
@@ -0,0 +1,35 @@
...
+def resolve_file_url(url):
+    '''
+    Make all file urls absolute.
+    '''
+    if url.startswith('/'):
+        # For local development, the environment variable DJANGO_REMOTE_STORAGE is
+        # set to "filesytem" and files (attachments, media files etc.) are stored on
+        # the filesystem and served via nginx. Example URL:
+        # http://odk.aether.local/media/<path-to-file>.
+        return f'http://{settings.HOSTNAME}{url}'
```
obdulia-losantos: What happens with `https`? This is the expected protocol in production servers :grimacing:

aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -174,7 +174,7 @@ class AttachmentSerializer(DynamicFieldsMixin, serializers.ModelSerializer):
     name = serializers.CharField(allow_null=True, default=None)
     submission_revision = serializers.CharField(allow_null=True, default=None)
 
-    url = serializers.HyperlinkedIdentityField('attachment-detail', read_only=True)
+    url = serializers.CharField(read_only=True, source='attachment_url')
```
obdulia-losantos: I think this is a misunderstand, this field it's the API endpoint to update/delete/path this entry, not the file url itself :wink:

aether-common-module/aether/common/utils.py
quesebifurcan: In production, only the absolute url will be used and that one uses https (see comments).
Not great to hardcode the http scheme though, do you have any ideas?

aether-kernel/aether/kernel/api/serializers.py
quesebifurcan: Ah, I'll take a look.

obdulia-losantos: To avoid confusion we can use the `attachment_url` field (also in the `AttachmentSerializerNested` serializer). What do you think?

aether-common-module/aether/common/utils.py
obdulia-losantos: You can use the `SECURE_PROXY_SSL_HEADER` variable to detect the current protocol.

https://github.com/eHealthAfrica/aether/blob/develop/aether-common-module/aether/common/conf/settings.py#L244

obdulia-losantos: I would want to mention that in production WE are going to use the absolute urls, but not the rest of the world :grimacing:

aether-common-module/aether/common/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    Make all file urls absolute.
+    '''
+    if url.startswith('/'):
+        # For local development, the environment variable DJANGO_REMOTE_STORAGE is
+        # set to "filesytem" and files (attachments, media files etc.) are stored on
+        # the filesystem and served via nginx. Example URL:
+        # http://odk.aether.local/media/<path-to-file>.
+        ssl_header = settings.SECURE_PROXY_SSL_HEADER
+        scheme = ssl_header[1] if ssl_header else 'http'
+        return f'{scheme}://{settings.HOSTNAME}{url}'
```
obdulia-losantos, line 33: Now that the `HOSTNAME` is needed for this, could you mention it in the README?

obdulia-losantos (dismissed):

README.md
```diff
@@ -256,8 +256,10 @@ To get started on building solutions on Aether, an [aether-bootstrap](https://gi
 
 ## Deployment
 
-Set the `HOSTNAME` and `CAS_SERVER_URL` environment variables if you want to
-activate the UMS integration in each container.
+The environment variable `HOSTNAME` is required.
```
admbtlr: Isn't it only required on a local setup?

quesebifurcan: We have it here as well: https://github.com/eHealthAfrica/helm-charts/blob/504d12984f62f41386a794bb4b3ff18efbdce0a4/src/aether-odk/templates/deployment.yaml#L104-L105

admbtlr: Interesting.... do you know how that env var is used?

quesebifurcan: Updated README with details, see: https://github.com/eHealthAfrica/aether/pull/335/commits/73079d208a0571a68f4e3b11c313d678195993fd

README.md
```diff
@@ -256,8 +256,10 @@ To get started on building solutions on Aether, an [aether-bootstrap](https://gi
...
+
+Set the environment variable `CAS_SERVER_URL` if you want to activate the UMS integration 
+in all django projects. Note that the UMS integration [also requires `HOSTNAME` to be set](https://github.com/eHealthAfrica/ums/blob/0561eeea74ebc1d937d58e5f212014bfeed4d6e8/documentation/django-ums-client.md#quick-start):
```
obdulia-losantos: I would not include any link to a private repo in the README, being this a public repo :grimacing:

quesebifurcan: Yeah, forgot about that, removing now.

aether-common-module/aether/common/conf/settings.py
```diff
@@ -192,7 +192,7 @@
 CAS_VERSION = 3
 CAS_LOGOUT_COMPLETELY = True
 CAS_SERVER_URL = os.environ.get('CAS_SERVER_URL', '')
-HOSTNAME = os.environ.get('HOSTNAME', '')
+HOSTNAME = os.environ['HOSTNAME']
```
obdulia-losantos: You are going to kill me...

Why don't you move that check to here? This is the moment in which you really require it :thinking: (if you are not using UMS).
```python
if DJANGO_STORAGE_BACKEND == 'filesystem':
    DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
    HOSTNAME = os.environ['HOSTNAME']
elif ...
```

quesebifurcan: ok, I'll do that

obdulia-losantos: I'm just trying to reduce the list of required variables...

obdulia-losantos (dismissed):

admbtlr (dismissed):

README.md
```diff
@@ -269,6 +269,8 @@ If the response is `Always Look on the Bright Side of Life!!!`
 it's not possible to connect, on the other hand if the message is
 `Brought to you by eHealth Africa - good tech for hard places` everything goes fine.
 
+The environment variable `HOSTNAME` is required when `DJANGO_STORAGE_BACKEND` is set to "filesystem".
+
```
obdulia-losantos, line 273: Maybe a small mention here too :grimacing: 

https://github.com/eHealthAfrica/aether/pull/335/files#diff-04c6e90faac2675aa89e2176d2eec7d8R141

obdulia-losantos (approved):

admbtlr (approved):

PR #337, opened 2018-08-28
feat: add mapping sets to kernel

lordmallam:
Introduces the concept of mapping sets

JIRA Ticket: https://jira.ehealthafrica.org/browse/AET-340

Regression occuring on `ui` and `odk` modules.
This PR only addresses the kernel module, refactors to other modules coming soon in other PRs


aether-kernel/aether/kernel/api/models.py
```diff
@@ -99,14 +99,12 @@ class Meta:
...
-class Mapping(ExportModelOperationsMixin('kernel_mapping'), TimeStampedModel):
+class MappingSet(ExportModelOperationsMixin('kernel_mappingset'), TimeStampedModel):
 
     id = models.UUIDField(primary_key=True, default=uuid.uuid4)
     revision = models.TextField(default='1')
     name = models.CharField(max_length=50, null=False, unique=True)
-
-    definition = JSONField(blank=False, null=False)
-
+    input = JSONField(blank=False, null=False)
```
obdulia-losantos: Is this field `input` really necessary?

aether-kernel/aether/kernel/api/models.py
```diff
@@ -118,37 +116,87 @@ def __str__(self):
...
+    class Meta:
+        app_label = 'kernel'
+        default_related_name = 'mappings'
+        ordering = ['-modified']
+        indexes = [
+            models.Index(fields=['-modified']),
+        ]
+
+
+class MappingSetMapping(ExportModelOperationsMixin('kernel_mappingsetmapping'), TimeStampedModel):
```
obdulia-losantos: Couldn't we just have a many to many field in the `MappingSet`model? 

```python
    mappings = models.ManyToManyField(to=Mapping, blank=True)
```

aether-kernel/aether/kernel/api/models.py
```diff
@@ -160,6 +208,28 @@ class Meta:
         ]
 
 
+class SubmissionMapping(ExportModelOperationsMixin('kernel_submissionmapping'), TimeStampedModel):
```
obdulia-losantos: Maybe I don't see the whole picture, do we need this model? and why?

aether-kernel/aether/kernel/api/models.py
lordmallam: Being that contracts (schemas + mappings) are now logically grouped by the input, i think it should be stored if available. It also helps to recreate a complete pipeline (input + contracts) on the ui for better UX when adding a new contract to a set.

aether-kernel/aether/kernel/api/models.py
lordmallam: Not sure how django's `many-to-many` works, I just know a properly normalised db should not have a `many-to-many`. I will look into it

aether-kernel/aether/kernel/api/models.py
obdulia-losantos: In that case it shouldn't be mandatory. The ODK module doesn't have an input, but it creates the artefacts...

aether-kernel/aether/kernel/api/models.py
lordmallam: This is used to keep track of the actual mappings used during entity extraction for replay purposes. (mapping revisions included). Bear in mind that mappings can be added, edited or removed to a set after a submission has been made.

obdulia-losantos: But that should be reflected in the Entity and not in the Submission. The Submission must be "agnostic" and "ignorant" of all the triggered actions. Or am I wrong?

lordmallam: I don't think the submission should be completely ignorant of triggered actions or there will be no need to store submissions at all (other than for audit). Thus, the current implementation of `mapping` and `map_revision` in the submission model. In my view a replay will included but not limited to resubmitting the same submission even after edits have been made to corresponding mappings and still have the same result.

admbtlr: I tend to agree with @obdulia-losantos here, but I'm not sure - what do you think @quesebifurcan ?

quesebifurcan: Yes, agreed, the submission is akin to "raw" data and shouldn't (need) to store too much other information. I have no idea how replay will actually work, but for the most basic version, the submission would only need to know which mapping set it was submitted against. This is the way it currently works, at least.

aether-kernel/aether/kernel/api/models.py
quesebifurcan: Do we really need many-to-many relationships? If so, why?
This is the implementation I had in mind when we discussed it, might have misunderstood something though:
```python
class MappingSet(TimeStampedModel):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4) 

class Mapping(TimeStampedModel):                                                                                              
    ...other fields...                                                               
    mapping_set = models.ForeignKey(to=MappingSet)
```

obdulia-losantos: I thin that the point is that Mapping can be reused in another MappingSet, so the question is, do we want that? or better, are we going to use that feature?

aether-kernel/aether/kernel/api/models.py
lordmallam: Knowing the mapping set a submission was made against doesn't mean it knows the actual mapping(s) that was used during the extraction
I see the point of having the mapping refs on the entity, which currently is implicitly referenced through the submission. I am assuming that is why the `map_revision` is on the submission. But with the introduction of mapping sets that will not be the case because the mappings in a mapping set are mutable.
For a basic version without considering replay, i agree, submission should be ignorant. It is interesting though to know why `map_revision` is currently in the submission.

aether-kernel/aether/kernel/api/models.py
lordmallam: Correct @obdulia-losantos it was an acceptance criteria.. 😄 About using it.... 🤷‍♂️

obdulia-losantos:
Could you clean the migration files and create only one with the final change (without our changing minds workflow)?

obdulia-losantos:
I didn't mean ALL the migration files, only the ones you created in this PR :joy:

obdulia-losantos:
### Final data model revision :crossed_fingers: 

```text
Data model schema:

    +------------------+          +------------------+         +------------------+       +------------------+
    | Project          |          | MappingSet       |         | Submission       |       | Attachment       |
    +==================+          +==================+         +==================+       +==================+
    | id               |<-----+   | id               |<----+   | id               |<---+  | id               |
    | created          |      |   | created          |     |   | created          |    |  | created          |
    | modified         |      |   | modified         |     |   | modified         |    |  | modified         |
    | revision         |      |   | revision         |     |   | revision         |    |  | name             |
    | name             |      |   | name             |     |   | payload          |    |  | attachment_file  |
    | salad_schema     |      |   +::::::::::::::::::+     |   +::::::::::::::::::+    |  | md5sum           |
    | jsonld_context   |      +--<| project          |     +--<| mappingset       |    |  +::::::::::::::::::+
    | rdf_definition   |      |   +------------------+     |   | project(**)      |    +-<| submission       |
    +------------------+      |                            |   +------------------+    |  | submission_rev   |
                              |                            |                           |  +------------------+
                              |                            |                           |
                              |                            |                           |
                              |                            |                           |
    +------------------+      |   +------------------+     |   +------------------+    |  +------------------+
    | Schema           |      |   | ProjectSchema    |     |   | Mapping          |    |  | Entity           |
    +==================+      |   +==================+     |   +==================+    |  +==================+
    | id               |<--+  |   | id               |<--+ |   | id               |<-+ |  | id               |
    | created          |   |  |   | created          |   | |   | created          |  | |  | modified         |
    | modified         |   |  |   | modified         |   | |   | modified         |  | |  | revision         |
    | revision         |   |  |   | name             |   | |   | revision         |  | |  | payload          |
    | name             |   |  |   | mandatory_fields |   | |   | name             |  | |  | status           |
    | definition       |   |  |   | transport_rule   |   | |   | definition       |  | |  +::::::::::::::::::+
    | type             |   |  |   | masked_fields    |   | |   +::::::::::::::::::+  | +-<| submission       |
    +------------------+   |  |   | is_encrypted     |   | +--<| mappingset       |  +---<| mapping          |
                           |  |   +::::::::::::::::::+   +---<<| projectschemas   |       | mapping_rev      |
                           |  +--<| project          |   |     | project(**)      |   +--<| projectschema    |
                           +-----<| schema           |   |     +------------------+   |   | project(**)      |
                                  +------------------+   +----------------------------+   +------------------+
```

aether-kernel/aether/kernel/api/filters.py
```diff
@@ -39,26 +39,71 @@ class Meta:
...
+
+    def mappingset_filter(self, queryset, name, value):
+        if is_uuid(value):
+            return queryset.filter(mappingset__pk=value)
+        else:
+            return queryset.filter(mappingset__name=value)
+
+    def projectschema_filter(self, queryset, name, value):
+        if is_uuid(value):
+            return queryset.filter(projectschema__pk=value)
```
obdulia-losantos: shouldn't it be `projectschemas__`?

admbtlr:
Can you change the name of this PR to something that fits the semantic commit convention please?
https://confluence.ehealthafrica.org/display/GCS/Github+Processes
Maybe something like `feat: add mapping sets to kernel`?

obdulia-losantos:
Could you include the new ERD in the models file? :art:

aether-kernel/aether/kernel/api/migrations/0016_auto_20180903_1512.py
```diff
@@ -0,0 +1,106 @@
...
+        ),
+        migrations.AddIndex(
+            model_name='mappingset',
+            index=models.Index(fields=['project', '-modified'], name='kernel_mapp_project_73242f_idx'),
+        ),
+        migrations.AddIndex(
+            model_name='mappingset',
+            index=models.Index(fields=['-modified'], name='kernel_mapp_modifie_46a12a_idx'),
+        ),
+    ]
```
obdulia-losantos: In this migration file you should migrate the data too. :nerd_face: 

If there were already mapping instances you need to recreate here their Mapping Sets. Maybe something like this:

```python
def migrate_current_mappings(apps, schema_editor):
    Mapping = apps.get_model('kernel', 'Mapping')
    MappingSet = apps.get_model('kernel', 'MappingSet')

    for mapping in Mapping.objects.all():
        # share the id and the name
        mappingset = MappingSet.objects.create(
            pk=mapping.pk,
            name=mapping.name,
            project=mapping.project,
        )
        mapping.mappingset=mappingset
        mapping.save()


class Migration(migrations.Migration):

    operations = [
        ...,
        migrations.RunPython(migrate_current_mappings, migrations.RunPython.noop),
    ]

```

aether-kernel/aether/kernel/api/migrations/0016_auto_20180903_1512.py
```diff
@@ -0,0 +1,106 @@
...
+        ),
+        migrations.AlterModelOptions(
+            name='mapping',
+            options={'ordering': ['-modified']},
+        ),
+        migrations.RemoveIndex(
+            model_name='mapping',
+            name='kernel_mapp_project_0cb3fd_idx',
+        ),
+        migrations.RemoveField(
```
obdulia-losantos: could you keep the `project` field in the Mapping model? It's redundant but needed to speed up searches and filters.

aether-kernel/aether/kernel/api/models.py
```diff
@@ -268,6 +264,40 @@ class Meta:
...
+    projectschemas = models.ManyToManyField(to=ProjectSchema)
+    is_active = models.BooleanField(default=True)
+    is_read_only = models.BooleanField(default=False)
+
+    def save(self, **kwargs):
+        entities = self.definition.get('entities')
+        if entities:
+            for entity in entities:
+                self.projectschemas.add(ProjectSchema.objects.get(pk=entities[entity]))
+        super(Mapping, self).save(**kwargs)
```
obdulia-losantos: Suggestion

```python
class Mapping(ExportModelOperationsMixin('kernel_mapping'), TimeStampedModel):
    ...

    # redundant but speed up queries
    project = models.ForeignKey(to=Project, on_delete=models.CASCADE, blank=True, null=True)

    def save(self, **kwargs):
        self.project = self.mappingset.project

        entities = self.definition.get('entities', {})
        for entity_pk in entities.values():
            # only project schemas in the same project
            self.projectschemas.add(ProjectSchema.objects.get(pk=entity_pk, project=self.project))

        super(Mapping, self).save(**kwargs)
```

aether-kernel/aether/kernel/api/migrations/0016_auto_20180903_1512.py
obdulia-losantos: Also with this operation you introduce the project schema values for each mapping :nerd_face:

aether-kernel/aether/kernel/api/migrations/0016_auto_20180904_1429.py
```diff
@@ -0,0 +1,129 @@
...
+            name='kernel_mapp_project_0cb3fd_idx',
+        ),
+        migrations.RemoveField(
+            model_name='submission',
+            name='map_revision',
+        ),
+        migrations.RemoveField(
+            model_name='submission',
+            name='mapping',
+        ),
```
obdulia-losantos: You cannot remove the `mapping` yet or the migration script will fail. This operation should be moved to the end :wink:

aether-kernel/aether/kernel/api/models.py
```diff
@@ -38,42 +38,41 @@
...
+    | name             |   |  |   | mandatory_fields |   | |   | name             |  | |  | status           |
+    | definition       |   |  |   | transport_rule   |   | |   | definition       |  | |  +::::::::::::::::::+
+    | type             |   |  |   | masked_fields    |   | |   | is_active        |  | +-<| submission       |
+    +------------------+   |  |   | is_encrypted     |   | |   | is_read_only     |  +---<| mapping          |
+                           |  |   +::::::::::::::::::+   | |   +::::::::::::::::::+       | mapping_rev      |
+                           |  +--<| project          |   | +--<| mappingset       |   +--<| projectschema    |
+                           +-----<| schema           |   +---<<| projectschemas   |   |   | project(**)      |
+                                  +------------------+   |     | project(**)      |   |   +------------------+
+                                                         |     +------------------+   |
+                                                         +----------------------------+
```
obdulia-losantos, line 75: :innocent:

aether-kernel/aether/kernel/api/models.py
```diff
@@ -268,6 +264,44 @@ class Meta:
...
+    is_active = models.BooleanField(default=True)
+    is_read_only = models.BooleanField(default=False)
+
+    # redundant but speed up queries
+    project = models.ForeignKey(to=Project, on_delete=models.CASCADE, blank=True, null=True)
+
+    def save(self, **kwargs):
+        self.project = self.mappingset.project
+        entities = self.definition.get('entities', {})
+        for entity_pk in entities.values():
```
obdulia-losantos: Reading this again I think we need to clear the field first `self.projectschemas.clear()` :grimacing:

aether-kernel/aether/kernel/api/models.py
```diff
@@ -268,6 +264,44 @@ class Meta:
...
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        app_label = 'kernel'
+        default_related_name = 'mappings'
+        ordering = ['-modified']
+        indexes = [
+            models.Index(fields=['-modified']),
+        ]
```
obdulia-losantos: As it was before...

```python
    class Meta:
        app_label = 'kernel'
        default_related_name = 'mappings'
        ordering = ['project__id', '-modified']
        indexes = [
            models.Index(fields=['project', '-modified']),
            models.Index(fields=['-modified']),
        ]
```

aether-kernel/aether/kernel/api/tests/utils/generators.py
```diff
@@ -83,6 +83,9 @@ def generate_project(
         projectschema_field_values=None,
         submission_field_values=None,
         entity_field_values=None,
+        mappingset_field_values=None,
+        mappingsetmapping_field_values=None,
+        submissionmapping_field_values=None
```
obdulia-losantos: are you using the last two ones?

aether-kernel/aether/kernel/api/migrations/0016_auto_20180904_1429.py
```diff
@@ -0,0 +1,129 @@
...
+            },
+        ),
+        migrations.AlterModelOptions(
+            name='mapping',
+            options={'ordering': ['-modified']},
+        ),
+        migrations.RemoveIndex(
+            model_name='mapping',
+            name='kernel_mapp_project_0cb3fd_idx',
+        ),
```
obdulia-losantos: You can delete these two operations because you kept the Mapping indices and ordering.

```python
        migrations.AlterModelOptions(
            name='mapping',
            options={'ordering': ['-modified']},
        ),
        migrations.RemoveIndex(
            model_name='mapping',
            name='kernel_mapp_project_0cb3fd_idx',
        ),
```

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -606,40 +606,45 @@ def extract_create_entities(submission_payload, mapping_definition, schemas):
...
+                return
+            # Get the schema of the projectschema
+            project_schemas = {
+                name: models.ProjectSchema.objects.get(pk=_id) for name, _id in
+                entity_ps_ids.items()
+            }
+            schemas = {
+                name: ps.schema.definition for name, ps in
+                project_schemas.items()
+            }
```
obdulia-losantos: Could you reformat these lines?
Instead of 
```python
{
    name: func(value) for value in 
    array
}
```
it's more readable like this
```python
{
    name: func(value)
    for value in array
}
```

aether-kernel/aether/kernel/api/models.py
```diff
@@ -279,6 +315,8 @@ class Entity(ExportModelOperationsMixin('kernel_entity'), models.Model):
 
     projectschema = models.ForeignKey(to=ProjectSchema, on_delete=models.SET_NULL, null=True)
     submission = models.ForeignKey(to=Submission, on_delete=models.SET_NULL, blank=True, null=True)
+    mapping = models.ForeignKey(to=Mapping, on_delete=models.SET_NULL, null=True)
+    mapping_revision = models.TextField(blank=False, null=False)
 
     # redundant but speed up queries
     project = models.ForeignKey(to=Project, on_delete=models.SET_NULL, blank=True, null=True)
```
obdulia-losantos, line 322: This is not in this PR scope but I think that a better definition could be with the CASCADE option
`project = models.ForeignKey(to=Project, on_delete=models.CASCADE, blank=True, null=True)`

Otherwise we have orphan entities if the project is deleted, and with the integration tests means, that we are not properly cleaning the database :scream:

aether-kernel/aether/kernel/api/models.py
```diff
@@ -279,6 +315,8 @@ class Entity(ExportModelOperationsMixin('kernel_entity'), models.Model):
 
     projectschema = models.ForeignKey(to=ProjectSchema, on_delete=models.SET_NULL, null=True)
     submission = models.ForeignKey(to=Submission, on_delete=models.SET_NULL, blank=True, null=True)
+    mapping = models.ForeignKey(to=Mapping, on_delete=models.SET_NULL, null=True)
```
obdulia-losantos, line 318: With this new field maybe you can include it the project check in the save method too :thinking:

aether-kernel/aether/kernel/api/models.py
lordmallam: From what i've gathered so far. I assume entities are not to be automatically deleted at any point. Am i missing something?

obdulia-losantos: No, it's part of my eternal _how to create even more confusion_ task... just wondering if it makes sense to have orphan entities, linked to nothing at all.

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -624,8 +624,8 @@ def run_entity_extraction(submission):
                 entity_ps_ids.items()
```
obdulia-losantos: same change here :grimacing:

aether-kernel/aether/kernel/api/models.py
lordmallam: Mappings are no longer directly linked to project. not sure the check is necessary.

obdulia-losantos: Yes, they still are (with that redundant field) :wink: But it's not necessary to include it the check...

obdulia-losantos:
I understand that part of the test must fail but this one is easy to fix :grimacing: 

https://travis-ci.com/eHealthAfrica/aether/jobs/143926977#L1375

`/code/./aether/kernel/api/utils.py:623:63: W291 trailing whitespace`

obdulia-losantos (approved):
:crossed_fingers:

PR #339, opened 2018-08-30
Remove readonly user creation

tooxie:
This code will instead be ran as part of an init container by the helm
chart. This is because the user provided for the kernel to connect to
the database does not have the necessary credentials to create new
users.


obdulia-losantos:
Should we remove the rest of related scripts, variables too? Like docker-compose entries, credentials script...

tooxie:
We need to keep part of this logic here, so I will close this PR for now.

PR #340, opened 2018-08-30
Remove role-creation logic from the `create_readonly_user.py` script

tooxie:


aether-kernel/sql/grant_readonly_user.py
```diff
@@ -72,7 +61,7 @@ def main():
         role = os.environ['KERNEL_READONLY_DB_USERNAME']
         password = os.environ['KERNEL_READONLY_DB_PASSWORD']
```
obdulia-losantos: `password` is not longer needed.

aether-kernel/sql/grant_readonly_user.py
```diff
@@ -72,7 +61,7 @@ def main():
...
         cursor = conn.cursor()
-        query = sql.SQL(CREATE_READONLY_USER).format(
+        query = sql.SQL(GRANT_READONLY_USER).format(
             database=sql.Identifier(dbname),
             role=sql.Identifier(role),
             rolename=sql.Literal(role),
```
obdulia-losantos: `rolename` either

tooxie: Good catch! 👍 🙌

tooxie:
@obdulia-losantos @quesebifurcan as stated in the PR we have to move the user creation logic out of this module because the user running this code won't have permissions to create other users. The problem is that when this is run in the tests, the user won't exist so it will fail, as it did already. How could we solve this issue?

aether-kernel/entrypoint.sh
```diff
@@ -75,7 +75,7 @@ setup () {
     ./manage.py migrate --noinput
 
     # Create readonly database user
-    python /code/sql/create_readonly_user.py
+    python /code/sql/grant_readonly_user.py
```
obdulia-losantos, line 78: Remove this line and execute the script outside the container. The readonly user is not necessary to start the app.
Do not change the command, leave it as it is now. We can fix it afterwards.

tooxie: What do you mean by "outside the container"? Where exactly? This has to be run after the migrations but before the app starts. Besides, the app is the one that has the knowledge about the tables involved, so I would prefer to have this code within aether and not in k8s.

tooxie:
We'll go with @obdulia-losantos proposal of removing the `setup` command from `start` and instead will be run independently by k8s. Closing.

PR #341, opened 2018-09-03
feat: test producer

shawnsarwar:
Added some non end-to-end test structure for the Aether Producer.
Made naming more conventional. This caused some files that were only moved to be shown as diff'd. Primarily look at the (few) tests and the changes made to integration testing and travis.


aether-producer/setup.py
```diff
@@ -0,0 +1,33 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+from setuptools import setup
+setup(
+    name='aether_producer',
+    author='Shawn Sarwar',
+    author_email='shawn.sarwar@ehealthafrica.org',
+    decription='Kafka Producer for Aether',
+    version='0.9.0',
```
obdulia-losantos: Maybe you prefer to read this value from the `VERSION` file...

aether-producer/producer/__init__.py
```diff
@@ -15,3 +15,644 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
```
shawnsarwar, line 18: You can ignore this file. The reason it shown as 100% diff is we moved from aether_producer.py to __init_.py which already existed to make the name spacing of the module work more cleanly. No substantive changes were made to producer function.

aether-producer/producer/aether_producer.py
```diff
@@ -1,651 +0,0 @@
-# Copyright (C) 2018 by eHealth Africa : http://www.eHealthAfrica.org
```
shawnsarwar, line 1: This was moved -> __init__.py

quesebifurcan (approved):

PR #342, opened 2018-09-03
fix: Remove `create_readonly_user.py` command from `setup`

tooxie:


aether-kernel/entrypoint.sh
```diff
@@ -138,7 +138,8 @@ case "$1" in
     ;;
 
     start )
-        setup
+        # check if required environment variables were set
+        ./conf/check_vars.sh
```
obdulia-losantos: Thinking on this issue a little bit more... could you use a condition here in order to keep local development working?

```bash
if [ "IS THIS A PROD DEPLOYMENT" ]; then
    ./conf/check_vars.sh
else
    setup
fi
```

Otherwise this is going to stop working as expected :scream:

tooxie: Correct me if I'm wrong but the `start` command, afaik, is not used for local development, and if you did it would only fail on a fresh install of aether.  What do you think if instead I add a mention to the `README` file stating explicitly that `start_dev` should be used for local development (which is the current state of things anyway)?

I would prefer that over adding a new flag or environment variable that we will have to keep track of. What do you think?

obdulia-losantos: The migration command should be run every time there is a change in the data model not only with a clean start. But yes, this is something that doesn't, or shouldn't, happen quite often.

`aether-bootstrap` and `gather-deploy` use the `start` command, if you change this we should spread this change to them and, also change the documentation (gh-pages).

@adamvert what do you think?

tooxie:
After discussing it with @obdulia-losantos she proposed that we simply call the script directly from k8s instead of modifying the code here. Will try that.

admbtlr:
That sounds like a sensible compromise.

scripts/test_integration.sh
```diff
@@ -64,6 +64,7 @@ $DC_TEST up -d producer-test
 
 echo "_____________________________________________ Starting Integration Tests"
 build_container integration
+python /code/sql/create_readonly_user.py
```
obdulia-losantos: `$DC_TEST run kernel-test eval python /code/sql/create_readonly_user.py` :wink:

test-aether-integration-module/Dockerfile
```diff
@@ -18,6 +18,7 @@ WORKDIR /code
 COPY ./conf/pip /code/conf/pip
 RUN pip install -f /code/conf/pip/dependencies -r /code/conf/pip/requirements.txt
 
+COPY ../aether-kernel/sql /code
```
obdulia-losantos: I think this line is not necessary :thinking:

admbtlr:
Can you change the name of this PR to something that fits the semantic commit convention please?
https://confluence.ehealthafrica.org/display/GCS/Github+Processes
Maybe something like `chore: remove create_readonly_user from setup`?

obdulia-losantos (approved):

PR #343, opened 2018-09-04
chore: fix setup and start scripts

obdulia-losantos:


README.md
```diff
@@ -167,28 +167,31 @@ The tests clean up will **DELETE ALL MAPPINGS!!!**
...
 
 ```bash
-docker-compose up --build    # this will update the containers if needed
+./scripts/docker_start.sh [options] [app]
 ```
 
-or
+Arguments:
+  `--force` | `-f`  will kill running containers
+  `--killl` | `-k`  will kill running containers
```
admbtlr: typo?

scripts/docker_start.sh
```diff
@@ -25,22 +25,23 @@
 #
 # arguments:
 #   --force | -f  will kill running containers
+#   --killl | -k  will kill running containers
```
admbtlr: typo

README.md
admbtlr: and what's the difference between `-f` and `-k`?

README.md
```diff
@@ -167,28 +167,31 @@ The tests clean up will **DELETE ALL MAPPINGS!!!**
...
+  `--build` | `-b`  will build containers before starting them
 
-```bash
-./scripts/docker_start.sh    # starts all
-```
+  `app`
+    Expected values: `kernel`, `odk`, `ui`, `couchdb-sync` or `sync`.
+    Another value will start all containers.
```
admbtlr: Actually "Any other...." is slightly more natural here than "Another..."

admbtlr (changes requested):

admbtlr:
Reading the README, I'm a bit confused: should I be using `./scripts/docker_start.sh` rather than `docker-compose up`? Both are mentioned, and it's not immediately clear to me why...

obdulia-losantos:
`docker_start.sh` takes care of the order in which the containers must be started to work smoothly and without health errors, missing resources...

`docker-compose` starts the containers in the same order they appear in the `docker-compose.yml` file.

We already know that this is not enough :sob:

admbtlr:
I guess this is nothing to do with this PR, but when I run it I get 

```couchdb-sync-rq_1  | psycopg2.OperationalError: FATAL:  database "couchdb-sync" does not exist```

admbtlr:
Also 
image

obdulia-losantos:
As expected if you are not running gather :sweat_smile:

obdulia-losantos:
I'll check the `couchdb-sync-rq` error :see_no_evil:

admbtlr:
Ah, true, good point

obdulia-losantos:
Hopefully this time works :crossed_fingers:

admbtlr:
And now nginx never starts up 💢

admbtlr (approved):
Everything works fine on a more powerful machine 🎉

obdulia-losantos:
:tada:

PR #348, opened 2018-09-06
chore: Rename user and password variables to identify the role they belong to

tooxie:


aether-kernel/sql/create_readonly_user.py
```diff
@@ -59,24 +66,33 @@ def main():
...
+        ro_user = os.environ['KERNEL_READONLY_DB_USERNAME']
+        ro_password = os.environ['KERNEL_READONLY_DB_PASSWORD']
         cursor = conn.cursor()
         query = sql.SQL(CREATE_READONLY_USER).format(
             database=sql.Identifier(dbname),
-            role=sql.Identifier(role),
-            rolename=sql.Literal(role),
-            password=sql.Literal(password),
+            role=sql.Identifier(ro_user),
+            rolename=sql.Literal(ro_user),
```
obdulia-losantos: Because you are already refactoring this... could you remove the `{rolename}` option? (being the same as `{role}` it's just error-prone) :grimacing:

obdulia-losantos:
Taken from: https://tableplus.io/blog/2018/04/postgresql-how-to-create-read-only-user.html

Even better

```sql
DO $$
BEGIN
  IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = {role})
  THEN
      CREATE ROLE {role} WITH LOGIN ENCRYPTED PASSWORD {password}
      INHERIT NOSUPERUSER NOCREATEDB NOCREATEROLE NOREPLICATION;
  END IF;
END
$$ LANGUAGE plpgsql;

REVOKE ALL PRIVILEGES ON DATABASE {database} FROM {role} CASCADE;

GRANT CONNECT ON DATABASE {database} TO {role};
GRANT USAGE  ON SCHEMA public TO {role};
GRANT SELECT ON ALL TABLES IN SCHEMA public TO {role};
ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO {role};
```

aether-kernel/sql/create_readonly_user.py
tooxie: Good catch! I completely missed that.

tooxie:
@obdulia-losantos the thing is that we don't want to provide full read access to all tables, just to those it needs. Is there something else proposed in that snippet? I can't see any other change.

obdulia-losantos (dismissed):

obdulia-losantos:
No that's all.

aether-kernel/sql/create_readonly_user.py
```diff
@@ -28,55 +28,71 @@
...
+DEBUG = os.environ.get('DEBUG')
+LEVEL = logging.DEBUG if DEBUG else logging.WARNING
+
+logging.basicConfig(level=LEVEL)
+logger = logging.getLogger(__name__)
+
+# Create a readonly user with username "{role_id}" if none exists.
 # Grant read permission for relevant tables.
 CREATE_READONLY_USER = '''
 DO $$
```
obdulia-losantos, line 44: Another option...

```sql

DROP ROLE IF EXISTS {role_id};
CREATE ROLE {role_id} WITH 
    LOGIN
    ENCRYPTED PASSWORD {password}
    INHERIT
    NOSUPERUSER
    NOCREATEDB
    NOCREATEROLE
    NOREPLICATION;

```

What do you think?

obdulia-losantos (dismissed):

aether-kernel/sql/create_readonly_user.py
tooxie: We can't go in that direction because we would be dropping a role that's being used by some other code. And this script will execute every time a new pod comes up (which happens very often). It's like SQL Russian roulette 😄

What don't you like about the current approach?

obdulia-losantos: If the user already exists, you are not updating the password nor privileges...

```sql
DO $$
BEGIN
  IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = {role_literal})
  THEN
      CREATE USER {role_id};
  END IF;
END
$$ LANGUAGE plpgsql;

ALTER USER {role_id} WITH
    NOBYPASSRLS
    NOCREATEDB
    NOCREATEROLE
    NOINHERIT
    NOREPLICATION
    NOSUPERUSER
    ENCRYPTED PASSWORD {password}
;

REVOKE ALL PRIVILEGES ON DATABASE {database} FROM {role_id} CASCADE;
GRANT CONNECT ON DATABASE {database} TO {role_id};
GRANT USAGE ON SCHEMA public TO {role_id};
GRANT SELECT ON kernel_entity TO {role_id};
GRANT SELECT ON kernel_mapping TO {role_id};
GRANT SELECT ON kernel_projectschema TO {role_id};
GRANT SELECT ON kernel_schema TO {role_id};
```

tooxie: You are right, good point. I wonder if this is the right approach, though, because if the password has to change and we let it up to this script to update it, it would lead to noticeable downtime. The moment we deploy a version with a new password there will be versions still running with the old password and those would stop working. I think we need a release strategy for password rotation instead.

obdulia-losantos: Change the password it's also a way to detect running and hidden instances...

tooxie: If we have instances running that we are not aware of, I think our biggest problem is not detecting which ones they are but figuring out how we ended up in that state. And we can't make users pay with downtime for our mistakes.

obdulia-losantos: Another solution...

```python
CREATE_READONLY_USER = '''
DO $$
BEGIN
  IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = {role_literal})
  THEN
      CREATE USER {role_id} ENCRYPTED PASSWORD {password};
  END IF;
END
$$ LANGUAGE plpgsql;

ALTER USER {role_id} WITH
    NOBYPASSRLS
    NOCREATEDB
    NOCREATEROLE
    NOINHERIT
    NOREPLICATION
    NOSUPERUSER    
;

...
'''

CHANGE_READONLY_PASSWORD = '''
ALTER USER {role_id} WITH ENCRYPTED PASSWORD {password};
'''


def main():
    # ...

    with psycopg2.connect(**postgres_credentials) as conn:
        # ...
        cursor.execute(query)

        change_password = bool(os.environ.get('CHANGE_RO_PASS', False))
        if change_password:
            cursor.execute(
                CHANGE_READONLY_PASSWORD.format(
                    role_id=sql.Identifier(ro_user),
                    password=sql.Literal(ro_password),
                )
            )


```

obdulia-losantos: On the other hand, we should use always the same password in the script, and if we want to change it... it's because something has happened :scream:

tooxie: Sure, we can log in to the DB and change it manually there if it's an emergency, there's no need to wait for PR + publishing + deployment. Again, I don't want to minimize the problem you are bringing up, my point is that this script doesn't seem to be the place to solve it. I personally don't want pods to be concerned about state changes, they should just come up and assume the world starts when they come to existence.

If the password would have to change, I would have a process in place that:
* Tells running pods to finish processing their work and exit.
* Brings up new pods with a **new** username and password.
* Deletes the old user when all pods have died.

That way we have no downtime and the process is automated. What do you think? Maybe we can bring a third opinion to the discussion like @quesebifurcan?

obdulia-losantos: Then something like this?

```sql
DO $$
BEGIN
  IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = {role_literal})
  THEN
      CREATE USER {role_id} WITH ENCRYPTED PASSWORD {password};
  END IF;
END
$$ LANGUAGE plpgsql;

ALTER USER {role_id} WITH
    NOBYPASSRLS
    NOCREATEDB
    NOCREATEROLE
    NOINHERIT
    NOREPLICATION
    NOSUPERUSER
;
```

This will never change the password but update the user options :thinking:

aether-kernel/sql/create_readonly_user.py
```diff
@@ -46,11 +46,19 @@
   IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles WHERE rolname = {role_literal})
   THEN
       CREATE ROLE {role_id} WITH LOGIN ENCRYPTED PASSWORD {password}
```
obdulia-losantos: Don't forget the `;`

obdulia-losantos (dismissed):

tooxie:
@obdulia-losantos Error:
```
psycopg2.ProgrammingError: must be superuser to alter superusers
```

tooxie:
@obdulia-losantos I removed that change, we can bring it back when we have the need to update the user's privileges if you don't mind.

obdulia-losantos (approved):

PR #345, opened 2018-09-05
fix: clean and prepare couchdb-sync

obdulia-losantos:
- Improve and clean documentation.
- Introduce *i18n* in messages.
- Remove kernel project fixture for tests, use `artefacts` endpoint to create kernel dependencies.
- Introduce `Schema` model to keep the relationship between the mobile schema and the documents submission to Kernel (remove name dependency with kernel mapping).

PR #350, opened 2018-09-06
fix: artefacts generation with mapping sets

obdulia-losantos:
CouchDB-Sync tests would pass with #345 merged in this branch :grimacing:


aether-common-module/aether/common/kernel/utils.py
```diff
@@ -145,16 +145,17 @@ def get_data(url):
     return results
 
 
-def submit_to_kernel(submission, mapping_id, submission_id=None):
+def submit_to_kernel(submission, submission_fk, submission_id=None):
```
lordmallam: Why not just call it `mappingset_id` instead of `submission_fk`? it got me confused. We also have `project` as a `fk`

aether-common-module/aether/common/kernel/tests/test_utils.py
```diff
@@ -140,12 +140,12 @@ def test__test_connection_get_fail(self, mock_get, mock_head):
             )
 
     @mock.patch.dict('os.environ', AETHER_ENV_MOCK)
-    def test_submit_to_kernel__without_mapping_id(self):
+    def test_submit_to_kernel__without_submission_fk(self):
```
lordmallam: Same thing with `mappingset_id` naming here

aether-common-module/aether/common/kernel/utils.py
obdulia-losantos: Yesterday it was `mapping_id`, now is `mappingset_id` and tomorrow... ?

I want to reduce kernel data model dependency in the external modules.

But I don't mind to change this back :woman_shrugging:

aether-kernel/aether/kernel/api/models.py
```diff
@@ -105,12 +105,14 @@ class MappingSet(ExportModelOperationsMixin('kernel_mappingset'), TimeStampedMod
...
+
     input = JSONField(null=True, blank=True)
+
     project = models.ForeignKey(to=Project, on_delete=models.CASCADE)
 
     @property
-    def definition_prettified(self):
-        return json_prettified(self.definition)
+    def input_prettified(self):
+        return json_prettified(self.input)
```
lordmallam, line 115: 👍

aether-common-module/aether/common/kernel/utils.py
admbtlr: I agree, this also got me confused. The move from `mapping` to `mappingset` is a major change, so I think it's OK to reflect that in the code rather than to abstract away from it. It's certainly not something that should happen very often.

obdulia-losantos: :+1:

lordmallam (approved):
LGTM

PR #354, opened 2018-09-07
fix(models): use __str__ only with own fields

obdulia-losantos:


obdulia-losantos:
I couldn't believe it but with the `__str__` method it took almost 4 minutes to load the entities page, without it the expected time.

image

quesebifurcan (approved):
+1

admbtlr (approved):
😮

PR #355, opened 2018-09-10
chore: cleaning, refactor, i18n and docs

obdulia-losantos:
_Fall cleaning_


admbtlr:
I don't fully understand what this is doing - is there a fix for the GRID issues buried in here somewhere? 

Also, do we really want to internationalise console messages?

aether-common-module/aether/common/conf/urls.py
```diff
@@ -112,9 +124,12 @@ def generate_urlpatterns(token=False, kernel=False):  # pragma: no cover
...
         if settings.TESTING:
-            msg = 'Environment variable "AETHER_KERNEL_URL_TEST" is not set'
+            msg = _('Environment variable "AETHER_KERNEL_URL_TEST" is not set')
         else:
-            msg = 'Environment variable "AETHER_KERNEL_URL" is not set'
+            msg = _('Environment variable "AETHER_KERNEL_URL" is not set')
         assert get_kernel_server_url(), msg
 
+        msg_token = _('Environment variable "AETHER_KERNEL_TOKEN" is not set')
+        assert get_kernel_server_token(), msg_token
```
admbtlr, line 133: But isn't this an odd place to put these asserts? I wouldn't think to look for them in `generate_urlpatterns`

obdulia-losantos: But it's the only place where you can indicate that you need access to kernel. I agree that's a weird place and I want to refactor this and move it to another place but... no time so far. :(

obdulia-losantos:
No, this is just what it says, a little bit of cleaning, refactor and missing docs. But it also includes the fix for the `AETHER_KERNEL_TOKEN`.

Maybe we don't need to translate the console message but it doesn't hurt to have the option of doing that. I can revert that change if you prefer.

admbtlr (approved):
OK, fair enough, let's go with this :)

PR #358, opened 2018-09-11
fix(common): improve LOGGING setting

obdulia-losantos:
This will reject blank values `""` too

### DIFF
--- a/aether-common-module/aether/common/conf/settings.py
+++ b/aether-common-module/aether/common/conf/settings.py
@@ -29,14 +29,6 @@ DEBUG = bool(os.environ.get('DEBUG'))
 TESTING = bool(os.environ.get('TESTING'))
 SECRET_KEY = os.environ['DJANGO_SECRET_KEY']
 
-
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
-if DEBUG:
-    logger.setLevel(logging.DEBUG)
-if TESTING:
-    logger.setLevel(logging.CRITICAL)
-
 LANGUAGE_CODE = 'en-us'
 TIME_ZONE = 'UTC'
 USE_I18N = True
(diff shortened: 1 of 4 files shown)


admbtlr:
We're in danger of scattering these checks everywhere...

aether-common-module/aether/common/conf/settings.py
```diff
@@ -246,12 +303,21 @@
 if DJANGO_STORAGE_BACKEND == 'filesystem':
     DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
     HOSTNAME = os.environ['HOSTNAME']
+    if not HOSTNAME:
+        raise RuntimeError('Missing HOSTNAME')
```
MrPink: can you add `environment variable` after `HOSTNAME` to make it explicit, same with the others

obdulia-losantos: Done!

aether-common-module/aether/common/conf/settings.py
```diff
@@ -246,12 +303,21 @@
...
     DEFAULT_FILE_STORAGE = 'storages.backends.s3boto3.S3Boto3Storage'
     AWS_STORAGE_BUCKET_NAME = os.environ['BUCKET_NAME']
+    if not AWS_STORAGE_BUCKET_NAME:
+        logger.critical('Missing BUCKET_NAME environment variable!')
+
 elif DJANGO_STORAGE_BACKEND == 'gcs':
     DEFAULT_FILE_STORAGE = 'storages.backends.gcloud.GoogleCloudStorage'
     GS_BUCKET_NAME = os.environ['BUCKET_NAME']
+    if not GS_BUCKET_NAME:
+        logger.critical('Missing BUCKET_NAME environment variable!')
```
obdulia-losantos: @adamvert and @MrPink what do you prefer? To raise a `RuntimeError` or to send the critical log to the console/sentry?

admbtlr: I guess a `RuntimeError` is better, since we want it to die

quesebifurcan: Can we do both?

obdulia-losantos: Always!

admbtlr (approved):

PR #361, opened 2018-09-11
feat: (ui) mappingsets to pipelines 

lordmallam:
Introduces the concept of pipeline ==> contracts to the ui.
Refactors ui model
JIRA Ticket: [AET-345](https://jira.ehealthafrica.org/browse/AET-345)


aether-kernel/aether/kernel/api/models.py
```diff
@@ -146,7 +146,9 @@ def payload_prettified(self):
         return json_prettified(self.payload)
 
     def __str__(self):
-        return '{} - {}'.format(str(self.mappingset), self.id)
+        return '%s - %s' % (
+            str(self.mappingset), str(self.id),
+        )
```
obdulia-losantos: As we already know... this method is a bad idea :scream:

aether-kernel/aether/kernel/api/models.py
```diff
@@ -306,6 +310,10 @@ class Meta:
         ]
 
 
+# @receiver(post_save, sender=Mapping, dispatch_uid="update_projectschemas_list")
+# def update_projectschemas(sender, instance, **kwargs):
```
obdulia-losantos: :roll_eyes: :wastebasket:

aether-kernel/aether/kernel/api/models.py
```diff
@@ -273,21 +275,23 @@ class Mapping(ExportModelOperationsMixin('kernel_mapping'), TimeStampedModel):
...
-        self.projectschemas.clear()
+        ps_list = []
         for entity_pk in entities.values():
-            self.projectschemas.add(ProjectSchema.objects.get(pk=entity_pk, project=self.project))
-
-        super(Mapping, self).save(**kwargs)
+            ps = ProjectSchema.objects.get(pk=entity_pk)
+            if ps:
+                ps_list.append(ps)
+        self.projectschemas.add(*ps_list)
```
obdulia-losantos: :flushed: 

If you move up `super(Mapping, self).save(*args, **kwargs)` next changes are not going to be saved, are they?

On the other hand, are you removing the useless `project_schema` entries?

This `.get(pk=entity_pk, project=self.project))` is needed to ensure that you are not using project schema entries that belong to other projects.

Note: the method `get` will raise an error `ObjectDoesNotExist` if it does not exist, you don't need the `if ps:` condition. :bomb:

aether-kernel/aether/kernel/api/migrations/0016_auto_20180907_1214.py
```diff
@@ -113,9 +128,4 @@ class Migration(migrations.Migration):
             model_name='mappingset',
             index=models.Index(fields=['-modified'], name='kernel_mapp_modifie_46a12a_idx'),
         ),
```
obdulia-losantos, line 130: why are you rewriting the migration file?

aether-ui/aether/ui/api/models.py
```diff
@@ -23,19 +23,40 @@
...
+    # this is a reference to the linked kernel mappingset
+    mappingset = models.UUIDField(null=True, blank=True)
+
+    def __str__(self):
+        return self.name
+
+    def save(self, *args, **kwargs):
+        read_only_contracts = self.contracts.filter(is_read_only=True)
+        if read_only_contracts:
+            raise Exception('Input is readonly for this pipeline')
```
obdulia-losantos: Why are you raising this error? Shouldn't you check first that you are changing the input? What happens if you are just changing the `mappingset` value?

aether-kernel/aether/kernel/api/models.py
lordmallam: `object.add()` is a commit action to the db. Save() is moved up because the mapping has to be saved with an id generated before you can add a m2m object. (here)[https://docs.djangoproject.com/en/2.1/topics/db/examples/many_to_many/]. The only issue with this is when using the rest-framework form, project-schemas must be explicitly selected from the list

obdulia-losantos: This resolves my first comment, what about the rest? :thinking: 

To avoid problem with forms you need the `blank=True` option, right!

aether-kernel/aether/kernel/api/migrations/0016_auto_20180907_1214.py
lordmallam: Because due to the `migrate_current_mappings_to_mappingsets` method being ran. The migration order had to be resorted

aether-ui/aether/ui/api/models.py
lordmallam: I assume a readonly mapping set means exactly that. All its properties included. In the case of the ui, same is true for the pipeline if any of its nested contracts are marked as readonly, and changes should only be made on the contract level. So as not to cause any form of regression to any apps dependent on the mapping set.

aether-kernel/aether/kernel/api/models.py
lordmallam: Not exactly, in the case of forms, all previous m2m entries even after saved manually are cleared and only those selected using the form are added before a commit. 
https://stackoverflow.com/a/1925784/4032647

aether-ui/aether/ui/api/models.py
```diff
@@ -72,18 +93,26 @@ class Pipeline(ExportModelOperationsMixin('ui_pipeline'), TimeStampedModel):
     output = JSONField(blank=True, null=True, editable=False)
     kernel_refs = JSONField(blank=True, null=True, editable=False)
     published_on = models.DateTimeField(blank=True, editable=False, null=True)
+    is_active = models.BooleanField(default=True)
+    is_read_only = models.BooleanField(default=False)
+
+    pipeline = models.ForeignKey(to=Pipeline, on_delete=models.CASCADE, related_name='contracts')
```
obdulia-losantos: Note: you don't need to indicate the option `related_name`if you have the Meta option `default_related_name`. ;-)

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
```diff
@@ -136,6 +136,35 @@ const parsePipeline = (pipeline) => {
...
+      pipeline.mapping = []
+      pipeline.mapping_errors = []
+      pipeline.output = {}
+      pipeline.kernel_refs = {}
+      pipeline.published_on = null
+      pipeline.contract_id = null
+    }
+    delete pipeline.contracts
+    return pipeline
+  }
```
obdulia-losantos, line 167: I can imagine that this (take only the first Contract and not all) will be fixed in further PRs, right?

aether-ui/aether/ui/api/utils.py
```diff
@@ -396,34 +388,61 @@ def convert_mappings(mapping_from_kernel):
...
-        kernel_refs={
-            'project': kernel_object['project'],
-            'schema': entity_types['ids'],
-            'projectschema': kernel_object['definition']['entities'],
-            'mapping': kernel_object['id']
-        }
-    )
+def create_new_pipeline_from_kernel(mappingset):
+    new_pipeline = linked_pipeline_object('mappingset', mappingset['id'])
+    mappings = kernel_data_request('mappings/?mappingset={}'.format(mappingset['id']))['results']
```
obdulia-losantos, line 402: Are you aware that kernel paginate its results and you are not checking that here?

aether-ui/aether/ui/assets/apps/pipeline/redux.jsx
lordmallam: Very correct, this is the "intermediate" mapping set solution as discussed to work with the current UI

aether-ui/aether/ui/api/utils.py
lordmallam: 🤔

obdulia-losantos: I would suggest to restrict this PR to the data model change, that looks fine from my point of view and fix the fetch from kernel in another one. Is this fine for you? @lordmallam

lordmallam: That's fine @obdulia-losantos

obdulia-losantos (approved):
let's fix the fetch from kernel in another PR ;-)

PR #362, opened 2018-09-12
chore: include HOSTNAME in settings

obdulia-losantos:
If `HOSTNAME` is not indicated in the docker-compose file, then docker will set the container id.

`http://7305e0f159d6/` without HOSTNAME
`http://example.com/` with `HOSTNAME=example.com`

Also the distinct count is needed to return the correct entities number to Gather.

### DIFF
--- a/docker-compose-base.yml
+++ b/docker-compose-base.yml
@@ -142,7 +142,7 @@ services:
     stdin_open: true
     tty: true
     environment:
-      # Uncomment thi line to enable single sign on if you use CAS
+      # Uncomment this line to enable single sign on if you use CAS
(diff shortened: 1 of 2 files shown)


docker-compose-base.yml
```diff
@@ -142,14 +142,14 @@ services:
     stdin_open: true
     tty: true
     environment:
-      # Uncomment these lines to enable single sign on if you use CAS
+      # Uncomment thi line to enable single sign on if you use CAS
```
admbtlr: typo

admbtlr (changes requested):

admbtlr (approved):

PR #363, opened 2018-09-12
chore: merge develop fixes into release-0.9

obdulia-losantos:


shawnsarwar:
Looks good. It'll be good to get on a fixed version so we don't keep running in to situations where we're using the RC and then it breaks...

shawnsarwar (approved):

admbtlr (approved):

PR #367, opened 2018-09-13
fix: updated swagger generation library

shawnsarwar:
Our openapi library was deprecated. We've replaced it and made sure that our versioning carries over ot the swagger api generation. We've also added a new API viewer at /{version}/swagger.


obdulia-losantos (approved):
:crocodile:

PR #368, opened 2018-09-14
fix: allow nginx to start even if an upstream container is down

shawnsarwar:
In the past any single container that nginx calls as upstream being missing at runtime would cause nginx not to start, which would take down kernel.aether.local and the kernel UI @ localhost:8000 wouldn't be useable since the CSFR cert is for the original url... and so on. 

This patch addresses that by allowing nginx to start with any or all upstream containers being down. It uses the docker resolver at 127.0.0.11 and somehow setting proxy_pass to a variable allows nginx to spawn. The only downside of the patch is if for example sync is down, it will return an ERR_NAME_NOT_RESOLVED instead of a 502, but this is still progress.


admbtlr (approved):
Nice!

PR #373, opened 2018-09-24
feat (ui): pipeline fetch/publish using mapping set

lordmallam:
Fetch kernel artefacts and stores them as `pipeline -> contracts` grouped by mapping sets
Allows pipeline create, edit and publish from UI to kernel using mapping sets
Revalidate grouped contracts on pipeline update
Refactor identity mapping to use mapping sets as contracts
Enforce read only pipelines (on input(pipeline) and contract levels individually)

Related JIRA Tickets:
[AET-346](https://jira.ehealthafrica.org/browse/AET-346)
[AET-342](https://jira.ehealthafrica.org/browse/AET-342)
[AET-277](https://jira.ehealthafrica.org/browse/AET-277)
[AET-347](https://jira.ehealthafrica.org/browse/AET-347)
[AET-343](https://jira.ehealthafrica.org/browse/AET-343)


shawnsarwar (commented):
So far so good. I'm going to take a look at the api spec generated from the changes, and make sure I can interface with the mapping_sets before I hit approve but I can't see any issues.

aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -134,10 +136,16 @@ def upsert_project_artefacts(
...
+        mapping_definition = raw_mapping.get('definition')
+        if mapping_definition is None:
             # in case of no mapping rules were indicated, do not update them.
             ignore_fields.append('definition')
-
+            mapping_definition = {'mappings': [], 'entities': {}}
+        else:
+            mapping_definition = {
+                'mapping': mapping_definition.get('mapping', []),
+                'entities': mapping_project_schemas,
```
obdulia-losantos: I found a future and very realistic problem here.

Example:

1. Pushing "Pipeline 1" with Entity Type "Person", the method creates the Schema "Person" with id 1.

2. Later, pushing another "Pipeline 2" with Entity Type "Person", the method create the Schema "Person_1" with id 2. In your mapping rules you have this:
```json
{
  "mapping": [
    ["zzz", "Person.id"],
  ],
  "entities": {
    "Person_1": 2,
  }
}
```

aether-ui/aether/ui/api/tests/test_views.py
```diff
@@ -170,65 +182,75 @@ def test_view_pipeline___publish(self):
...
 
     def test_view_pipeline__publish(self):
         pipeline = Pipeline.objects.create(
             name='Pipeline Mock 2',
             schema={'name': 'test', 'type': 'record', 'fields':
                     [{'name': 'name', 'type': 'string'}]},
             input={'test': {'name': 'myValue'}},
+        )
+        contract = Contract.objects.create(
             entity_types=[{'name': 'Test', 'type': 'record', 'fields':
```
obdulia-losantos: Styling question, what do you find more readable?

1.
```
[{'name': 'Test', 'type': 'record', 'fields':
 [{'type': 'string', 'name': 'name'}, {'type': 'string', 'name': 'id'}]}],
```

2.
```
[{
  'name': 'Test', 
  'type': 'record', 
  'fields': [{'type': 'string', 'name': 'name'}, {'type': 'string', 'name': 'id'}]
}],
```

3.
```
[{
  'name': 'Test', 
  'type': 'record', 
  'fields': [
    {'type': 'string', 'name': 'name'},
    {'type': 'string', 'name': 'id'}
  ]
}],
```

lordmallam: 🤣

aether-kernel/aether/kernel/api/project_artefacts.py
lordmallam: Valid point. I am not sure if schema auto generated names should be a thing, since mapping rules depend on them. Unless the rules are to be programmatically changed on the fly. However, on the UI side of things this issue is taken care of by a preflight function which warns of existing objects with the same name and gives the user the options of updating the existing schema or change the name  for another schema to be created which in turn forces the user to modify the mapping rules.

obdulia-losantos: Wondering if this small change in the line `115` would make the trick

```python
# take the original given name to be used as entity entry in the related mapping instances
mapping_project_schemas[raw_schema.get('name', schema.name)] = str(project_schema.pk)
```
or maybe this one:
```python
# take the original given name to be used as entity entry in the related mapping instances
mapping_project_schemas[raw_schema.get('name', project_schema.name)] = str(project_schema.pk)
```

obdulia-losantos: Another point is that ALL the schemas/project schemas are being assigned to ALL mappings... is there a simple way to detect only the used ones?

aether-ui/aether/ui/api/serializers.py
```diff
@@ -47,6 +47,24 @@ class PipelineSerializer(DynamicFieldsMixin, serializers.ModelSerializer):
 
     contracts = ContractSerializer(many=True, read_only=True)
 
+    def update(self, instance, validated_data):
+        read_only_contracts = instance.contracts.filter(is_read_only=True)
+        if read_only_contracts:
+            raise serializers.ValidationError({
+                'description': 'Input is readonly'
```
obdulia-losantos: As an enhancement, could you start using the translation library for the messages in the python side?

```python

                # ...
                'description': _('Input is readonly')
```

aether-ui/aether/ui/api/utils.py
```diff
@@ -216,167 +188,130 @@ def is_object_linked(kernel_refs, object_name, entity_type_name=''):
...
-    pipeline.kernel_refs = pipeline.kernel_refs if pipeline.kernel_refs else {}
-    project_data = {
-        'revision': str(uuid.uuid4()),
-        'name': projectname,
-        'salad_schema': '[]',
-        'jsonld_context': '[]',
-        'rdf_definition': '[]'
-    }
+    project_id = str(uuid.uuid4())
+    project_name = 'Aux'
```
obdulia-losantos: Why is the project name hard coded?

aether-ui/aether/ui/api/utils.py
```diff
@@ -216,167 +188,130 @@ def is_object_linked(kernel_refs, object_name, entity_type_name=''):
...
-
-    for entity_type in pipeline.entity_types:
-        schema_name = entity_type['name']
-        schema_data = {
-            'revision': str(uuid.uuid4()),
-            'name': schema_name,
-            'type': entity_type['type'],
-            'definition': entity_type
-        }
+        projects = kernel_data_request(f'projects/?name=Aux')['results']
```
obdulia-losantos: If there are already `kernel_refs` values, why aren't you using them instead of assigning new ones?
I mean here and in the code below.

aether-ui/aether/ui/api/utils.py
```diff
@@ -216,167 +188,130 @@ def is_object_linked(kernel_refs, object_name, entity_type_name=''):
...
-            'entities': pipeline.kernel_refs.get('projectSchema', {}),
-            'mapping': mapping
-            },
-        'revision': str(uuid.uuid4()),
-        'project': pipeline.kernel_refs['project']
-    }
+            'mapping': [
+                [rule['source'], rule['destination']]
+                for rule in contract.mapping
+            ]
```
obdulia-losantos, line 330: if you are assigning ids for the mapping sets, why don't you do the same for the schema/project schema and set the `entities` property here? 

Wouldn't that simplify the whole process?

aether-ui/aether/ui/api/views.py
```diff
@@ -52,41 +52,51 @@ def fetch(self, request):
     @action(methods=['post'], detail=True)
     def publish(self, request, pk=None):
         '''
-        This view transform the supplied pipeline to kernal models,
+        This view transforms the supplied pipeline to kernal models,
         publish and update the pipeline with related kernel model ids.
         '''
         project_name = request.data.get('project_name', 'Aux')
```
obdulia-losantos, line 54: Did I mention before that I hate the hard coded `Aux` everywhere? :roll_eyes:

aether-ui/aether/ui/api/utils.py
lordmallam: Mapping set ids are set here because the kernel artefact generation function does not know how to link mappings to specific mapping set automatically because it accepts a list of mapping sets

aether-ui/aether/ui/api/utils.py
lordmallam: Because all pipelines not previously linked to a project have to be identified as coming from AUX, thus an aux project on kernel. And there is no way to specify a project from the ui

obdulia-losantos: > And there is no way to specify a project from the ui

But you agree with me that there should be one (not in this PR of course), right?

obdulia-losantos: But my doubt goes more to the comment below. I'll try to put all together.

You include the `projectname` as an argument
```python
def publish_pipeline(pipeline, projectname, contract, objects_to_overwrite={}):
```
The next thing you do is to overwrite that value
```python
    project_name = 'Aux'
```
Wait, it's another variable that looks similar :scream: `projectname` is not used at all :woman_shrugging: 

Now with the project id...

Firstly, you set a new value instead of checking `pipeline.kernel_refs['project']` value
```python
    project_id = str(uuid.uuid4())
```
Next, you don't try to find that existing project, you go directly to search the "Aux" project
```python
    projects = kernel_data_request(f'projects/?name=Aux')['results']
```

Do you see my confusion when I review this?

lordmallam: > > And there is no way to specify a project from the ui
> 
> But you agree with me that there should be one (not in this PR of course), right?

Yes I do, but it's more of a UX problem than it is code.

aether-ui/aether/ui/api/utils.py
obdulia-losantos: And why don't you do it here to avoid further problems in the other side (AKA Kernel).

obdulia-losantos: This is the time to redesign the data model to make this as simpler as possible. Really, take your time and if you feel that you need to break something PLEASE GO AHEAD!

Something in the air tells me that the Kernel mapping id and the UI contract id should be exactly the same. I also feel that the UI entities should have their own model and not be a JSON Field in the Contract model (and yes, this implies to change the react part too). With those changes you can achieve pretty easily a one to one relationship between the UI instances and their respective Kernel instances.

aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -134,10 +136,24 @@ def upsert_project_artefacts(
...
         ignore_fields = ['name']
-        if raw_mapping.get('definition') is None:
+        mapping_definition = raw_mapping.get('definition')
+        if mapping_definition is None:
             # in case of no mapping rules were indicated, do not update them.
             ignore_fields.append('definition')
-
+            mapping_definition = {'mappings': [], 'entities': {}}
+        else:
+            used_schemas = []
```
obdulia-losantos, line 145: :open_mouth:

aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -134,10 +136,24 @@ def upsert_project_artefacts(
...
     for raw_mapping in mappings:
         ignore_fields = ['name']
-        if raw_mapping.get('definition') is None:
+        mapping_definition = raw_mapping.get('definition')
+        if mapping_definition is None:
             # in case of no mapping rules were indicated, do not update them.
             ignore_fields.append('definition')
-
+            mapping_definition = {'mappings': [], 'entities': {}}
+        else:
```
obdulia-losantos: This `else` is making CoudhDB-Sync tests fail, could you calculate the `entities` property only if it doesn't come within the mapping definition?

obdulia-losantos: `else if 'entities' not in mapping_definition:`

obdulia-losantos: Take in mind that ODK and CouchDB-Sync provide the whole definition, rules and entities.

aether-ui/aether/ui/api/utils.py
obdulia-losantos: This is what I mean

```text

    +------------------+       +------------------+       +------------------+
    | Pipeline         |       | Contract         |       | Entity           |
    +==================+       +==================+       +==================+
    | id               |<--+   | id               |<--+   | id               |
    | name             |   |   | name             |   |   | name             |
    | schema           |   |   | rules            |   |   | definition       |
    | input            |   |   | ...              |   |   | ...              |
    | mappingset_id    |   |   | mapping_id       |   |   | schema_id/ps_id  |
    +------------------+   |   +::::::::::::::::::+   |   +::::::::::::::::::+
                           +--<| pipeline         |   +--<| contract         |
                               +------------------+       +------------------+
```

You don't need the `kernel_refs` field anymore because those values are saved in each object.

What do you think?

lordmallam: This will help especially for querying, but I am of the option that since we now have a kernel based `project/artefacts` generation function, ui configuration of these artefacts should be reduced to the barest minimum, while the artefacts generator be made more flexible and robust. So some other 3rd part apps and consume it with minimum configurations

aether-ui/aether/ui/api/utils.py
```diff
@@ -171,212 +169,147 @@ def kernel_data_request(url='', method='get', data=None):
...
-def publish_preflight(pipeline, project_name, outcome):
+def publish_preflight(pipeline, outcome, contract):
     '''
     Performs a check for possible pipeline publish errors against kernel
     '''
-    if pipeline.mapping_errors:
-        outcome['error'].append('Mappings have errors')
+    if contract.mapping_errors:
+        outcome['error'].append('{} mappings have errors'.format(contract.name))
         return outcome
```
obdulia-losantos, line 207: Even with mapping errors, maybe it's a good option to continue checking the rest of artefacts and return all the problems together :thinking: 
_just thinking_

aether-ui/aether/ui/api/utils.py
```diff
@@ -171,212 +169,147 @@ def kernel_data_request(url='', method='get', data=None):
...
-
-def publish_preflight(pipeline, project_name, outcome):
+def publish_preflight(pipeline, outcome, contract):
     '''
     Performs a check for possible pipeline publish errors against kernel
     '''
-    if pipeline.mapping_errors:
-        outcome['error'].append('Mappings have errors')
+    if contract.mapping_errors:
+        outcome['error'].append('{} mappings have errors'.format(contract.name))
```
obdulia-losantos: Same comment as in the `serializers.py` file... Parlez vous français? Anglais? Espagnol? :woman_shrugging:

aether-ui/aether/ui/api/utils.py
```diff
@@ -171,212 +169,147 @@ def kernel_data_request(url='', method='get', data=None):
...
         return outcome
-    for entity_type in pipeline.entity_types:
-        if is_object_linked(pipeline.kernel_refs, 'schema', entity_type['name']):
+    for entity_type in contract.entity_types:
+        if is_object_linked(contract.kernel_refs, 'schemas', entity_type['name']):
             outcome['exists'].append({entity_type['name']: '{} schema with id {} exists'.format(
                                         entity_type['name'],
-                                        pipeline.kernel_refs['schema'][entity_type['name']])})
+                                        contract.kernel_refs['schemas'][entity_type['name']])})
```
obdulia-losantos: another styling comment... lines with several indentations are hard to follow.

```python
outcome['exists'].append({
    entity_type['name']: '{} schema with id {} exists'.format(
        entity_type['name'],
        contract.kernel_refs['schemas'][entity_type['name']]),
})
```

As an example... obviously there are more and better options

aether-ui/aether/ui/api/utils.py
```diff
@@ -171,212 +169,147 @@ def kernel_data_request(url='', method='get', data=None):
...
-                            'project': pipeline.kernel_refs['project'],
-                            'schema': schema_id
-                        }
-        create_new_kernel_object('projectSchema',
-                                 pipeline, project_schema_data, entity_name=entity_name)
-    return pipeline
-
-
-def publish_preflight(pipeline, project_name, outcome):
+def publish_preflight(pipeline, outcome, contract):
```
obdulia-losantos: This is checking the pipeline or the contract? Because if it's the contract the pipeline is accessible via `contract.pipeline` and you can remove the "duplicated" argument.

Also the outcome could be declared inside and not include as an argument (it's the returned value)

aether-ui/aether/ui/api/utils.py
```diff
@@ -171,212 +169,147 @@ def kernel_data_request(url='', method='get', data=None):
...
+            mappingset = kernel_data_request(f'mappingsets/{pipeline.mappingset}/')
+            if json.dumps(mappingset['input'], sort_keys=True) != json.dumps(pipeline.schema, sort_keys=True):
+                outcome['exists'].append({pipeline.name: 'Input data will be changed'})
+        except Exception:
+            pass
     return outcome
 
 
-def publish_pipeline(pipeline, projectname, overwrite=False):
+def publish_pipeline(pipeline, project_name, contract, objects_to_overwrite={}):
```
obdulia-losantos: Same comment, publishing pipeline and all its contracts or only one contract?

aether-ui/aether/ui/api/utils.py
```diff
@@ -171,212 +169,147 @@ def kernel_data_request(url='', method='get', data=None):
...
+        project_name = aux_project.get('name', 'Aux')
+    except Exception as e:
+        pass
+    if contract.kernel_refs and 'project' in contract.kernel_refs:
+        try:
+            project = kernel_data_request(f'projects/{contract.kernel_refs["project"]}/')
+            project_id = project.get('id')
+            project_name = project.get('name')
+        except Exception:
+            pass
```
obdulia-losantos, line 293: I think that the project id should be extracted from the pipeline mapping set and not kept within the contract...

```python
    # check linked Project I (using pipeline mapping set)
    project_id = None
    mappingset_id = None
    
    if pipeline.mappingset:
        mappingset_id = pipeline.mappingset
        try:
            kernel_mappingset = kernel_data_request(f'mappingsets/{pipeline.mappingset}/')
            project_id = mappingset.get('project')
            # take from kernel_mappingset the data to be push back to kernel
        except Exception:
            # take from pipeline the data to be pushed to kernel
            pass
    else:
        mappingset_id = str(uuid.uuid4())
        # take from pipeline the data to be pushed to kernel

    # check linked Project II (using given project name)
    if not project_id and project_name:
        try:
            projects = kernel_data_request(f'projects/?name={project_name}')['results']
            aux_project = projects[0] if len(projects) else {}
            project_id = aux_project.get('id')
        except Exception as e:
            pass

    # check linked Project III (if everything else failed...)
    if not project_id:
        project_id = str(uuid.uuid4())
```

and continue with contract...

aether-ui/aether/ui/api/views.py
```diff
@@ -41,52 +41,59 @@ def fetch(self, request):
...
         This view gets kernel objects, transforms and loads into a pipeline
         '''
-        try:
-            ui_utils.kernel_to_pipeline()
-        except Exception:
-            pass
+        ui_utils.kernel_to_pipeline()
         pipelines = models.Pipeline.objects.all()
         serialized_data = serializers.PipelineSerializer(pipelines, context={'request': request}, many=True).data
         return Response(serialized_data, status=HTTPStatus.OK)
```
obdulia-losantos: You can skip these lines just calling the the method `list`
`return self.list(request)`

aether-ui/aether/ui/api/utils.py
```diff
@@ -442,15 +376,9 @@ def linked_pipeline_object(object_name, id):
...
-    while True:
-        mappingsets = mappingsets_response['results']
-        for mappingset in mappingsets:
-            pipelines.append(create_new_pipeline_from_kernel(mappingset))
-        if mappingsets_response['next']:
-            url_segments = urlparse(mappingsets_response['next'])
-            mappingsets_response = kernel_data_request('{}?{}'.format(url_segments.path, url_segments.query)[1:])
-        else:
-            break
+    for mappingset in mappingsets:
```
obdulia-losantos, line 422: If you are not using the returned pipelines, you can save memory and resources not creating the array.

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -90,6 +78,9 @@ class Pipeline extends Component {
             </Link>
             <span> // </span>
             { selectedPipeline.name }
+            { selectedPipeline.is_read_only &&
+              <span className='tag'>read-only</span>
```
obdulia-losantos: I feel like I'm the only one caring about translations... :roll_eyes:

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -110,6 +113,9 @@ class PipelineList extends Component {
         </Modal>
         <div
           onClick={() => { this.onSelectPipeline(pipeline) }}>
+          { pipeline.is_read_only &&
+            <span className='tag'>read-only</span>
```
obdulia-losantos: :de: :gb: :fr: :es:

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -157,14 +165,15 @@ class PipelineList extends Component {
...
 const mapStateToProps = ({ pipelines }) => ({
   pipelineList: pipelines.pipelineList,
   error: pipelines.error,
-  isNewPipeline: pipelines.isNewPipeline
+  isNewPipeline: pipelines.isNewPipeline,
+  selectedPipeline: pipelines.selectedPipeline
 })
 
-export default connect(mapStateToProps, { getPipelines, selectedPipelineChanged, addPipeline, fetchPipelines, getKernelURL })(PipelineList)
+export default connect(mapStateToProps, { getPipelines, selectedPipelineChanged, addPipeline, fetchPipelines, getKernelURL, addInitialContract })(PipelineList)
```
obdulia-losantos: Do we try to achieve a 1000 chars line? :anguished:

docker-compose.yml
```diff
@@ -88,7 +88,7 @@ services:
       file: ./docker-compose-base.yml
       service: ui-base
     depends_on:
-      db:
+      kernel:
```
obdulia-losantos: Merge conflict?

aether-ui/aether/ui/assets/css/base/_global.scss
```diff
@@ -209,6 +227,25 @@ select:-webkit-autofill {
...
+  .pipeline-section__entityTypes .textarea-header {
+    box-shadow: none;
+  }
+  .tabs .tab.selected {
+    box-shadow: none;
+  }
+
+  .pipeline-section__mapping .rule-input {
+    box-shadow: none;
+    &.source:before,
```
obdulia-losantos: I see sass-lint complaining...
```text
> sass-lint --verbose

./css/_pipelines.scss
  100:1  warning  Space expected between blocks  empty-line-between-blocks

./css/base/_global.scss
  236:1   warning  Space expected between blocks                  empty-line-between-blocks
  242:1   warning  Space expected between blocks                  empty-line-between-blocks
  242:13  warning  Pseudo-elements must start with double colons  pseudo-element
  243:18  warning  Pseudo-elements must start with double colons  pseudo-element

✖ 5 problems (0 errors, 5 warnings)
```

aether-ui/aether/ui/api/utils.py
lordmallam: Mapping errors are calculated during the pipeline edit, and are blockers for any publish. other checks require calls be made to the kernel and after a publish is initiated. i don't think the process should continue if there are mapping errors.

obdulia-losantos: My point is that the `input` only represents a possible scenario but not all of them and the mapping rules reflect one of them but not this one.

AVRO Schema
```
{
    "name": "Person",
    "type": "record",
    "fields": [
        {
            "name": "first_name",
            "type": "string",
        },
        {
            "name": "middle_name",
            "type": ["null", "string"],
        },
        {
            "name": "last_name",
            "type": "string",
        }
    ]
}
```
Input sample
```
{
  "name": "peter",
  "last_name": "smith"
}
```
Mapping rules:
```
['$.first_name', 'Person.first_name'],
['$.middle_name', 'Person.middle_name'],
['$.last_name', 'Person.last_name']
```

obdulia-losantos: But hey, it's just a comment. No action requested here.

aether-ui/aether/ui/api/utils.py
lordmallam: The algorithm here is to retrieve the linked project using `pipeline.mappingset` or `contract.kernel_refs['project']` if previously publish, or create and link an `Aux` project to  all new pipelines coming from the ui

docker-compose.yml
lordmallam: Should the ui depend on kernel which in turn depends on db or just db?

aether-ui/aether/ui/api/views.py
lordmallam: I'll stick with this implementation. Don't want to deal with pagination

obdulia-losantos: You can easily skip pagination with this option :wink: 

```python
    pagination_class = None
```

The PRO of using `return self.list(request)` is that if you have searches/filters/permissions/ordering/... you don't need to implement them twice.

As an example `/pipelines/fetch/?fields=id` should return only the list of ids, with your manual approach it doesn't. :woman_shrugging:

docker-compose.yml
obdulia-losantos: To avoid "healthy" problems we removed the containers dependencies, only the databases are included, but if you prefer to have it, the most important dependency is the database :wink:

```yaml
  ui:
    extends:
      file: ./docker-compose-base.yml
      service: ui-base
    depends_on:
      db:
        condition: service_healthy
      kernel:
        condition: service_started
      # use this dependency with HRM, otherwise comment it out
      # ui-assets:
      #   condition: service_started
    networks:
      internal:
        aliases:
          - ui.aether.local
```

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -110,6 +113,14 @@ class PipelineList extends Component {
         </Modal>
         <div
           onClick={() => { this.onSelectPipeline(pipeline) }}>
+          { pipeline.is_read_only &&
+            <span className='tag'>
+              <FormattedMessage
+                id='pipeline.navbar.input'
```
obdulia-losantos: could you use an id that indicates that the text is "read only"? ;-)

aether-ui/aether/ui/api/tests/test_views.py
```diff
@@ -123,117 +167,251 @@ def test_view_pipeline___publish(self):
...
         url = reverse('pipeline-fetch')
         response = self.client.post(url, content_type='application/json')
         response_data = json.loads(response.content)
-        self.assertEqual(len(response_data), 2)
-        self.assertEqual(len(response_data[0]['entity_types']), 2)
+        self.assertEqual(len(response_data), 4)
+        self.assertEqual(len(response_data[1]['contracts'][0]['entity_types']), 2)
 
-        # Ensure linked mappings are not recreated
+        # Ensure linked mappings are not recreated]
```
obdulia-losantos: :grimacing:

obdulia-losantos:
Could you merge current `parallel-extraction` into this branch? All tests may pass now :crossed_fingers:

aether-ui/aether/ui/api/utils.py
obdulia-losantos: Any action here?

aether-ui/aether/ui/api/views.py
```diff
@@ -35,58 +35,57 @@ class PipelineViewSet(viewsets.ModelViewSet):
...
+            contract = get_object_or_404(models.Contract, pk=contract_id)
         except Exception as e:
+            outcome = {'error': []}
             outcome['error'].append(str(e))
             return Response(outcome, status=HTTPStatus.BAD_REQUEST)
-        outcome = ui_utils.publish_preflight(pipeline, project_name, outcome)
+        outcome = ui_utils.publish_preflight(contract)
 
-        if outcome['error']:
+        if 'error' in outcome and len(outcome['error']):
```
obdulia-losantos, line 69: Simpler: `if outcome.get('error'):` :wink:

obdulia-losantos (approved):
I still think that there are a couple of things to be done/rethought... but so far it's great.
:clap:

PR #384, opened 2018-10-03
refactor: cleaning model definition

obdulia-losantos:
Cleaning:
- Removed useless `null=False` and `blank=False`
- Included `blank=True` when needed.

Following Django suggestions

- Replaced `default=[]` with `default=list`
- Replaced `default={}` with `default=dict`


lordmallam (approved):
LGTM

PR #389, opened 2018-10-05
feat(kernel): create an empty mapping along with the passthrough one

obdulia-losantos:


aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -215,11 +215,21 @@ def upsert_project_with_avro_schemas(
...
     mappings = []
 
-    for raw_schema in avro_schemas:
-        schema, mapping = parser(raw_schema.get('id'), raw_schema.get('definition'))
+    for avro_schema in avro_schemas:
+        schema, mapping = parser(avro_schema.get('id'), avro_schema.get('definition'))
         schemas.append(schema)
         mappings.append(mapping)
 
+        # Include an empty mapping if the schema is new
```
lordmallam, line 223: Shouldn't this logic be on the 3rd party app instead of `kernel`, so as to define `n` number of extra mappings to include or not

obdulia-losantos: No, the third party doesn't know if the schema was already created or not and also this is a temporal solution, as far as I understood, because the UI cannot add new "contracts" to an existing pipeline yet. Am I wrong?

admbtlr: I think you're probably both right. Yes, if this were intended to be permanent functionality, it should be handled by the 3rd party module, but actually it's just a temporary workaround, so I think it's fine to implement it here.

lordmallam (approved):
Looks good as a temporary solution

admbtlr:
Do we have a Jira ticket for the proper solution?

obdulia-losantos:
We can add a ticket in Jira that indicates the number of empty mappings that can be created along with the AVRO schema :thinking:

admbtlr:
Actually we have https://jira.ehealthafrica.org/browse/AET-352 - once that's done we can remove this temporary workaround

admbtlr (approved):

PR #396, opened 2018-10-08
fix: remove pip freeze from integration tests

shawnsarwar:
Removes pip_freeze from integration tests. requirements.txt is now generated as part of the docker build process for the integration tests container.


obdulia-losantos (approved):
This works for the integration tests but what about the producer?

shawnsarwar:
@obdulia-losantos The producer works on the normal pip_freeze mechanism. It will always be shipped with the latest client version because of how the build/ deploy process works. If we don't specify a version of the client or do >= in the requirements, it'll use the latest because that's what's been included in its /dependencies folder. And that should work because to deploy those two would have had to pass integration tests together.

PR #405, opened 2018-10-11
fix(kernel): set input to avro schema

lordmallam:
Sets avro schema to mappingset input field not input json data


aether-kernel/aether/kernel/api/avro_tools.py
```diff
@@ -469,7 +468,7 @@ def avro_schema_to_passthrough_artefacts(item_id, avro_schema):
         # this is an auto-generated mapping that shouldn't be modified manually
         'is_read_only': True,
         'is_active': True,
-        'input': sample,  # include a data sample
+        'input': definition,  # include a data sample
```
obdulia-losantos: I thought that the input was the data sample, not the AVRO schema itself... was I wrong?

lordmallam: Was to be the AVRO schema 😬

aether-kernel/aether/kernel/api/avro_tools.py
```diff
@@ -469,7 +468,7 @@ def avro_schema_to_passthrough_artefacts(item_id, avro_schema):
         # this is an auto-generated mapping that shouldn't be modified manually
         'is_read_only': True,
         'is_active': True,
-        'input': sample,  # include a data sample
+        'input': definition,  # include avro schema
```
obdulia-losantos: If this is the schema, for coherence we should rename it to schema, don't you think the same?

https://github.com/eHealthAfrica/aether/blob/develop/aether-ui/aether/ui/api/models.py#L32-L36

```python
class Pipeline(ExportModelOperationsMixin('ui_pipeline'), TimeStampedModel):
    # ...

    # this is the avro schema
    schema = JSONField(null=True, blank=True, default=dict)

    # this is an example of the data using the avro schema
    input = JSONField(null=True, blank=True, default=dict)
```

or have both :woman_shrugging:

admbtlr: It's slightly annoying that you don't get randomly generated data with a schema that is shared with a read-only pipeline - would it be possible to have both?

obdulia-losantos: something like this?

```python
def avro_schema_to_passthrough_artefacts(item_id, avro_schema):
    # ...

    schema = {
        # ...
    }
    mappingset = {
        'id': item_id,
        'name': name,
        'schema': avro_schema,  # the original AVRO schema (without the missing id)
        'input': random_avro(avro_schema),
    }
    mapping = {
        'id': item_id,
        'name': name,
        'definition': {
            'entities': {name: item_id},
            'mapping': rules,
        },
        # this is an auto-generated mapping that shouldn't be modified manually
        'is_read_only': True,
        'is_active': True,
    }

    return schema, mappingset, mapping
```

admbtlr: Yes, I guess so - it hadn't occurred to me that the generated data belongs to the `mappingset`, but that would actually be totally fine

lordmallam:
Should have both now.

aether-kernel/aether/kernel/api/avro_tools.py
```diff
@@ -468,7 +469,8 @@ def avro_schema_to_passthrough_artefacts(item_id, avro_schema):
         # this is an auto-generated mapping that shouldn't be modified manually
         'is_read_only': True,
         'is_active': True,
-        'input': definition,  # include avro schema
+        'input': sample,  # include a data sample
+        'schema': definition,  # include avro schema
```
obdulia-losantos: This should be the original AVRO schema, `definition` is the tweaked one with the missing id :grimacing:

```python
'input': random_avro(avro_schema),  # include a data sample
'schema': avro_schema,  # include avro schema
```

But not sure if I prefer to return the mappingset instance rather than include the fields withing the mapping definition :thinking:

obdulia-losantos: I mean this

```python
def upsert_project_with_avro_schemas(
    action='upsert',
    project_id=None,
    project_name=None,
    avro_schemas=[],
):
    '''
    Creates or updates the project and links it with the given AVRO schemas
    and related artefacts: project schemas and passthrough mappings.

    Returns the list of project and its affected artefact ids by type.
    '''

    schemas = []
    mappingsets = []
    mappings = []

    for avro_schema in avro_schemas:
        schema, mapping = parser(avro_schema.get('id'), avro_schema.get('definition'))
        schemas.append(schema)
        mappings.append(mapping)

        id = schema.get('id')
        mappingsets.append({
            'id': id,
            'name': schema.get('name'),
            'schema': avro_schema.get('definition'),
            'input': random_avro(avro_schema.get('definition')),
        })

        # Include an empty mapping if the schema is new
        if not Schema.objects.filter(pk=id).exists():
            mappings.append({
                # even being the same name it's going to append the suffix `-1`
                'name': schema.get('name'),
                # the passthrough mapping has created a mappingset with this id
                'mappingset': id,
            })

    return upsert_project_artefacts(
        action=action,
        project_id=project_id,
        project_name=project_name,
        schemas=schemas,
        mappingsets=mappingsets,
        mappings=mappings,
    )
```

lordmallam: Understood, the cool thing is `upsert_project_artefacts` knows how to handle both

obdulia-losantos: :stuck_out_tongue:

admbtlr:
So is this ready now?

obdulia-losantos:
Not yet, this is still a thing

```python
'input': random_avro(avro_schema),  # include a data sample
'schema': avro_schema,  # include avro schema
```

I don't mind if this is the fix or the another one with the mappingset, but one of them.

obdulia-losantos (changes requested):
the mappingset schema should be the original one `avro_schema` not the tweaked one.

aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -178,6 +183,7 @@ def upsert_project_artefacts(
                 action=action,
                 name=mapping_name,  # use same name as mapping
                 input=raw_mapping.get('input', {}),
+                schema=raw_mapping.get('schema', {}),
```
obdulia-losantos, line 186: I see an enhancement here, but I would leave it for a future task and it's to derive the input from the schema and the other way around if one of them is missing :thinking:

obdulia-losantos (dismissed):

obdulia-losantos (approved):

PR #412, opened 2018-10-12
feat: filter submissions by payload content

quesebifurcan:
Add a dynamic filter to SubmissionViewSet and EntityViewSet. 

Any query parameter prefixed by "payload__" will filter submissions
based on the contents of e.g. Submission.payload.

Example:
The URL "/submissions?payload__a__b__c=1" will yield
the queryset

  models.Submission.objects.filter(payload__a__b__c=1)

and return a list of submissions with a JSON payload like

  '{"a": {"b": {"c": 1}}}'

Note that it is possible to compare not just strings, but
numbers, lists and objects as well.

### DIFF
--- a/aether-kernel/aether/kernel/api/views.py
+++ b/aether-kernel/aether/kernel/api/views.py
@@ -261,6 +261,37 @@ class ProjectViewSet(CustomViewSet):
         return Response(data=results)
 
 
+class PayloadFilterMixin(object):
+    def get_queryset(self):
+        '''
+        Adds a dynamic filter to a viewset. Any query
+        parameter prefixed by "payload__" will filter entries
+        based on the contents of <ViewSet>.payload.
+
+        Example:
+        The URL "/submissions?payload__a__b__c=1" will yield
+        the queryset
+
+            models.Submission.objects.filter(payload__a__b__c=1)
+
+        and return a list of submissions with a JSON payload like
+
+            '{"a": {"b": {"c": 1}}}'
+
+        Note that it is possible to compare not just strings, but
+        numbers, lists and objects as well.
+        '''
+        queryset = self.queryset
+        for k, v in self.request.query_params.items():
+            if k.startswith('payload__'):
+                try:
+                    kwargs = {k: json.loads(v)}
+                except json.decoder.JSONDecodeError:
+                    kwargs = {k: v}
+                queryset = queryset.filter(**kwargs)
+        return queryset
+
+
 class ProjectStatsViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = models.Project \
                      .objects \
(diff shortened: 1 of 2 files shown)


aether-kernel/aether/kernel/api/views.py
```diff
@@ -310,6 +312,16 @@ class SubmissionViewSet(CustomViewSet):
...
+    def get_queryset(self):
+        queryset = models.Submission.objects.all()
+        for k, v in self.request.query_params.items():
+            if k.startswith('payload__'):
+                try:
+                    kwargs = {k: json.loads(v)}
+                except json.decoder.JSONDecodeError:
+                    kwargs = {k: v}
+                queryset = queryset.filter(**kwargs)
+        return queryset
```
obdulia-losantos: Don't you prefer to include this in the `filters.SubmissionFilter` class as a method? :thinking:

aether-kernel/aether/kernel/api/views.py
```diff
@@ -310,6 +312,16 @@ class SubmissionViewSet(CustomViewSet):
     serializer_class = serializers.SubmissionSerializer
     filter_class = filters.SubmissionFilter
 
+    def get_queryset(self):
+        queryset = models.Submission.objects.all()
```
obdulia-losantos: This line will remove previous filters executed by the `SubmissionFilter` class like the `project` filter :scream: 
The correct line is `queryset = self.queryset` :stuck_out_tongue:

aether-kernel/aether/kernel/api/views.py
obdulia-losantos: Maybe it's too much complicated :woman_shrugging:

aether-kernel/aether/kernel/api/views.py
quesebifurcan: Oh thanks, I thought this ran before the other filters. I'll take a look.

aether-kernel/aether/kernel/api/views.py
quesebifurcan: As far as I can tell, FilterSets (from the django filters library) do not accept completely dynamic query parameters, e.g. `payload__x__y__z` (we need this, since the models know nothing about the structure of the JSON payload). That's why I put it in the views.

aether-kernel/aether/kernel/api/views.py
obdulia-losantos: Maybe you are right, but imagine that the default `queryset` is something like this:

```python
    queryset = Project.objects \
                      .prefetch_related('xforms', 'xforms__media_files') \
                      .order_by('name')
```
or
```python
    queryset = models.Project \
                     .objects \
                     .values('id', 'name', 'created') \
                     .annotate(
                         first_submission=Min('submissions__created'),
                         last_submission=Max('submissions__created'),
                         submissions_count=Count('submissions__id', distinct=True),
                         entities_count=Count('submissions__entities__id', distinct=True),
                     )
```

:woman_shrugging:

quesebifurcan: See docs here: https://www.django-rest-framework.org/api-guide/filtering/#overriding-the-initial-queryset

quesebifurcan: Yes, agreed it's much clearer to use `self.queryset`. Fixing.

obdulia-losantos (approved):

PR #287, opened 2018-07-13
chore: upgrade to Django@2.1

obdulia-losantos:
Once we realized that the problem with ODK Collect was with `uwsgi` and not `django@2` itself...

PR #416, opened 2018-10-16
fix(kernel): filter entity stats by schema family

obdulia-losantos:
What I forgot for Gather...

### DIFF
--- a/aether-kernel/aether/kernel/api/views.py
+++ b/aether-kernel/aether/kernel/api/views.py
@@ -436,6 +436,7 @@ class SubmissionStatsMixin(object):
                 distinct=True,
             )
             # Django 2: filter in Count
+            # (replace code above with this block when we upgrade to Django 2)
             # entities_count = Count(
             #     expression='submissions__entities__id',
             #     distinct=True,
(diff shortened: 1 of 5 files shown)


aether-kernel/aether/kernel/api/views.py
```diff
@@ -450,6 +416,57 @@ def get_entity_linked_data(entity, request, resolved, depth, start_depth=0):
...
+            entities_count = Count(
+                Case(
+                    When(
+                        submissions__entities__projectschema__schema__family=family,
+                        then='submissions__entities__id',
+                    ),
+                ),
+                distinct=True,
+            )
+            # Django 2: filter in Count
```
admbtlr, line 438: So is this a proposed alternative implementation? How can we choose between them?

obdulia-losantos: We need to migrate to Django 2 to have the second option.

https://github.com/eHealthAfrica/aether/pull/287

admbtlr: Ah, I didn't realise that `1` and `2` just referred to versions! Could you just put in something like:

```suggestion
            # Django 2: filter in Count
            # (replace code above with this block when we upgrade to Django 2)
```

admbtlr (approved):

PR #417, opened 2018-10-17
fix: jsonpath key wildcard handling

shawnsarwar:
Given data of structure:
```json
    {
        "households": [
            {
                "address": "74 Whyioughta St.",
                "name1": "Larry",
                "number1": 1,
                "name2": "Curly",
                "number2": 2
            },
            {
                "address": "1600 Ipoke Ave",
                "name1": "Moe",
                "number1": 3
            }
        ]
    }
```

Previously paths like the following were failing:
`$.households[*].name*`
This was for two reasons. The old regex was failing to identify this is a path with a wildcarded key : `.name*`. Fixing that, there remained the issue of handling indexed (`households[*]` or `households[1]`) paths within the filtering stage after the permissive JSONPath search is performed. This PR addresses both of those issues and brings the behavior back to what is expected.

### DIFF
--- a/aether-kernel/aether/kernel/api/utils.py
+++ b/aether-kernel/aether/kernel/api/utils.py
@@ -124,7 +124,7 @@ def json_printable(obj):
         return obj
 
 
-custom_jsonpath_wildcard_regex = re.compile('(\$\.)*([a-zA-Z0-9_-]+\.)*?[a-zA-Z0-9_-]+\*')
+custom_jsonpath_wildcard_regex = re.compile('(\$\.)+([a-zA-Z0-9_-]*(\[.*\])*\.)?[a-zA-Z0-9_-]+\*')
 incomplete_json_path_regex = re.compile('[a-zA-Z0-9_-]+\*')
 
 
--- a/aether-kernel/aether/kernel/api/utils.py
+++ b/aether-kernel/aether/kernel/api/utils.py
@@ -17,6 +17,7 @@
 # under the License.
 
 import collections
+import fnmatch
 import json
 import logging
 import re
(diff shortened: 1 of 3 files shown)


aether-kernel/aether/kernel/api/utils.py
```diff
@@ -55,14 +56,25 @@ class CachedParser(object):
     def parse(path):
         # we never need to call parse directly; use find()
         if path not in CachedParser.cache.keys():
-            CachedParser.cache[path] = jsonpath_ng_ext_parse(path)
+            try:
+                CachedParser.cache[path] = jsonpath_ng_ext_parse(path)
+            except Exception as err:  # jsonpath-ng raises the base exception type
+                new_err = 'exception parsing path %s : %s', (path, err)
```
obdulia-losantos: Could you user the `_` function for the messages?
```python
new_err = _('exception parsing path {path} : {error}').format(path=path, error=str(err))
```

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -55,14 +56,25 @@ class CachedParser(object):
...
 
     @staticmethod
     def find(path, obj):
         # find is an optimized call with a potentially cached parse object.
         parser = CachedParser.parse(path)
-        return parser.find(obj)
+        try:
+            return parser.find(obj)
+        except Exception as err:  # jsonpath-ng raises the base exception type
+            new_err = 'exception finding path %s : %s', (path, err)
```
obdulia-losantos: Same :gb: :de: :fr: comment

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -124,7 +136,7 @@ def json_printable(obj):
         return obj
 
 
-custom_jsonpath_wildcard_regex = re.compile('(\$\.)*([a-zA-Z0-9_-]+\.)*?[a-zA-Z0-9_-]+\*')
+custom_jsonpath_wildcard_regex = re.compile('(\$)?(\.)?([a-zA-Z0-9_-]*(\[.*\])*\.)?[a-zA-Z0-9_-]+\*')
```
obdulia-losantos, line 140: regular expressions are hard to read and understand, if you could write a line with an example of the expected string... I would love it! 
_simple suggestion, no action requested_

obdulia-losantos (changes requested):
If there hadn't been a Babel Tower...

aether-kernel/aether/kernel/api/utils.py
shawnsarwar: @obdulia-losantos  Good idea. See new changes for explanations.

obdulia-losantos: :heart_eyes:

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -69,12 +71,7 @@ def parse(path):
     def find(path, obj):
         # find is an optimized call with a potentially cached parse object.
         parser = CachedParser.parse(path)
-        try:
-            return parser.find(obj)
-        except Exception as err:  # jsonpath-ng raises the base exception type
-            new_err = 'exception finding path %s : %s', (path, err)
-            logger.error(new_err)
-            raise EntityValidationError(new_err) from err
+        return parser.find(obj)
```
obdulia-losantos: No try+catch here?

obdulia-losantos (approved):

PR #422, opened 2018-10-18
fix: better and faster passthrough items detection

obdulia-losantos:
As a quick workaround to detect passthrough items, just mark the schema family with the project id


quesebifurcan (changes requested):
Looks good. Do we have any documentation on what e.g. a "passthrough mapping" means?

obdulia-losantos:
Do you mean documentation about the also known as "identity mapping"?

aether-kernel/aether/kernel/api/views.py
```diff
@@ -65,15 +66,20 @@ def schemas_skeleton(self, request, pk=None, *args, **kwargs):
         docs = {}
         name = None
 
-        family = request.GET.get('family')
+        if 'passthrough' in request.query_params:
```
quesebifurcan: Would `/<entity>?passthrough=true` and `/<entity>?passthrough=false` lead to the same result (using project id as family)?

quesebifurcan (changes requested):

quesebifurcan:
@obdulia-losantos yes, that's what I meant. I forgot that the terminology changed. Now I know what it means, at least :smile:

aether-kernel/aether/kernel/api/views.py
obdulia-losantos: Being project id `abc`

These endpoints return the same result:
- `/projects/abc/schemas-skeleton/?passthrough=`
- `/projects/abc/schemas-skeleton/?passthrough=false`
- `/projects/abc/schemas-skeleton/?passthrough=true`
- `/projects/abc/schemas-skeleton/?family=abc`

Another thing is the entities endpoint

Having:
- Project `abc`
- Project `def`
- Schema `xyz` with family `abc` and with project schema linked to Project `abc` and Entity `E`.
- Schema `uvw` with family `abc` and with project schema linked to Project `def` and Entity `F`.

How the entities endpoint works:
- `/entities/?passthrough=` returns entity `E`
- `/entities/?family=abc` returns entity `E` and `F`

obdulia-losantos: I'm going to include one more test to make this more obvious :crossed_fingers:

quesebifurcan: Thanks for the explanation! I find it a bit confusing that `passthrough=false` also creates a passthrough pipeline. In other words, that the value of the query parameter doesn't matter. How about only creating a pipeline when `passthrough=true` is used? Are there any disadvantages with that approach?

quesebifurcan: In other words:
```
if json.loads(self.request.query_params.get('passthrough', 'false')):
    # assignments here
```
(Or just a direct equality check against `true`)

quesebifurcan (approved):

PR #423, opened 2018-10-18
fix(kernel): passthrough filter and read only mappings

obdulia-losantos:
As @adamvert pointed here https://github.com/eHealthAfrica/gather/pull/43#issuecomment-431014686

Include at least one more restriction to the filter


admbtlr (approved):
🎉 🎉

PR #424, opened 2018-10-19
feat: project/schema CASCADE deletion

obdulia-losantos:
_cleaning the data_ :broken_heart: 

- delete a project/schema implies that **ALL** associated data (including entities) will be deleted too; `CASCADE` option in the foreign key definition.

- delete a project will also delete the associated passthrough schemas if they are used only by the project, otherwise will remove the passthrough flag (`family` content)

- delete a project will not delete the rest of associated schemas, even if they are used only by the project.


admbtlr (dismissed):

obdulia-losantos:
@adamvert one more time after merging develop :grimacing:

quesebifurcan (approved):

PR #425, opened 2018-10-19
fix(odk): make AVRO fields nullable

obdulia-losantos:
AVRO schemas cannot handle all the possible workflows as the xForms. To avoid further validation errors in the data submissions we decided to make all the fields "nullable".

### DIFF
--- a/aether-odk-module/aether/odk/api/xform_utils.py
+++ b/aether-odk-module/aether/odk/api/xform_utils.py
@@ -323,22 +323,22 @@ def parse_xform_to_avro_schema(xml_definition, default_version=DEFAULT_XFORM_VER
                             {
                                 'name': 'latitude',
                                 'doc': _('latitude'),
-                                'type': __get_avro_primitive_type('float', True),
+                                'type': __get_avro_primitive_type('float', required=False),
                             },
                             {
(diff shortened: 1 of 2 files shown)


aether-odk-module/aether/odk/api/xform_utils.py
```diff
@@ -323,22 +323,22 @@ def parse_xform_to_avro_schema(xml_definition, default_version=DEFAULT_XFORM_VER
                             {
                                 'name': 'latitude',
                                 'doc': _('latitude'),
-                                'type': __get_avro_primitive_type('float', True),
+                                'type': __get_avro_primitive_type('float', required=False),
```
admbtlr, line 326: Could you add a comment saying something like "Since an Avro schema does not contain the same branching logic as an XForm, a field that is mandatory in a form is not actually always present, and therefore cannot be required in the schema"?

obdulia-losantos: Done!

obdulia-losantos: But a couple of lines below :grimacing:

admbtlr (approved):

PR #427, opened 2018-10-19
fix: retrieve POST data via request.data

quesebifurcan:
Prior to this commit, attempting to create/retrieve a user token via basic auth:
```
curl -H "Content-Type: application/json" -d '{"username": "<username>"}' -u <username:password> http://kernel.aether.local:8000/accounts/token
```
resulted in an exception being thrown in `aether.common.views.obtain_auth_token`.


obdulia-losantos (approved):
:scream:

PR #430, opened 2018-10-21
feat: travis auto release versioning

lordmallam:
NOTE: GitHub `personal access token` should be added to travis env as `GITHUB_TOKEN`

JIRA Ticket: [AET-320](https://jira.ehealthafrica.org/browse/AET-320)


scripts/release.sh
```diff
@@ -43,57 +42,165 @@ release_app () {
...
+        git fetch ${REMOTE} $TRAVIS_BRANCH
+        git branch $TRAVIS_BRANCH FETCH_HEAD
+        COMMIT_BRANCH=HEAD
+    fi
+
+    git checkout $TRAVIS_BRANCH
+    echo ${BRANCH_OR_TAG_VALUE} > VERSION
+    git add VERSION
+    # make Travis CI skip this build
+    git commit -m "Version updated to ${BRANCH_OR_TAG_VALUE} [ci skip]"
```
admbtlr: Please use a semantic message, e.g. `chore: update VERSION file to ${BRANCH_OR_TAG_VALUE}`.

Also, probably good to put this in a variable to reuse below

admbtlr: Please use a semantic message, e.g. `chore: update VERSION file to ${BRANCH_OR_TAG_VALUE}`.

Also, probably good to put this in a variable to reuse below

shawnsarwar:
@lordmallam What happens on RC builds?

shawnsarwar:
@lordmallam What happens on RC builds?

shawnsarwar:
@lordmallam What happens on RC builds?

shawnsarwar:
@lordmallam What happens on RC builds?

scripts/release.sh
```diff
@@ -43,57 +42,165 @@ release_app () {
...
+        git commit -m "Version updated to ${BRANCH_OR_TAG_VALUE} [ci skip]" #Skip travis build on develop commit
+        git push ${REMOTE} develop
+    else
+        echo "Develop branch VERSION value is not updated"
+        echo "New VERSION ${BRANCH_OR_TAG_VALUE} is same or less than develop VERSION ${DEV_VERSION}"
+    fi
+}
+
 # release version depending on TRAVIS_BRANCH/ TRAVIS_TAG
 if [[ $TRAVIS_TAG =~ ^[0-9]+\.[0-9]+[\.0-9]*$ ]]
```
obdulia-losantos: within this `if+elif+else` instruction you don't need any `exit` command :wink:

admbtlr (changes requested):

lordmallam:
@shawnsarwar on RC builds the `VERSION` is set to the branch version, e.g on branch `release-1.2`, VERSION is forced to be `1.2.0` and a docker image of `1.2.0-rc` is created. If the current `develop` VERSION is lower than `1.2.0` it is changed to match

admbtlr (dismissed):

quesebifurcan (dismissed):
Would be great to be able to test this somehow, is there any way we could run it on an example repo?

shawnsarwar:
@obdulia-losantos Has this functionality been handled in your more recent PR?

obdulia-losantos:
No, we are still pending on testing this. We suggested to copy this script in a new repo and play with PRs, tags, commits... before merging it

scripts/release.sh
```diff
@@ -43,57 +42,163 @@ release_app () {
...
+        git branch $TRAVIS_BRANCH FETCH_HEAD
+        COMMIT_BRANCH=HEAD
+    fi
+
+    git checkout $TRAVIS_BRANCH
+    echo ${BRANCH_OR_TAG_VALUE} > VERSION
+    git add VERSION
+    # make Travis CI skip this build
+    COMMIT_MESSAGE="chore: Update VERSION file to ${BRANCH_OR_TAG_VALUE} [ci skip]"
+    git commit -m COMMIT_MESSAGE
```
admbtlr:
```suggestion
    git commit -m "${COMMIT_MESSAGE}"
```

scripts/release.sh
```diff
@@ -43,55 +42,161 @@ release_app () {
...
+    fi
+
+    for (( p=`grep -o "\."<<<".$BRANCH_OR_TAG_VALUE"|wc -l`; p<3; p++)); do 
+        BRANCH_OR_TAG_VALUE+=.0;
+        done;
+    echo "Setting VERSION to " ${BRANCH_OR_TAG_VALUE}
+    
+    if [[ $3 = "tag" ]];
+    then
+        git fetch ${REMOTE} $TRAVIS_BRANCH
```
admbtlr, line 185:
```suggestion
        # https://github.com/travis-ci/travis-ci/issues/4745#issuecomment-345711626
        TRAVIS_BRANCH=$(git ls-remote origin | sed -n "\|$TRAVIS_COMMIT\s\+refs/heads/|{s///p}")
        git fetch ${REMOTE} $TRAVIS_BRANCH
```
On tagged builds, the value of `TRAVIS_BRANCH` is the tag name, not the branch name.

admbtlr:
Actually, I think there are deeper problems here. It seems that:

- if I tag a commit with e.g. `1.5.2`, the script sets the `VERSION` file to 1.5.2 - it should be 1.5.3
- if I then skip a release and tag the next commit with `1.5.4`, the `VERSION` file just gets incremented to 1.5.3 - it should actually be 1.5.5

You can test all this on [aether-release-test](https://github.com/eHealthAfrica/aether-release-test)

obdulia-losantos:
Once again, could you remove the `REVISION` and `VERSION` files? :stuck_out_tongue:

scripts/release.sh
```diff
@@ -59,43 +58,169 @@ function release_app {
...
+    local v=$1
+    if [ -z $2 ]; then 
+        local rgx='^((?:[0-9]+\.)*)([0-9]+)($)'
+    else 
+        local rgx='^((?:[0-9]+\.){'$(($2-1))'})([0-9]+)(\.|$)'
+        for (( p=`grep -o "\."<<<".$v"|wc -l`; p<$2; p++)); do 
+            v+=.0; done;
+    fi
+    val=`echo -e "$v" | perl -pe 's/^.*'$rgx'.*$/$2/'`
+    TAG_INCRESED_VERSION=$(echo "$v" | perl -pe s/$rgx.*$'/${1}'`printf %0${#val}s $(($val+1))`/)
```
obdulia-losantos: is this a typo? `TAG_INCREASED_VERSION` :wink:

admbtlr:
Looks good! Let me do some testing...

scripts/release.sh
```diff
@@ -18,7 +18,6 @@
 # specific language governing permissions and limitations
 # under the License.
 #
```
obdulia-losantos, line 20: Do you need to remove `set -Eeuo pipefail`?

scripts/release.sh
```diff
@@ -59,43 +58,169 @@ function release_app {
...
+    COMPARE=$?
+    if [[ ${COMPARE} = 2 ]]
+    then
+        echo "Updating develop branch version to " ${BRANCH_OR_TAG_VALUE}
+        echo ${BRANCH_OR_TAG_VALUE} > VERSION
+        git add VERSION
+        git commit -m "${COMMIT_MESSAGE}" #Skip travis build on develop commit
+        git push ${REMOTE} develop
+    else
+        echo "Develop branch VERSION value is not updated"
```
obdulia-losantos: Do you want to use here the word `VERSION`or the env var `$VERSION`?

lordmallam: Word VERSION referring to the VERSION file

scripts/release.sh
lordmallam: Yes, because there's a return function here https://github.com/eHealthAfrica/aether/pull/430/files#diff-0d28f82f2d1cc0deb12146b212778bc9R125

obdulia-losantos: Maybe you only want to remove the `e` option... :woman_shrugging: 

https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin

_This is a personal doubt, not a request or a blocker_

scripts/release.sh
```diff
@@ -59,43 +59,169 @@ function release_app {
...
+    then
+        git fetch ${REMOTE} $TRAVIS_BRANCH
+        git branch $TRAVIS_BRANCH FETCH_HEAD
+        COMMIT_BRANCH=HEAD
+        increment_version $BRANCH_OR_TAG_VALUE 3
+        BRANCH_OR_TAG_VALUE=$TAG_INCREASED_VERSION
+        echo "Version incremented to ${BRANCH_OR_TAG_VALUE}"
+    fi
+
+    git checkout $TRAVIS_BRANCH
```
admbtlr: This will check out a *new* branch with the same name as the tag (e.g. `2.4.3`). It's really hard to check out the existing branch (e.g. `release-2.4`), given only the tag (which is all you get on tag builds). I started looking at it, but haven't cracked it yet.

This command will normally give you the first branch for a given tag:

`git branch -a --contains tags/2.4.0 | grep 'release\-[0-9]' | awk '{print $2}'`

But for some reason it doesn't work on Travis 🤷‍♂️

scripts/release.sh
```diff
@@ -59,43 +60,178 @@ function release_app {
...
+        REMOTE=https://$GITHUB_TOKEN@github.com/$TRAVIS_REPO_SLUG
+    else
+        echo "Missing environment variable GITHUB_TOKEN=[GitHub Personal Access Token]"
+        exit 1
+    fi
+    version_compare $1 $2
+    COMPARE=$?
+    if [[ ${COMPARE} = 1 ]]
+    then
+        echo "VERSION value" $1 "is greater than" $3 "version" $2
```
admbtlr: I think something more needs to happen here. Right now, if I'm on `release-3.7` and the `VERSION` file says `3.7.1`, and then I add commit, the script releases `3.7.0-rc` and sets the `VERSION` file back to `3.7.0` :(

admbtlr:
Still not quite right :(

If I make a new branch, `release-5.0`, `VERSION` gets set to `5.0.1`. It should be `5.0.0` until I add a tag to trigger the 5.0.0 release.

admbtlr (dismissed):
It works! 🎉 🎉 🎉

lordmallam:
@adamvert Experiencing regression on the overall build process due to the missing `VERSION` file. Working on that.

lordmallam:
@adamvert , @obdulia-losantos issues fixed... Final review? 🤞

scripts/release.sh
```diff
@@ -59,43 +60,178 @@ function release_app {
...
+    local v=$1
+    if [ -z $2 ]; then 
+        local rgx='^((?:[0-9]+\.)*)([0-9]+)($)'
+    else 
+        local rgx='^((?:[0-9]+\.){'$(($2-1))'})([0-9]+)(\.|$)'
+        for (( p=`grep -o "\."<<<".$v"|wc -l`; p<$2; p++)); do 
+            v+=.0; done;
+    fi
+    val=`echo -e "$v" | perl -pe 's/^.*'$rgx'.*$/$2/'`
+    TAG_INCREASED_VERSION=$(echo "$v" | perl -pe s/$rgx.*$'/${1}'`printf %0${#val}s $(($val+1))`/)
```
obdulia-losantos, line 92: Did I say before that I love regular expressions?... :roll_eyes:

scripts/release.sh
```diff
@@ -59,43 +60,178 @@ function release_app {
     echo "_____________________________________________"
```
obdulia-losantos, line 54: I always think that this is not the way to flag the "latest" image, now that you check the different VERSION values and branches we could find out if the current version number is really the latest one. (Obviously in another PR)

```bash
    if [[ $VERSION != "alpha" ]]
    then
        echo "Pushing Docker image ${IMAGE_REPO}/${AETHER_APP}:latest"
        docker tag ${AETHER_APP} "${IMAGE_REPO}/${AETHER_APP}:latest"
        docker push "${IMAGE_REPO}/${AETHER_APP}:latest"
    fi
```

My concern:

We have `release-1.0`, ...`release-2.1` branches. We tagged `1.7.4`. With this condition the release `1.7.4`  is going to be the `latest` but there is already a `2.0.0` one... :woman_shrugging:

scripts/release.sh
```diff
@@ -59,43 +60,178 @@ function release_app {
...
+    else 
+        local rgx='^((?:[0-9]+\.){'$(($2-1))'})([0-9]+)(\.|$)'
+        for (( p=`grep -o "\."<<<".$v"|wc -l`; p<$2; p++)); do 
+            v+=.0; done;
+    fi
+    val=`echo -e "$v" | perl -pe 's/^.*'$rgx'.*$/$2/'`
+    TAG_INCREASED_VERSION=$(echo "$v" | perl -pe s/$rgx.*$'/${1}'`printf %0${#val}s $(($val+1))`/)
+}
+
+version_compare () {
```
obdulia-losantos, line 95: Could you comment what the different returned values (0, 1, 2) mean?

scripts/release.sh
```diff
@@ -59,43 +60,178 @@ function release_app {
...
+        fi
+        if ((10#${ver1[i]} < 10#${ver2[i]}))
+        then
+            return 2
+        fi
+    done
+    return 0
+}
+
+function git_branch_commit_and_release() {
```
obdulia-losantos, line 129: This function expects 3 arguments, once again, could you add a comment explain the meaning of each one?

scripts/aether_functions.sh
```diff
@@ -53,7 +53,11 @@ function build_ui_assets {
 function build_container {
     container=$1
     APP_REVISION=`git rev-parse --abbrev-ref HEAD`
-    APP_VERSION=`cat ./VERSION`
+    if [ ! -f VERSION ]; then
+        APP_VERSION="0.0.0"
```
obdulia-losantos, line 57: Back to `0.0.0` :joy:

lordmallam: for build purposes only when `VERSION` isn't present

obdulia-losantos: I know it ;-)

scripts/release.sh
```diff
@@ -119,17 +120,27 @@ version_compare () {
...
             return 2
         fi
     done
     return 0
 }
 
 function git_branch_commit_and_release() {
+    # Evaluates the VERSION file, increases the version value if its a tag and commit changes to base branch
+    # Expected Args:
+    ## <VERSION_ON_FILE> <VERISON_FROM_BRANCH/TAG> <BRANCH_TYPE> <IS_RC>
```
obdulia-losantos: A small ~type~ typo :grimacing:

scripts/release.sh
obdulia-losantos: Thanks!

scripts/release.sh
obdulia-losantos: So the first argument is `<VERSION_ON_FILE>` and the second one `<VERSION_FROM_BRANCH/TAG>`. :+1:

scripts/release.sh
obdulia-losantos: Are you going to remove this piece of code?

obdulia-losantos: @lordmallam ?

lordmallam: @obdulia-losantos Yes, removed

scripts/release.sh
```diff
@@ -240,9 +234,6 @@ elif [[ ${TRAVIS_BRANCH} =~ ^release\-[0-9]+\.[0-9]+$ ]]; then
     BRANCH_VERSION=${ver_number[1]}
     # Release with unified branch and file versions
     git_branch_commit_and_release ${FILE_VERSION} ${BRANCH_VERSION} branch rc
-elif [[ $TRAVIS_BRANCH = "develop" ]]; then
```
obdulia-losantos: Could you not remove this one? This is the release used in gather to test the last changes :grimacing:

obdulia-losantos (approved):
LGTM

PR #440, opened 2018-10-26
feat: include VERSION and REVISION in templates

obdulia-losantos:
It will appear in the page meta tags and in the REST API pages along with the brand.

image

image

To test it locally it's hard because in our `docker-compose` files we included the whole folder as volume and that overrides the files created in the `Dockerfile` step. I changed `common` set up to replicate the real behavior during release process.

```yml
    volumes:
      # this line overrides locally the VERSION and REVISION files
      # fixing that implies to list the folders and files one by one :(
      - ./aether-kernel:/code  
```

### DIFF
--- a/aether-common-module/Dockerfile
+++ b/aether-common-module/Dockerfile
@@ -20,6 +20,15 @@ RUN pip install -q -r /code/conf/pip/requirements.txt
 COPY ./ /code
 
 ################################################################################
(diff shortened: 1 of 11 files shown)


quesebifurcan (approved):

quesebifurcan:
Very useful!

PR #450, opened 2018-10-31
chore: add integration-test package to release script

shawnsarwar:
This will allow users of bootstrap to run integration tests, and allow us to use the same integration tests in the travis process of Aether Bootstrap.

### DIFF
--- a/docker-compose-release.yml
+++ b/docker-compose-release.yml
@@ -1,0 +1,45 @@
+version: "2.1"
+
+services:
+
+  # ---------------------------------
+  # Core
+  # ---------------------------------
+
+  kernel:
+    extends:
+      file: ./docker-compose-base.yml
+      service: kernel-base
+
+  odk:
+    extends:
+      file: ./docker-compose-base.yml
+      service: odk-base
+
+  couchdb-sync:
+    extends:
+      file: ./docker-compose-base.yml
+      service: couchdb-sync-base
+
+  ui:
+    extends:
+      file: ./docker-compose-base.yml
+      service: ui-base
+
+  # ---------------------------------
+  # Connect
+  # ---------------------------------
+
+  producer:
+    extends:
+      file: ./docker-compose-base.yml
+      service: aether-producer-base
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -87,6 +87,8 @@ CORE_APPS=( kernel odk couchdb-sync ui )
 CORE_COMPOSE='docker-compose.yml'
 CONNECT_APPS=( producer )
 CONNECT_COMPOSE='docker-compose-connect.yml'
+TEST_APPS=( integration-test )
+TEST_COMPOSE='docker-compose-test.yml'
```
obdulia-losantos: Maybe it's time to create the `docker-compose-release.yml` file with the containers to push to dockerhub.

What do you think?

shawnsarwar: So a single compose inheriting the definitions from the other files? I don't anything that contains a `depends_on:` field can't be inherited from. Any other ideas?

shawnsarwar: Yeah, boo: `ERROR: Cannot extend service 'kernel' in docker-compose.yml: services with 'depends_on' cannot be extended`

obdulia-losantos: Use `docker-compose-base.yml` instead?

shawnsarwar: :drooling_face: Yep!

docker-compose-release.yml
```diff
@@ -0,0 +1,45 @@
...
+      service: aether-producer-base
+
+  # ---------------------------------
+  # Test
+  # ---------------------------------
+
+  integration-test:
+    extends:
+      file: ./docker-compose-test.yml
+      service: integration-test
```
obdulia-losantos, line 45: _Loving the end of file warning_ :joy:

obdulia-losantos (approved):

docker-compose-release.yml
shawnsarwar: :sob:

PR #454, opened 2018-11-02
docs: add models definition

obdulia-losantos:
_filling the gaps_


quesebifurcan (approved):
+1
Could you add the command you're running to generate the ~~models~~ diagrams to the README as well, so that we're all using the same one?

PR #456, opened 2018-11-06
chore: travis pip caching and speed improvements

shawnsarwar:
The goal of this PR was to decrease the amount of time it takes for travis tests to complete. The primarily mechanism for this speedup was allowing a pip cache to speed of the myriad python builds. Travis allows for pip caching, but pip, running inside of a docker container during the docker build process cannot use a cache.

So to speedup the tests, we have parallel dockerfiles for each travis-testable module, which build pip dependencies at runtime instead of at build time, allowing for dependencies to be cached and shared between container types and test runs. This is of high time impact for modules like spavro which compiles many c extensions.

To coordinate the new Dockerfile.test builds and allow a test mode travis that has these enhancements, we have a new compose file for the travis build docker-compose-travis-test.yml which reads from docker-compose-test.yml and changes the build instructions.

The effect of these changes is that Travis time is decrease by a significant amount, from approx 10:30 of test time to approx 6:00. The caveat being we've expanded from 3 simultaneous tests to 4, which made sense as the slowest tests are much faster allowing total travis time decrease and programmer time decrease while incurring the extra overhead of running ui tests on their own.


shawnsarwar:
The goal of this PR was to decrease the amount of time it takes for travis tests to complete. The primarily mechanism for this speedup was allowing a pip cache to speed of the myriad python builds. Travis allows for pip caching, but pip, running inside of a docker container during the docker build process cannot use a cache.

So to speedup the tests, we have parallel dockerfiles for each travis-testable module, which build pip dependencies at runtime instead of at build time, allowing for dependencies to be cached and shared between container types and test runs. This is of high time impact for modules like `spavro` which compiles many c extensions.

To coordinate the new `Dockerfile.test` builds and allow a test mode `travis` that has these enhancements, we have a new compose file for the travis build `docker-compose-travis-test.yml` which reads from `docker-compose-test.yml` and changes the build instructions.

shawnsarwar:
After further review of current travis tests, Client + Kernel is also passing in about 5 minutes. For whatever reason it was taking 10:30 days ago with no apparent changes. I'll take a look at a wider sample of the current tests behavior in Travis, but this is odd. If this is the case, the improvements of this PR may not justify the additional code overhead. Smaller changes to Integration tests can be made and UI can be split from the core module tests.

shawnsarwar:
selection_010
Typical Performance.

shawnsarwar:
Closing in favor of the simpler but slower solution in #457

PR #458, opened 2018-11-08
fix: stop sending HTTP requests to Kafka to see if it's alive

shawnsarwar:
The previous method of checking to see if Kafka was alive (sending a head request and having it reject it) was causing a warning in Kafka. As @tremolo discovered, the first 4 bytes of the Kafka protocol are for message size and it looked like we wanted to send a 1.2GB message to it.
```
# org.apache.kafka.common.network.InvalidReceiveException: Invalid receive (size = 1212498244 larger than 104857600)

In [8]: struct.unpack('!i', b'HEAD')
Out[8]: (1212498244,)
```
We've replaced that check with a simple socket.connect that works and doesn't offend Kafka.

### DIFF
--- a/aether-producer/conf/pip/primary-requirements.txt
+++ b/aether-producer/conf/pip/primary-requirements.txt
@@ -17,6 +17,5 @@ flask
 gevent
 psycopg2-binary
 psycogreen
-requests
 spavro
 sqlalchemy
--- a/aether-producer/producer/aether_producer.py
+++ b/aether-producer/producer/aether_producer.py
@@ -171,8 +171,11 @@ class ProducerManager(object):
         try:
             with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
                 s.connect((kafka_ip, kafka_port))
-        except (InterruptedError, ConnectionRefusedError) as rce:
-            self.logger.debug("Could not connect to Kafka on url: %s" % kafka_url)
+        except (InterruptedError,
+                ConnectionRefusedError,
+                socket.gaierror) as rce:
+            self.logger.debug(
+                "Could not connect to Kafka on url: %s:%s" % (kafka_ip, kafka_port))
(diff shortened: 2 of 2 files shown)


aether-producer/producer/aether_producer.py
```diff
@@ -165,21 +164,21 @@ def safe_sleep(self, dur):
 
     # Connectivity
 
+    # see if kafka's port is available
     def kafka_available(self):
-        kafka_url = "http://%s" % self.settings["kafka_url"]
+        kafka_ip, kafka_port = self.settings["kafka_url"].split(":")
+        kafka_port = int(kafka_port)
```
admbtlr, line 170: Default to `80`/`443` in the (admittedly unlikely) event that there's no port?

shawnsarwar: The Kafka port defaults to 9092. It shouldn't ever be on the default http/s ports

aether-producer/conf/pip/primary-requirements.txt
```diff
@@ -17,6 +17,5 @@ flask
 gevent
 psycopg2-binary
 psycogreen
-requests
```
obdulia-losantos, line 20: If you change the primary requirements you should also change the frozen requirements.
`docker-compose -f docker-compose-connect.yml run --no-deps producer pip_freeze`

aether-producer/producer/aether_producer.py
admbtlr: OK, that's what I figured.

aether-producer/conf/pip/primary-requirements.txt
shawnsarwar: I believe that `requests` is still required by `aether.client` so it's a dependency even though it's no longer a primary one.

admbtlr (approved):
Fine by me, although @obdulia-losantos 's comment still stands

PR #460, opened 2018-11-15
chore: prepare 1.0.1 release

admbtlr:


admbtlr:
There's a lot here, but it's actually not as much as it looks... That one `feat` commit was actually cherry-picked into 1.0, so we're still being semantic...

admbtlr:
@shawnsarwar can you take a look? also, what's with the conflicts in the producer requirements?

PR #464, opened 2018-11-29
feat: producer api

shawnsarwar:
In order to make the producer easier to understand and control, we're extending the API to allow some new things.

- /topics : Allow a user to see how many individual messages are stored in Kernel and should be in Kafka / available to consumers
- /pause?topic={topic} : Temporarily stop a topic from being produced
- /resume?topic={topic} : Resume production on a paused topic
- /rebuild?topic={topic} : Allow a user to force the producer to delete a topic. Required /resume to then recreate / repopulate it. Cannot pause/ resume a topic that's in the process of being rebuilt. /status for that topic will indicate it's rebuild status.

To allow for this to operate securely, all endpoints besides /healthcheck now require authentication via basic auth. There is only one user and the credentials are set through environment.


aether-producer/producer/__init__.py
```diff
@@ -258,7 +268,7 @@ def serve(self):
         self.app.config['JSONIFY_PRETTYPRINT_REGULAR'] = self.settings.get(
             'flask_settings', {}).get('pretty_json_status', False)
         pool_size = self.settings.get(
-            'flask_settings', {}).get('max_connections', 1)
+            'flask_settings', {}).get('max_connections', 3)
```
obdulia-losantos: Wouldn't it be easier to read in this way?

```python
pool_size = self.settings \
                .get('flask_settings', {}) \
                .get('max_connections', 3)
 ```

File: aether-producer/producer/__init__.py [hunk 2]
```diff
@@ -302,6 +381,16 @@ class TopicManager(object):
         AND s.name = {schema_name}
         LIMIT 1; '''
 
+    # Count how many unique (controlled by kernel) messages should currently be in this topic
+    COUNT_STR = '''
+            SELECT
+                count(*)
```
obdulia-losantos: **count unique** would be `count(distinct e.id)`

aether-producer/producer/__init__.py
```diff
@@ -302,6 +381,16 @@ class TopicManager(object):
...
+            FROM kernel_entity e
+            inner join kernel_projectschema ps on e.projectschema_id = ps.id
+            inner join kernel_schema s on ps.schema_id = s.id
```
obdulia-losantos, line 394: we included the schema value within the entity model so you can skip one join
```sql
FROM kernel_entity e
INNER JOIN kernel_schema s ON e.schema_id = s.id
```

aether-producer/producer/__init__.py
```diff
@@ -268,12 +278,31 @@ def serve(self):
         )
         self.http.start()
 
+    # Basic Auth implementation
+
+    def check_auth(self, username, password):
+        return username == self.admin_name and password == self.admin_password
+
+    def authenticate(self):
```
obdulia-losantos: This name is confusing to me, do you pretend to *authenticate*? or to throw the missing authentication error response?

aether-producer/producer/__init__.py
```diff
@@ -100,13 +100,19 @@ def __init__(self, settings):
         signal.signal(signal.SIGINT, self.kill)
         gevent.signal(signal.SIGTERM, self.kill)
         # Turn on Flask Endpoints
+        # Get Auth details from env
+        self.admin_name = settings.get('PRODUCER_ADMIN_USER')
+        self.admin_password = settings.get('PRODUCER_ADMIN_PW')
```
obdulia-losantos, line 105: Does user require to be "admin"? :thinking:

aether-producer/producer/__init__.py
shawnsarwar: If you're on a browser, I believe this response is what prompts you to authenticate with the basic auth pop-up. Maybe `request_authentication` is more clear?

aether-producer/producer/__init__.py
shawnsarwar: I think using distinct is slightly more expensive because by constraint we can't have duplicate entity IDs in the table, right?

aether-producer/producer/__init__.py
shawnsarwar: I think we still need it. A topic in Kafka should reflect all entities that share a project schema. Several project schemas sharing the same schema could generate multiple topics.

aether-producer/producer/__init__.py
obdulia-losantos: It was a simple question, I'm not sure if this is the naming pattern for this case :woman_shrugging:

aether-producer/producer/__init__.py
obdulia-losantos: right!

aether-producer/producer/__init__.py
shawnsarwar: Since the controls allow wiping of topics from Kafka which could have large downstream consequences, I think the person doing so would be an admin or more likely in our current setup another module. The mechanism is similar to how each module has an admin key that it shares around.

aether-producer/producer/__init__.py
obdulia-losantos: but following your reasoning, why not `COUNT(e.id)`?

shawnsarwar: Yeah, that's probably better. I'm not great at SQL optimization.

quesebifurcan (approved):

PR #465, opened 2018-12-10
fix: repeated names in xform definition

obdulia-losantos:
We didn't take into consideration the cases in which the same field name is reuse in the xform definition and how the AVRO schema deals with them.

The fix: include the namespace (xpath) in each field definition

```json
{
  "name": "example",
  "type": "record",
  "fields": [
    {
      "name": "value",
      "namespace": "example",
      "type": "int"
    },
    {
      "name": "property",
      "namespace": "example",
      "type": [
        "null",
        {
          "name": "property",
          "namespace": "example",
          "type": "record",
          "fields": [
            {
              "name": "value",
              "namespace": "example.property",
              "type": "int"
            }
          ]
        }
      ]
    }
  ]
}
```


quesebifurcan:
Personally, I've never seen namespaced _fields_ in Avro records in the wild (only namespaced records). Is there a specific reason we don't use aliases for the field names?

obdulia-losantos:
There is always a first time for almost everything :grimacing: 

I think that the reason of not using aliases is that it could complicate the export to include the real names, the ones that the user included in the xform definition; and not the ones we use to avoid this kind of issues :thinking:

quesebifurcan (approved):

shawnsarwar (approved):

admbtlr (approved):
I still haven't tested it, but after looking again I'm pretty sure that this won't have any impact on existing data, so 💥

PR #466, opened 2018-12-10
feat(kernel): add extract endpoint in submission viewset

obdulia-losantos:
In case of the entity extraction failed and you don't want to use the command for ALL submissions.

Reachable at `PATCH /submissions/{pk}/extract/`


quesebifurcan (approved):

admbtlr (approved):
Nice! We should document this... 🤔

obdulia-losantos:
In other projects we use sphinx to generate the docs... maybe it's time to include it here too.

PR #467, opened 2018-12-11
feat: improve data export options

obdulia-losantos:


admbtlr (dismissed):

obdulia-losantos:
It's almost ready, I only want to include more tests with the "TAB" option and more flatten cases.

admbtlr:
Slow down, you're too fast!

aether-kernel/aether/kernel/api/tests/test_exporter.py
```diff
@@ -259,115 +283,502 @@ def setUp(self):
...
+
+        # check workbook header
+        header = [c.value for c in wb['#'][1]]
+        self.assertEqual(
+            header,
+            [
+                '@', '@id', 'id',
+                'list_1.1.1', 'list_1.2.1', 'list_1.2.2', 'list_1.1.2', 'list_1.1.3', 'list_1.4.1',
+                'list_2.1.1', 'list_2.2.1', 'list_2.1.2', 'list_2.1.3', 'list_2.1.4', 'list_2.1.5',
+            ]
```
obdulia-losantos: Wondering how to solve this :thinking: :disappointed_relieved: 

@quesebifurcan :bulb: ?

obdulia-losantos:
@adamvert you are right! I was too fast. :unamused: :snail:

admbtlr:
This looks good, but I'm going to do some tests before approving... meanwhile, what do you think @shawnsarwar? are happy to have all this export logic in kernel?

shawnsarwar:
@adamvert As long as it works for any entities and not just Gather-y things, I'm ok with it.

obdulia-losantos:
It works for any submission/entity no matter the source of the data. I hope so :thinking: 

The only method that is "odk-ish" is the `filter_paths` that removes the xForm internal fields:

```python
# remove xForm internal fields
path not in ['_id', '_version', 'starttime', 'endtime', 'deviceid', 'meta'] and
not path.startswith('meta.') and
```

This is also handled by Gather so it can be removed :woman_shrugging:

shawnsarwar:
@obdulia-losantos if Gather handles it and it's context specific I think we should remove that small section from kernel.

obdulia-losantos:
Do you want to keep this filter (aether specific, valid in submissions) ? Or do we remove it too?

```
# remove aether internal fields
f'{ENTITY_EXTRACTION_ENRICHMENT}.' not in path and
f'{ENTITY_EXTRACTION_ERRORS}.' not in path and
path not in [ENTITY_EXTRACTION_ENRICHMENT, ENTITY_EXTRACTION_ERRORS] and
```

shawnsarwar:
I think it's fine.

admbtlr (approved):
🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈🎈

PR #468, opened 2018-12-13
fix: prioritized db access in Aether Producer

shawnsarwar:
This PR shores up our previous work moving from a SQLite persistence of Offset in the producer to a Postgres instance which is easier for devops to maintain. As we currently use the same PG instance for Kernel and Producer offset, _and_ producer adds database overhead from it's constant read of entity state, we needed a careful allocation of database resources to the producer. This PR addresses that.

There are two main types of DB access. Reads of data from the Kernel PG, and reads and writes of offset data to the OffsetDB. Again, as currently configured, these are the same database. We have a couple of requirements; we can't use so much DB resource that Kernel can't read/write. Offsets need to be written ASAP so the producer can fail well on partial system outage.

To address these issues, we implemented a prioritized queue for all database connections. There are two pools, one for Kernel Reads and one for Offset Reads and writes. Offset Writes take advantage of the prioritization feature to make sure they always occur on the next open connection, while reads have to wait until there are no writes in the queue. Kernel reads are given priority due to their relative importance in the Kafka write loop. Generally, it's more important for us to know if a topic has new data to read from Kernel ( the negative case allows for a long sleep of the worker ) than it is for us to get data once we know we need to. The current default setting is 6 connections / pool, which allow plenty of overhead on the database for Kernel usage. Under the maximum current Kernel write load, Producer uses about 0.24 cores of CPU while Postgres is at about 0.5. Capping connections and keeping them alive between usage also seems to have sped up Offset I/O considerably and with liveness tests it takes on the order of 5ms.


aether-producer/producer/db.py
```diff
@@ -103,58 +131,206 @@ def make_uuid():
...
+        except Exception as err:
+            raise err
+        finally:
+            try:
+                OFFSET_DB.release(call, conn)
+            except UnboundLocalError:
+                logger.error(f'{call} could not release a connection it never received.')
+
+
+class PriorityDatabasePool(object):
```
obdulia-losantos, line 224: Is there any reason for not using the psycopg connection pool classes?

http://initd.org/psycopg/docs/pool.html

shawnsarwar: I don't believe any of those implementations allow for prioritization when requesting a connection.

obdulia-losantos: I was more thinking on extending, let's say, `ThreadedConnectionPool`, and implement the prioritization methods... but yes, it makes little sense in this case anyway.

shawnsarwar: I'd read that mixing threads with greenlets could be problematic. Originally we were using SqlAlchemy's connection pooling, but it was losing track of connections and overflowing randomly for no good reason (postgres was not nearly stressed). I assumed it was for lack of explicit compatibility with greenlets so I decided to base the design on the greenlet implementation of queue which seems to be quite stable.

obdulia-losantos: in that case... :crossed_fingers:

obdulia-losantos (approved):

PR #470, opened 2019-01-08
chore: better VERSION and REVISION handling

obdulia-losantos:
- updated VERSION file to real and current version number
- added `/check-app` endpoint that will return the current app version number and...
- use temporal files to handle VERSION and REVISION locally

Pending (and maybe in a new PR) cleaning the script files :nerd_face:


shawnsarwar (approved):
:+1:

PR #474, opened 2019-01-21
feat(sync): load backup files as synced data

obdulia-losantos:
I can imagine that loading the backup files generated from a device is not going to be a one-time task... this could be a permanent solution. (still missing tests)

Open question:
- the endpoint is only available for "admin" users, should this be less restrictive? I'm not sure about the multi-tenancy effect...


admbtlr:
The problem with this approach is that it requires a new deployment - I was hoping we could just write a script that would upload the JSON records directly to Kernel.

obdulia-losantos:
Changing docker-compose file to include CouchDB server url and credentials will push the document to that server, not the local one.

Although I would test it locally first :scream:

admbtlr (approved):
Oops, forgot about this! Looks good 👍

PR #479, opened 2019-01-28
wip /submission performance

shawnsarwar:


quesebifurcan:
TODO:
- ~~fix tests~~
- ~~make django_rq optional~~
- rq worker configuration (how many workers?)
- check in with devops re:helm-chart for kernel-rq

shawnsarwar:
The theoretical minimum operation time for a single submission without extraction is on the order of 100MS due to request overhead (checking auth, etc). To improve throughput significantly, we need to do the following.

- Allow bulk submission in a single request, with each submission being enqueued individually.
- Allow bulk extraction in a background process with work dispatched and product written to Redis. This should be a CPU bound operation.
- Refactor serialize.save() operations into a serialize.presave() method. (bulk_create does not call save.)
- Allow bulk_create writing on many extracted entities from Redis via the Django ORM, as single inserts are slow.

PR #484, opened 2019-02-04
feat: extractor handle nested jsonpath destinations

shawnsarwar:
Adding the ability to insert data into nested objects in entities. Previously, you could have an entity with path `Person.location.lat`, but you could only assign an object to `Person.location`. This object would have to satisfy all requirements for the `location` object by itself.

With this patch, you can now have two separate mapping steps for:
`lat -> Person.location.lat` 
`lng -> Person.location.lng`
This makes the mappings slightly more flexible.

I've updated the mapping validators to support this behavior and added tests. It's working as expected from AUX.


aether-kernel/aether/kernel/api/entity_extractor.py
```diff
@@ -175,14 +177,23 @@ def get_entity_requirements(entities, field_mappings):
         entity_requirements = {}
         # find mappings that start with the entity name
         # and return a list with the entity_type ( and dot ) removed from the destination
-        matching_mappings = [[src, dst.split(entity_type+'.')[1]]
+        start_string = entity_type+'.'
```
obdulia-losantos: didn't flake8 complain? :flushed:

aether-kernel/aether/kernel/api/entity_extractor.py
```diff
@@ -436,7 +481,9 @@ def extract_entity(entity_type, entities, requirements, data, entity_stub):
     # make empty stubs for our expected outputs. One for each member
     # identified in count process
     entities[entity_type] = [
-        {field: None for field in entity_stub.keys()} for i in range(count)]
+        {field: None for field in entity_stub.keys()
+            if field in requirements.get(entity_type)}  # make sure there are mappings
```
obdulia-losantos: I know that this is me but don't you think that it's more clear to read this way?
```python
{
    field: None
    for field in entity_stub.keys()
    if field in requirements.get(entity_type)  # make sure there are mappings
}
```

aether-kernel/aether/kernel/api/entity_extractor.py
shawnsarwar: Strange that it didn't...

aether-kernel/aether/kernel/api/entity_extractor.py
shawnsarwar: :+1: for readability

aether-kernel/aether/kernel/api/tests/__init__.py
```diff
@@ -78,6 +90,34 @@
...
+            'name': 'lng',
+            'type': 'int'
+          }
+        ],
+        'name': 'Nested'
+      }
+    }
+  ],
+  'name': 'Nested'
+}
```
obdulia-losantos, line 119: In python the indentation is with 4 spaces ;-)

shawnsarwar: Where is my linter when I need it!?

aether-kernel/aether/kernel/api/tests/__init__.py
```diff
@@ -140,6 +180,26 @@
...
+            'name': 'b',
+            'lat': 11,
+            'lng': 21
+        },
+        {
+            'name': 'c',
+            'lat': 12,
+            'lng': 22
+        },
+    ]
```
obdulia-losantos: :straight_ruler:

aether-kernel/aether/kernel/api/tests/__init__.py
```diff
@@ -167,13 +227,55 @@
...
+            },
+            {
+                'name': 'c',
+                'location': {
+                    'lat': 12,
+                    'lng': 22
+                }
+            },
+        ]
+    }
```
obdulia-losantos: spaces all around :rocket:

obdulia-losantos (approved):

PR #486, opened 2019-02-12
feat: place values into Entity array

shawnsarwar:
Previously when dealing with Entities with a field that had an array as its value, it was only possible to map a whole array from the source into that field. We've extended entity extraction to allow for elements to be added one to by the an array inside of an entity.

For example this is now a valid example.
For source
```
{
  "a": 1.0,
  "b": 2.0
}
```
and entity Obj with structure:
```
{
  "geo" : [ ]
}
```
mappings:
```
[
  ["$.a", "Obj.geo[0]"],
  ["$.b", "Obj.geo[1]"]
]
```
should yield:
```
{
  "geo" : [1.0, 2.0]
}
```

### DIFF
--- a/aether-kernel/aether/kernel/api/entity_extractor.py
+++ b/aether-kernel/aether/kernel/api/entity_extractor.py
@@ -54,6 +54,9 @@ ENTITY_EXTRACTION_ENRICHMENT = 'aether_extractor_enrichment'
 # $.path.key_*.to[*].field for $.path.key_27.to[1].field
 CUSTOM_JSONPATH_WILDCARD_REGEX = re.compile(
     r'(\$)?(\.)?([a-zA-Z0-9_-]*(\[.*\])*\.)?[a-zA-Z0-9_-]+\*')
+# RegEX to split off array accessors in keynames
+ARRAY_ACCESSOR_REGEX = re.compile(
+    r'[^[\]]')
 
 # RegEx for the part of a JSONPath matching the previous RegEx which is non-compliant with the
 # JSONPath spec.
@@ -380,8 +383,27 @@ def put_nested(_dict, keys, value):
         except KeyError:  # Level doesn't exist yet
             _dict[keys[0]] = put_nested({}, keys[1:], value)
     else:
-        _dict[keys[0]] = value
+        last_key = keys[0]
+        if '[' not in last_key:
+            _dict[last_key] = value
+        else:
+            matches = ARRAY_ACCESSOR_REGEX.findall(last_key)
+            key, index = matches[0], int(matches[1])
+            if key not in _dict:
+                _dict[key] = None
+            _dict[key] = put_in_array(_dict[key], index, value)
     return _dict
+
+
+def put_in_array(arr, idx, val):
+    out = None
+    if not isinstance(arr, list):
+        out = []
+    else:
+        out = arr
+    # best effort to insert at the requested index
+    out.insert(idx, val)
+    return out
 
 
 def resolve_source_reference(path, entities, entity_name, i, field, data):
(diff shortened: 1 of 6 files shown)


aether-kernel/aether/kernel/api/mapping_validation.py
```diff
@@ -19,11 +19,15 @@
 import collections
 import json
 from django.utils.translation import ugettext as _
+import re
 
 from .entity_extractor import find_by_jsonpath
 
 import spavro
 
+ARRAY_ACCESSOR_REGEX = re.compile(
```
obdulia-losantos: You can take this constant from `entity_extrator`, do not redefine it twice :wink:

aether-kernel/aether/kernel/api/entity_extractor.py
```diff
@@ -379,11 +383,43 @@ def put_nested(_dict, keys, value):
             _dict[keys[0]] = put_nested(_dict[keys[0]], keys[1:], value)
         except KeyError:  # Level doesn't exist yet
             _dict[keys[0]] = put_nested({}, keys[1:], value)
-    else:
+    # Key has an array accessor like child[0]
+    # Array accessors are only allowed at the last path element
+    elif '[' in keys[0]:
```
obdulia-losantos: wouldn't it be better `if ARRAY_ACCESSOR_REGEX.match(keys[0]) is not None:`?

aether-kernel/aether/kernel/api/mapping_validation.py
```diff
@@ -59,15 +63,30 @@ def validate_getter(obj, path):
...
+
+
 def validate_existence_in_fields(field_name, path, definition):
     # looks an a schema object with property fields
     # if field_name matches one of the properties, we mark success
     # if a field_name is a dotted match child.name -> child
     # we recurse and look for a nested object
+
+    # if this field is an array accessor, we'll perform extra checking
+    check_array = is_array_accessor if '[' in field_name else always_false
```
obdulia-losantos: Same comment
`if ARRAY_ACCESSOR_REGEX.match(field_name)`
instead of 
`if '['`

obdulia-losantos (changes requested):
little things

aether-kernel/aether/kernel/api/mapping_validation.py
shawnsarwar: :+1:

aether-kernel/aether/kernel/api/entity_extractor.py
shawnsarwar: The regex is greedy and only splits on `[` & `]` so anything with <2 matches isn't an array accessor.

shawnsarwar:
@obdulia-losantos Update from review.

aether-kernel/aether/kernel/api/tests/test_mapping_validation.py
```diff
@@ -20,13 +20,15 @@
...
+
 valid_schemas = {
+    'Nested': NESTED_ARRAY_SCHEMA,
     'Person': {
         'name': 'Person',
         'type': 'record',
         'namespace': 'test.Person',
-        'fields': [
-            {
+        'fields': [{
```
obdulia-losantos, line 31: :joy: Your linter has its own thoughts!

obdulia-losantos (approved):

aether-kernel/aether/kernel/api/tests/test_mapping_validation.py
shawnsarwar: I just added an auto-linter extension to my editor and it complained about the original format. I'm also not sure I agree. :frowning_man:

PR #499, opened 2019-02-19
fix: invalid table name for producer offsets

shawnsarwar:
We were using the name `offset` for the offsets table used by the producer. This is a reserve word in Postgres.


obdulia-losantos (approved):

shawnsarwar:
Offsets are moving out of Postgres in an upcoming patch

PR #503, opened 2019-02-27
feat: filter on dates

shawnsarwar:
Previously only exact value would work for filtering entities and submissions on created / modified. This patch allow for gt(e)/lt(e) operations on those fields via the API.

### DIFF
--- a/aether-kernel/aether/kernel/api/filters.py
+++ b/aether-kernel/aether/kernel/api/filters.py
@@ -112,10 +112,13 @@ class SubmissionFilter(filters.FilterSet):
                 'filter_class': filters.IsoDateTimeFilter
             },
         }
-        fields = {
-            'created': ('lt', 'gt', 'lte', 'gte'),
-            'modified': ('lt', 'gt', 'lte', 'gte')
-        }
+        # since we can't use __all__ and then extend the syntax on specific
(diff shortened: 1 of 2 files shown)


aether-kernel/aether/kernel/api/filters.py
```diff
@@ -105,9 +105,20 @@ def mappingset_filter(self, queryset, name, value):
...
+        filter_overrides = {
+            model.created: {
+                'filter_class': filters.IsoDateTimeFilter
+            },
+        }
+        # since we can't use __all__ and then extend the syntax on specific
+        # fields, we have to enumerate all fields and give them the exact
+        # filter
+        fields = {str(k.name): ('exact',) for k in model._meta.get_fields()
+                  if k.name not in ['payload']}  # exclude not accessible from here
```
obdulia-losantos, line 120: You know my comment in this line, don't you?

aether-kernel/aether/kernel/api/filters.py
```diff
@@ -202,9 +213,20 @@ def passthrough__filter(self, queryset, name, value):
...
```
obdulia-losantos, line 229: and the same comment :innocent:

obdulia-losantos (approved):

PR #505, opened 2019-03-01
docs: add Troubleshooting section

obdulia-losantos:
I don't think that I'm the person for this PR but if you can check it (do not hesitate to fix it if you find English errors...) and add your own solutions. It would be GREAT.


admbtlr (approved):
¡Muy bien!

PR #508, opened 2019-03-04
fix(ui): deprecated componentWillReceiveProps

obdulia-losantos:
Closes: https://jira.ehealthafrica.org/browse/AET-390

Also fixes a bug on `deepEqual` (was breaking on array values).

Added `objectToString` utility that replaces `JSON.stringify` in the components.


admbtlr (approved):
Nice

PR #509, opened 2019-03-04
fix(ui): restyling contract name form

obdulia-losantos:
Still needs @sabeen magic on it.

image

Closes https://jira.ehealthafrica.org/browse/AET-479

### DIFF
--- a/aether-ui/aether/ui/assets/apps/pipeline/sections/Settings.jsx
+++ b/aether-ui/aether/ui/assets/apps/pipeline/sections/Settings.jsx
@@ -169,33 +169,39 @@ class Settings extends Component {
                 defaultMessage='Contract name'
               />
             </label>
-            <input
-              type='text'
-              required
-              name='name'
-              className='input-d contract-name'
-              value={this.state.contractName}
-              onChange={(e) => { this.setState({ contractName: e.target.value }) }}
-              disabled={contract.is_read_only}
-            />
+            <div className='row'>
+              <div className='col'>
+                <input
+                  type='text'
+                  required
+                  name='name'
+                  className='input-d contract-name'
+                  value={this.state.contractName}
+                  onChange={(e) => { this.setState({ contractName: e.target.value }) }}
+                  disabled={contract.is_read_only}
+                />
+              </div>
 
-            { !contract.is_read_only &&
-              <button
-                className='btn btn-d btn-primary btn-big ml-4'
-                onClick={() => {
-                  this.props.updateContract({
-                    ...contract,
-                    name: this.state.contractName
-                  })
-                }}>
-                <span className='details-title'>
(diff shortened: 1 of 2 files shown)


admbtlr:
LGTM - what do you think @sabeen ?

sabeen:
we usually have the buttons under the form on the left hand side.
Also there is a classname like "large" or "big" input, that adds a lot of padding, which I would apply since there is not a lot of content on that page.

admbtlr:
OK, so let's leave this one for @sabeen  to handle - is that OK for you @obdulia-losantos ?

obdulia-losantos:
@sabeen all yours!!!

sabeen:
ok :-)

aether-ui/aether/ui/assets/apps/pipeline/sections/Settings.jsx
```diff
@@ -230,17 +210,24 @@ class Settings extends Component {
             }
           </div>
 
-          <button
```
obdulia-losantos, line 227: In this page, each section is a form, so this "save" button refers only to the "name" field, could you move it to its section? :grimacing:

sabeen: but this comes with the constraint, that the save action needs to be placed right after the text input. and then we need another button at the bottom. 
It changes the flow and adds noise, that is why I am not totally happy about it.
But it is also not a show-stopper, so I will bring it back and maybe do something with sizes and wording.

admbtlr: We don't need another save button at the bottom because there's nothing on the page to save apart from the title. If you press _Create identity contract_, the relevant stuff is created in UI without needing to press another _save_ button.

In fact, since the _Save_ button is only saving the name, could we think about taking it out and just saving whenever the name changes? Or maybe do something like Jira's individual button for the title field?

image

(Can't believe I'm taking UX ideas from Jira...)

sabeen: Yes, actually that was something I was thinking of, just having a little check button at the end of the textfield.
And at the bottom I’ll put a big „finish“ button, which actually does nothing but closing the page.

sabeen:
Yes, actually that was something I was thinking of, just having a little check button at the end of the textfield.
And at the bottom I’ll put a big „finish“ button, which actually does nothing but closing the page.

admbtlr:
If we have a big _Finish_ button, then I wonder whether we also need a _Close_ button at the top?

obdulia-losantos:
@sabeen can we close this PR?

sabeen:
yes

PR #512, opened 2019-03-07
feat(kernel): project schema CASCADE entities

obdulia-losantos:
I know that this is not the time for this PR but I'm cleaning all the migrations files for multitenancy and including a new one in the middle of the process will kill me :( So I prefer to do it right now and update my branch with this.


admbtlr (approved):
👍

PR #522, opened 2019-03-18
fix: remove any name field restriction

obdulia-losantos:
Closes: https://jira.ehealthafrica.org/browse/AET-495

In kernel:
- Remove in `name` field the unique constraint for all models except for `Schema` model.
- Remove in `name` field the length size restriction in all models
- Remove in `type` field the length size restriction in `Schema` model
- Adapt `upsert_project_artefacts` method to new `name` restrictions.

In UI:
- Remove in `name` field the length size restriction


lordmallam (dismissed):
LGMT, We will have to refactor the `ui - preflight` functions to remove the constraints as well

obdulia-losantos:
There are no more "name" constraints in the preflight methods :wink:

aether-kernel/aether/kernel/api/models.py
```diff
@@ -304,7 +304,7 @@ class Attachment(ExportModelOperationsMixin('kernel_attachment'), ProjectChildAb
 
     # http://www.linfo.org/file_name.html
     # Modern Unix-like systems support long file names, usually up to 255 bytes in length.
```
admbtlr: So we should probably take out this comment too, shouldn't we?

admbtlr (changes requested):

aether-kernel/aether/kernel/api/models.py
obdulia-losantos: Right!

admbtlr (approved):

PR #525, opened 2019-03-18
feat(ui): pipeline is_read_only property

obdulia-losantos:
Not sure is this will fix the following error, but these are the only "contracts" mentions I found in UI. :woman_shrugging: 

image


admbtlr (approved):
I can't actually recreate this bug, but the PR looks good, so 🤷‍♂️

PR #533, opened 2019-03-22
feat: introduce keycloak

obdulia-losantos:
# Before start the app

- Generate a new `.env` file with the keycloak variables.
- Execute once  `./scripts/setup_keycloak.sh` to generate keycloak initial assets.
- Check `.env` file for the initial user and password.
- Visit `http://localhost:8080` and create more users and realms.

---
**IMPORTANT NOTE**

App users authenticate against keycloak and admin users against Django Auth. This means that no keycloak user can access the admin section (forced by the fact that the admin section has no realm restrictions).

---

**Note**: you don't need to build `common` if you don't want to; include the `volumes` entry in each aether modules in `docker-compose.yml` file and the module will be accessible with the last changes.

```yml
  kernel:
    extends:
      file: ./docker-compose-base.yml
      service: kernel-base
    # ...
    volumes:
      - ./aether-common-library/aether/common:/code/aether/common
```
---
## keycloak

Go to http://localhost:8080

Read: https://www.keycloak.org/documentation.html if you need help to set it up properly.

### first page
image

### `dev` realm

image

image

### `aether` client in the realm
image


README.md
```diff
@@ -364,6 +370,21 @@ To start any container separately:
 
 ### Users & Authentication
 
+Set the `KEYCLOAK_SERVER_URL` and `KEYCLOAK_CLIENT_ID` environment variables if
+you want to use Keycloak as authentication server.
+`KEYCLOAK_CLIENT_ID` (defaults to `aether`) is the public client that allows
+the aether module to authenticate using the Keycloak REST API.
+This client id must be added to all the realms used by the aether module.
```
shawnsarwar, line 377: It is required to create the `aether` client in each realm and include KEYCLOAK_CLIENT_ID in a production instance that doesn't use basic_auth?

README.md
```diff
@@ -364,6 +370,21 @@ To start any container separately:
...
+Execute once the `./scripts/setup_keycloak.sh` script to create the keycloak
+database and the default realm+client along with the first user
+(find credentials in the `.env` file).
+
+Read more in [Keycloak](https://www.keycloak.org).
+
+**Note**: Multi-tenancy is automatically enabled if the authentication server
+is keycloak.
+
 Set the `HOSTNAME` and `CAS_SERVER_URL` environment variables if you want to
```
shawnsarwar, line 388: Is the CAS stuff still relevant?

aether-common-library/aether/common/conf/urls.py
```diff
@@ -61,39 +61,43 @@ def generate_urlpatterns(token=False, kernel=False, app=[]):  # pragma: no cover
...
+            template_name=settings.LOGIN_TEMPLATE,
+            authentication_form=get_auth_form(),
+        )
         logout_view = views.LogoutView.as_view(template_name=settings.LOGGED_OUT_TEMPLATE)
 
     auth_views = [
-        path('login/', view=login_view, name='login'),
-        path('logout/', view=logout_view, name='logout'),
+        path(route='login/', view=login_view, name='login'),
+        path(route='logout/', view=logout_view, name='logout'),
```
shawnsarwar, line 74: Does the logout view hit the Keycloak session logout endpoint? Or is that only be required for OIDC?

README.md
obdulia-losantos: We cannot remove it, it's being used in all our current deployments. So, yes, it's still relevant.

aether-common-library/aether/common/conf/urls.py
obdulia-losantos: Yes, but only when there is a keycloak token within the session values. Otherwise there is no way of telling keycloak which user is trying to log out.

Take a look at the signal:
https://github.com/eHealthAfrica/aether/pull/533/files?file-filters%5B%5D=.py#diff-ff424d29dd76ca35340d7d00829f59f0R126

README.md
obdulia-losantos: Because we are using keycloak "behind the scenes" we need a client to get the user token, the client can be different for each module I use the same in order to simplify the development, but you could have one for kernel and another for the rest... :woman_shrugging: 

```
curl --data \
    "grant_type=password&client_id={CLIENT}&username={user}&password={password}" \
    http://localhost:8080/auth/realms/{REALM}/protocol/openid-connect/token`
```

This is the one of the possible implementation, the other one is going to keycloak, authenticate there and take the cookie from the response...

shawnsarwar (approved):
Looks great!

obdulia-losantos:
@lordmallam @adamvert  is it ok for you?

admbtlr (approved):
🎉 🎉 🎉 🎉

PR #539, opened 2019-04-02
feat: static_url env

shawnsarwar:
Allow static assets to have a different base_url than the rest of the application. This allows UWSGI static assets to be served from a public endpoint when the rest of the service is multi tenanted.

For example, we can have and `APP_URL` that contains wildcards and is multi-tenanted like:
`APP_URL = <realm>/<service>/`
and 
`STATIC_URL = _static/kernel/static/`

Both will be proxied properly with the kong routes configuration:

```
{
    "name": "kernel",
    "service_url": "http://kernel:8000",
    "oidc_endpoints": [
        {
            "name": "base",
            "url": "/",
            "strip_path": false
        }
    ],
    "global_public_endpoints": [
        {
            "name": "static",
            "url": "/_static/kernel/static/",
            "strip_path": true
        }
    ]
}
```

### DIFF
--- a/aether-common-library/aether/common/conf/settings.py
+++ b/aether-common-library/aether/common/conf/settings.py
@@ -39,8 +39,7 @@ USE_I18N = True
 USE_L10N = True
 USE_TZ = True
 
-STATIC_URL_PREFIX = os.environ.get('STATIC_URL_PREFIX', '/')
-STATIC_URL = f'{STATIC_URL_PREFIX}static/'
+STATIC_URL = os.environ.get('STATIC_URL', '/static/')
 STATIC_ROOT = os.environ.get('STATIC_ROOT', '/var/www/static/')
 
 PRETTIFIED_CUTOFF = int(os.environ.get('PRETTIFIED_CUTOFF', 10000))
--- a/aether-common-library/aether/common/conf/settings.py
+++ b/aether-common-library/aether/common/conf/settings.py
@@ -39,7 +39,8 @@ USE_I18N = True
 USE_L10N = True
 USE_TZ = True
 
-STATIC_URL = f'{APP_URL}static/'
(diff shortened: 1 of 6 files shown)


aether-common-library/aether/common/conf/settings.py
```diff
@@ -39,7 +39,8 @@
 USE_L10N = True
 USE_TZ = True
 
-STATIC_URL = f'{APP_URL}static/'
+STATIC_URL_PREFIX = os.environ.get('STATIC_URL_PREFIX', '/')
+STATIC_URL = f'{STATIC_URL_PREFIX}static/'
```
obdulia-losantos: Sorry, what I was going to say is that the expected variable for this is `STATIC_URL`

```python
# settings.py
STATIC_URL = os.environ.get('STATIC_URL', '/static')
```

```bash
# uwsgi/start.sh
if [ ! -z "${CUSTOM_UWSGI_SERVE_STATIC:-}" ]; then
    export UWSGI_STATIC_EXPIRES=${UWSGI_STATIC_EXPIRES:-"/* 7776000"}

    STATIC_URL=${STATIC_URL:-/static}
    STATIC_DIR="/var/www/static"

    MAP_STATIC="--static-map ${STATIC_URL}=${STATIC_DIR}"
    # favicon.ico is always in root
    MAP_FAVICO="--static-map2 /favicon.ico=${STATIC_DIR}/aether/images/aether.ico"
    STATIC_CONTENT="$MAP_STATIC $MAP_FAVICO"
fi
```

shawnsarwar: Changed to `STATIC_URL` and removed the default static append from uwsgi.

obdulia-losantos (approved):

PR #548, opened 2019-04-15
chore: use docker network for internal urls

obdulia-losantos:
Instead of using the simplest docker network for establish the communication among the containers let's go one step further and take advantage of its features to use the same naming internally and externally.

*What do we need?*
- create the docker network with `subnet` and `gateway` arguments.
- use a fixed IP for nginx container (`ipv4_address`: `nginx_ip`).
- create hosts entries for the rest of containers (`extra_hosts`: `aether.local:nginx_ip`).

To make it simple the way the containers are accessed changed from `http://{module].aether.local/` to `http://aether.local/{module}/`

*PROS*: 
- `etc/hosts` file contains only one entry for the whole aether platform.
- If we need to expose the kernel url (as it happens in UI with the submission url) the url is not `http://kernel:8100/` but `http://aether.local/kernel/`.

*CONS*
- You need to change your `etc/hosts` file.
- You also need to recreate the `.env` file with the new environment variables.

This is not a breaking change, but it breaks the way we currently develop the app.


README.md
```diff
@@ -312,13 +285,13 @@ The default values for the export feature:
 #### Aether ODK Module
 
 - `AETHER_KERNEL_TOKEN`: `kernel_any_user_auth_token` Token to connect to kernel server.
-- `AETHER_KERNEL_URL`: `http://kernel:8100` Aether Kernel Server url.
+- `AETHER_KERNEL_URL`: `http://aether.local/kernel/` Aether Kernel Server url.
 - `AETHER_KERNEL_URL_TEST`: `http://kernel-test:9100` Aether Kernel Testing Server url.
```
admbtlr, line 289: So the test URLs don't change?

shawnsarwar:
Does this imply that in a deployment, inter-app communication would use the external url for the other services?

README.md
obdulia-losantos: No, they don't because for the tests we are not using nginx.

obdulia-losantos:
@shawnsarwar you can use both, the internal and the external, but if we want to mimic the "production" behavior we should use the "external" urls.

shawnsarwar (dismissed):

admbtlr:
When I go to http://aether.local/kernel, it redirects me to http://aether.local/kernel/accounts/login/?next=/kernel/ and can log in just fine.

But when I go to http://aether.local/ui, it redirects me to http://aether.local/accounts/login/?next=/ui/ and 404s.

admbtlr:
Also, I thought that using Keycloak would solve the problem that logging into a module logs you out of other modules, but right now that's not the case :(

shawnsarwar:
@adamvert It absolutely should not. Are you using the /dev/kernel/ paths?

obdulia-losantos:
I changed it a little bit. Now UI is the root url `http://aether.local/`.

The problem with single-sign-on is that you need to indicate to the other modules the realm, once you do that (in the realm form page) you don't need to log in again.

admbtlr (approved):

PR #557, opened 2019-04-25
feat: bulk submission of entities

shawnsarwar:
The REST API is slow in single submission mode, even for entities. With request overhead, it takes about 120ms / request. Stretching that out for 1000 submissions yielded this:
`generated 1000 entities @ 7.961903184992965 / second `
This patch allows submission of 100 entities in 0.58 seconds or  172 / second.
There exists an optimal batch size, probably approaching 500 entities, which should take around 2 seconds to process.

### DIFF
--- a/aether-kernel/aether/kernel/api/views.py
+++ b/aether-kernel/aether/kernel/api/views.py
@@ -435,6 +435,15 @@ class EntityViewSet(MtViewSetMixin, ExporterViewSet):
     schema_field = 'schemadecorator__schema__definition'
     schema_order = '-schemadecorator__schema__created'
 
+    def get_serializer(self, *args, **kwargs):
+        if 'data' in kwargs:
+            data = kwargs['data']
+
+            if isinstance(data, list):
+                kwargs['many'] = True
+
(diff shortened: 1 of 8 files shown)


aether-kernel/aether/kernel/api/views.py
```diff
@@ -435,6 +435,15 @@ class EntityViewSet(MtViewSetMixin, ExporterViewSet):
     schema_field = 'schemadecorator__schema__definition'
     schema_order = '-schemadecorator__schema__created'
 
+    def get_serializer(self, *args, **kwargs):
+        if 'data' in kwargs:
+            data = kwargs['data']
+
+            if isinstance(data, list):
+                kwargs['many'] = True
```
obdulia-losantos: All in one line :nerd_face: 
```python
kwargs['many'] = isinstance(kwargs.get('data') , list)
```

shawnsarwar: Nice one liner! Very clear.

obdulia-losantos (approved):

PR #567, opened 2019-05-08
fix(odk): add content-disposition in downloadUrl (#556)

obdulia-losantos:


djmoran64 (approved):
That looks beautiful!

admbtlr (approved):

PR #568, opened 2019-05-09
feat: (ui, kernel) delete contracts and pipelines

lordmallam:
Implements deleting contracts/pipelines with related kernel artefacts from the UI

## Details
Implements logic to determine if a schema is uniquely used in the context of a list of mappings.
`POST /schemas/unique-usage`. Call made with a list of mapping ids 

Implements deleting the following on kernel:
- Mappings and related schemas, entities, schema decorators `POST /mappings/[contract__id]/delete-artefacts/`
- Mappingsets and related mappings, schemas, entities, schema decorators `POST /mappingsets/[mappingset__id]/delete-artefacts/`

Delete request are made with an options object indicating the artefacts to delete
```
{
schemas: true,
entities: true,
submissions: true
}
```
@sabeen we will need some styling on the [`DeleteStatus`](https://github.com/eHealthAfrica/aether/blob/ui-delete-operations/aether-ui/aether/ui/assets/apps/pipeline/components/DeleteStatus.jsx) Modal pleasssseeeeee 😃


aether-kernel/aether/kernel/api/migrations/0103_submission_no_cascade_delete.py
```diff
@@ -0,0 +1,42 @@
...
+                to='kernel.MappingSet',
+                verbose_name='mapping set',
+            ),
+        ),
+        migrations.AlterField(
+            model_name='submission',
+            name='project',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='submissions', to='kernel.Project', verbose_name='project'),
+        ),
+        migrations.RunPython(
```
obdulia-losantos: shouldn't this migration be before the `submission.project` one? I mean, please ensure that all the submission entries have `project` field and after that alter the field.

aether-kernel/aether/kernel/api/models.py
```diff
@@ -246,16 +246,15 @@ class Submission(ExportModelOperationsMixin('kernel_submission'), ProjectChildAb
 
     mappingset = models.ForeignKey(
         to=MappingSet,
-        on_delete=models.CASCADE,
+        on_delete=models.SET_NULL,
```
obdulia-losantos: Linked to this change and to ensure that the multitenancy will working you need to remove this method in this model

```python
    def get_mt_instance(self):
        # because project can be null we need to override the method
        return self.mappingset.project
```

You need to adapt also the serializer and the ViewSet.

obdulia-losantos: Another required change...
```python
def save(self, *args, **kwargs):
        self.project = self.mappingset.project  # <-- this fails if mappingset was deleted!!!
        super(Submission, self).save(*args, **kwargs)
```

aether-kernel/aether/kernel/api/project_artefacts.py
```diff
@@ -184,7 +185,12 @@ def upsert_project_artefacts(
             }
 
         # check for the mapping set
-        mappingset_id = raw_mapping.get('mappingset', raw_mapping.get('id'))
+        mappingset_id = raw_mapping.get('mappingset')
+        if not mappingset_id:
+            if mappingset:
```
obdulia-losantos: why are you using a global mappingset?

lordmallam: To allow a one-many pairing for mappings to a mappingset in one call.

obdulia-losantos: Can you write and example of that case? I don't understand why using a global mappingset helps in that case...

obdulia-losantos: Take in mind that with your approach you are always using the LAST mappingset, you need to indicate the specific mappingset for the mapping, otherwise you create a new mappingset that shares the mapping id.

When you want to create a "group" (avoiding set here) of mappings assigned to a certain mappingset you create the mappingset providing its id and the mappings using that id.

```json
{
  "mappingsets": [
    {"id": 1, "name": "1"}, 
    {"id": 2, "name": "2"}, 
    {"name": "with-a-random-id"}
  ],
  "mappings": [
    {"name": "m-1-1", "mappingset": 1},
    {"name": "m-1-2", "mappingset": 1},
    {"name": "m-2-1", "mappingset": 2},
    {"name": "m-2-2", "mappingset": 2},
    {"name": "with-a-new-mappingset"},
  ]
}
```

lordmallam: In a case where you can't know the assigned `mappingset.id` to set the `mapping.mappingset` field before making the call.

Assuming I have the payload: 
```
{
   mappingsets: [
       {name: 'mappingset-1'}
   ],
   mappings: [
     {name: 'mapping-1'},
     {name: 'mapping-2'}
   ]
}
```
Note: without any `ids`.
I will assume both mappings should belong to the same mappingset. With the previous implementation you will end up with 3 mappingsets and the mappings each linked to one of the two extra mappingsets created.

lordmallam: Moreover, I would like to think of the concept of posting 'artefacts' in the context of a `pipeline` = 1 `mappinset`. Be it from the `ui` or `gather` its always in the context of creating a (1) pipeline within a project

obdulia-losantos: My problem now is that you are breaking this method and when Gather propagates the xForms from ODK to Kernel it's creating more than ONE mappingset.... and the side effect can be a nightmare. :disappointed:

aether-kernel/aether/kernel/api/views.py
```diff
@@ -312,13 +313,39 @@ class MappingSetViewSet(MtViewSetMixin, viewsets.ModelViewSet):
     filter_class = filters.MappingSetFilter
     mt_field = 'project'
 
+    def destroy(self, request, pk=None):
```
obdulia-losantos: Could it be possible not to override the standard method and create a new one for this? The standard REST expects a 204 NO CONTENT response for the DELETE action... :woman_shrugging:

aether-kernel/aether/kernel/api/views.py
```diff
@@ -612,3 +638,38 @@ def run_mapping_validation(submission_payload, mapping_definition, schemas):
         })
     except Exception as e:
         return Response(str(e), status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@api_view(['POST'])
+@renderer_classes([JSONRenderer])
+@permission_classes([permissions.IsAuthenticated])
+def mapping_schema_unique_usage_view(request, *args, **kwargs):
```
obdulia-losantos: Why isn't this method declared inside the `SchemaViewSet` as a custom action? with that you don't need to indicate the renderer classes, the permission classes... and you take advantage of all the things we made for multitenancy, security, gateway... you could also take advantage of the filtering and search already implemented.

```python
@action(detail=True, methods=['get', 'post'], url_path='unique-usage')
def unique_usage(self, request, pk=None, *args, **kwargs):
    # return the list of schemas that are reference only once, in one project, in one mappingset or in one mapping...
```

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -52,3 +56,125 @@ def merge_objects(source, target, direction):
...
+    schemas = models.Schema.objects.filter(schemadecorators__mappings__id__in=mappings_ids)
+    for schema in schemas:
+        other_linked_mappings = models.Mapping.objects.filter(
+            schemadecorators__schema__id=schema.id
+        ).exclude(id__in=mappings_ids)
+        result[schema.name] = {
+            'id': schema.id,
+            'name': schema.definition['name'],
+        }
+        if other_linked_mappings:
```
obdulia-losantos: This condition can be changed to `result[schema.name]['is_unique'] = not other_linked_mappings` :nerd_face:

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -52,3 +56,125 @@ def merge_objects(source, target, direction):
...
+                ...
+            },
+            ...
+        }
+
+    }
+    '''
+
+    if mappingset:
+        mappings = models.Mapping.objects.filter(mappingset=mappingset).values_list('id', flat=True)
```
obdulia-losantos: Here you are overwriting the given `mappings` list, does this mean that or the mappingset or the mappings list but not both?

obdulia-losantos: also... the shortcut `mappingset.mappings.values_list('id', flat=True)` ;-)

then 
```python
mappings += mappingset.mappings.values_list('id', flat=True)
```

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -52,3 +56,125 @@ def merge_objects(source, target, direction):
...
+    result = {}
+    entities_opt = delete_opts.get('entities')
+    schemas_opt = delete_opts.get('schemas')
+    submissions_opt = delete_opts.get('submissions')
+    if entities_opt is True:
+        entities = models.Entity.objects.filter(mapping__id__in=mappings)
+        by_schemas = models.Schema.objects.filter(schemadecorators__entities__in=entities)\
+            .values('definition__name').annotate(count=Count('schemadecorators__entities'))
+        by_schemas_list = list(by_schemas.values('count', 'name'))
```
obdulia-losantos: why not directly 
```python
by_schemas_list = models.Schema.objects \
                               .filter(schemadecorators__entities__in=entities) \
                               .values('definition__name') \  # are you using this value???
                               .annotate(count=Count('schemadecorators__entities')) \
                               .values('count', 'name')
```

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -52,3 +56,125 @@ def merge_objects(source, target, direction):
...
+                try:
+                    schema_to_be_deleted.delete()
+                    schema_deletables[key]['is_deleted'] = True
+                except Exception as e:  # pragma: no cover
+                    schema_deletables[key]['is_deleted'] = False
+                    schema_deletables[key]['reason'] = str(e)
+        result['schemas'] = schema_deletables
+
+    if submissions_opt is True and mappingset:
+        submissions = models.Submission.objects.filter(mappingset=mappingset)
```
obdulia-losantos: the shortcut `mappingset.submissions`

aether-ui/aether/ui/api/utils.py
```diff
@@ -661,3 +662,23 @@ def kernel_data_request(url='', method='get', data=None, headers=None):
 
 def wrap_kernel_headers(instance):
     return mt_utils.add_instance_realm_in_headers(instance, kernel_utils.get_kernel_auth_header())
+
+
+def delete_operation(url, data, object):
```
obdulia-losantos: `object` is a sort of reserved word (like `int`) could you rename the variable? The typical `obj` works ;-)

aether-ui/aether/ui/api/utils.py
```diff
@@ -661,3 +662,23 @@ def kernel_data_request(url='', method='get', data=None, headers=None):
...
+    try:
+        response = kernel_data_request(
+            url=url,
+            method='post',
+            data=data,
+            headers=wrap_kernel_headers(object),
+        )
+        object.delete()
+        return Response(response)
```
obdulia-losantos: I don't like so much the idea of returning a `Response` object in a utility method, because this cannot be reused later in another context... why don't let the view method handle the response statuses...

```python
    try:
        response = kernel_data_request(
            url=url,
            method='post',
            data=data,
            headers=wrap_kernel_headers(obj),
        )
        obj.delete()
        return response

    except HTTPError as e:
        if e.response.status_code == status.HTTP_404_NOT_FOUND:
            obj.delete()
            return None
        raise e
```

In the view:

```python
    @action(detail=True, methods=['post'], url_path='delete-artefacts')
    def delete_artefacts(self, request, pk=None, *args, **kwargs):
        pipeline = self.get_object_or_404(pk=pk)
        
        artefacts_result = {'not_published': True}  # default response
        if pipeline.mappingset:
            try:
                artefacts_result = utils.delete_operation(
                    f'mappingsets/{pipeline.mappingset}/delete-artefacts/',
                    request.data,
                    pipeline,
                )
            except HTTPError as http_e:
                return Response(str(http_e), status=http_e.response.status_code)
            except Exception as e:
                return Response(str(e), status=status.HTTP_500_INTERNAL_SERVER_ERROR)
        else:
            pipeline.delete()
        
        return Response(artefacts_result)
```

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteStatus.jsx
```diff
@@ -0,0 +1,151 @@
...
+                  defaultMessage='Entity types deleted'
+                />
+              </label>
+              <div className='ml-5'>
+                {
+                  Object.keys(this.props.deleteStatus.schemas).map(schema => (
+                    <div key={schema}>
+                      <i className='fa fa-check mr-2' />
+                      <label>
+                        { `${schema} : ${this.props.deleteStatus.schemas[schema].is_deleted ? 'Deleted' : 'Not deleted, used by other mappings'}` }
```
obdulia-losantos:
```javascript
<label>
  { schema } :
  { 
    this.props.deleteStatus.schemas[schema].is_deleted
    ? <FormattedMessage
        id='delete.modal.entity.types.delete'
        defaultMessage='Deleted'
      />
    : <FormattedMessage
        id='delete.modal.entity.types.not-delete'
        defaultMessage='Not deleted, used by other mappings'
      />
  }
</label>
```

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -52,3 +56,125 @@ def merge_objects(source, target, direction):
...
     return result
+
+
+def get_unique_schemas_used(mappings_ids):
+    result = {}
+    schemas = models.Schema.objects.filter(schemadecorators__mappings__id__in=mappings_ids)
+    for schema in schemas:
+        other_linked_mappings = models.Mapping.objects.filter(
+            schemadecorators__schema__id=schema.id
+        ).exclude(id__in=mappings_ids)
```
obdulia-losantos, line 67: What happens when the schema is not used in any mapping but in the entities?

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -294,6 +324,126 @@ class Pipeline extends Component {
     )
   }
 
+  renderDeletionModal () {
```
obdulia-losantos, line 326: _one component to rule them all_ :sob: 

Would it be hard to refactor all of this to a new `Component`? :grimacing:

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -294,6 +324,126 @@ class Pipeline extends Component {
...
+          <label for='check3'>
+            <FormattedMessage
+              id='pipeline.delete.modal.all.data-0'
+              defaultMessage='Data'
+            />
+            <span className='bold'>
+              <FormattedMessage
+                id='pipeline.delete.modal.all.data-1'
+                defaultMessage='&nbsp;created by&nbsp;'
+              />
```
obdulia-losantos: The class `mx-2` introduce the space between the span tags.

```javascript
            <span className='bold mx-2'>
              <FormattedMessage
                id='pipeline.delete.modal.all.data-1'
                defaultMessage='created by'
              />
            </span>
```

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -294,6 +324,126 @@ class Pipeline extends Component {
...
+            }}
+          />
+          <label for='check3'>
+            <FormattedMessage
+              id='pipeline.delete.modal.all.data-0'
+              defaultMessage='Data'
+            />
+            <span className='bold'>
+              <FormattedMessage
+                id='pipeline.delete.modal.all.data-1'
```
obdulia-losantos: A meaningful id helps while translating the messages :books:

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -22,19 +22,210 @@ import React, { Component } from 'react'
...
+      showDeleteModal: false,
+      showDeleteProgress: true
+    })
+    this.props.deletePipeline(
+      this.props.pipeline.id,
+      this.state.deleteOptions
+    )
+  }
+
+  renderDeletionModal () {
```
obdulia-losantos, line 69: Wouldn't this be the same piece of code included in the `Pipeline` component? :see_no_evil:

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -22,19 +22,210 @@ import React, { Component } from 'react'
...
+          />
+          <label for='check1'>
+            <FormattedMessage
+              id='pipelineList.delete.modal.all.submissions-0'
+              defaultMessage='Data'
+            />
+            <span className='bold'>
+              <FormattedMessage
+                id='pipelineList.delete.modal.all.submissions-1'
+                defaultMessage='&nbsp;submitted to&nbsp;'
```
obdulia-losantos: Same comment as in the `Pipeline`...

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteStatus.jsx
```diff
@@ -0,0 +1,151 @@
...
+                  defaultMessage='Entities deleted'
+                />
+              </label>
+              <div className='ml-5'>
+                {
+                  this.props.deleteStatus.entities.schemas.map(schema => (
+                    <div key={schema.name}>
+                      <i className='fa fa-check mr-2' />
+                      <label>
+                        { `${schema.name} : ${schema.count}` }
```
obdulia-losantos: you don't need to use string templates for this 
```javascript
<label>
  { schema.name } : { schema.count }
</label>
```

aether-ui/aether/ui/assets/apps/pipeline/components/PipelineOptions.jsx
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (C) 2018 by eHealth Africa : http://www.eHealthAfrica.org
```
obdulia-losantos: Time flies and we are now in `2019` :scream:

scripts/docker_start.sh
```diff
@@ -125,7 +125,7 @@ case $app in
     ui)
         PRE_CONTAINERS=(ui-assets db nginx minio keycloak)
         SETUP_CONTAINERS=(kernel ui)
-        POST_CONTAINERS=()
+        POST_CONTAINERS=(ui)
```
obdulia-losantos: why?

aether-ui/aether/ui/assets/package.json
```diff
@@ -28,6 +28,7 @@
     "popper.js": "~1.15.0",
     "react": "~16.8.0",
     "react-dom": "~16.8.0",
+    "react-onclickoutside":"6.8.0",
```
obdulia-losantos: `~`

aether-kernel/aether/kernel/api/views.py
```diff
@@ -391,6 +420,40 @@ def skeleton(self, request, pk=None, *args, **kwargs):
...
+            # False if the schema is used by some other mapping
+            # e.g
+
+            Person : true
+
+            # indicating the schema 'Person' is used by one or more
+            # of the supplied mappings only
+        }
+        '''
+        mappings = request.data
```
obdulia-losantos: Being :smiling_imp: 

We need to check that the list of mapping ids belong to the current realm if multitenancy is enabled... :man_facepalming: 

```python
from django_eha_sdk.multitenancy.utils import filter_by_realm

mappings = filter_by_realm(self.request, 
                           model.Mapping.objects.filter(pk__in=request.data or []), 
                           'mappingset__project').values_list('id', flat=True)
```

aether-kernel/aether/kernel/api/utils.py
lordmallam: It means if a mappingset is provided, use only the mappings within that mappingset. Thus, overwrite the provided mappings if any

lordmallam: also about the shortcut, `mappingset` passed is a string not ofType `MappingSet`

aether-kernel/aether/kernel/api/utils.py
lordmallam: same here

aether-kernel/aether/kernel/api/utils.py
obdulia-losantos: It's up to you but maybe in this case will be helpful to rename the var to `mappingset_id` and `mapping_ids` or include a comment in the definition ;-)

I would come back with this question several times!!! :grimacing:

aether-kernel/aether/kernel/api/utils.py
obdulia-losantos: :woman_facepalming:

aether-kernel/aether/kernel/api/utils.py
lordmallam: The purpose of this for now is to identify delete-able schemas from the perspective of mappings. If it isn't used in a mapping, it will not be returned as the filter is on mappings

obdulia-losantos: Let me explain it better... what happens in this case:

1. Schema `S` is used but mappings `M` and `N`, and entities `E` (generated by `M`) and `F` (generated by `N`).
2. User `A` decides to delete mapping `N` (no more entities need to be produced with this mapping) but not the linked entities, so entity `F` is now _"mapping-orphan"_ but still belongs to `S`.
3. User `B` decides to check mapping `M` and the schemas linked to it.

With your approach schema `S` is only used by `M` (`result['S']['is_unique'] = True`) so user `B` decides to delete also the schema and the `schema.delete()` not only allows you to do that it also deletes entity `F` that user `A` wanted to keep.

lordmallam: Yes, I understand the issue. Which is why I brought up the suggestion sometime back about `Entity` not having cascade deletes at all. well except on projects

lordmallam: The question now, is should schema's be prevented from being deleted if entities outside the delete scope utilise them?

obdulia-losantos: Do we want to keep Entities without the schema that generated them? @adamvert 
Do we want orphan data? :scream:

admbtlr: According to the [spec](https://jira.ehealthafrica.org/browse/AET-476?focusedCommentId=36853&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-36853), deleting a Schema or Project Schema will _always_ delete the Entities

aether-kernel/aether/kernel/api/views.py
lordmallam: Why do we need to do that? 🤔

obdulia-losantos: Because if multi-tenancy is enabled you need to ensure that nobody is trying to get schemas form mappings that belong to different tenants.

Schemas are shared among all tenants but not the mappings/contracts.

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
lordmallam: 😓

aether-ui/aether/ui/api/views.py
```diff
@@ -53,16 +53,22 @@ def fetch(self, request, *args, **kwargs):
...
-                data,
-                pipeline
-            )
+            try:
+                artefacts_result = utils.delete_operation(
+                    f'mappingsets/{pipeline.mappingset}/delete-artefacts/',
+                    request.data,
+                    pipeline,
+                )
+                if not artefacts_result:
```
obdulia-losantos, line 64: this `if` can go outside, can't it? :thinking: 
```python
if not artefacts_result:
    return Response(status=status.HTTP_204_NO_CONTENT)
return Response(artefacts_result)
```

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -0,0 +1,138 @@
...
+              />
+              <span className='bold mx-2'>
+                <FormattedMessage
+                  id='delete.modal.all.submissions-1'
+                  defaultMessage='submitted to'
+                />
+              </span>
+              <FormattedMessage
+                id='delete.modal.all.submissions-2'
+                defaultMessage={`this ${this.props.objectType} (Submissions)`}
```
obdulia-losantos: https://github.com/formatjs/react-intl/wiki/Components#formattedmessage

```javascript
<FormattedMessage
    id='app.greeting'
    description='Greeting to welcome the user to the app'
    defaultMessage='Hello, {name}!'
    values={{
        name: <b>Eric</b>
    }}
/>
```

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteStatus.jsx
```diff
@@ -104,7 +104,18 @@ class DeleteStatus extends Component {
...
-                        { `${schema} : ${this.props.deleteStatus.schemas[schema].is_deleted ? 'Deleted' : 'Not deleted, used by other mappings'}` }
+                        { schema } :
+                        {
+                          this.props.deleteStatus.schemas[schema].is_deleted
+                            ? <FormattedMessage
+                              id='delete.modal.entity.types.delete'
+                              defaultMessage='Deleted'
+                            />
+                            : <FormattedMessage
+                              id='delete.modal.entity.types.not-delete'
```
obdulia-losantos: It's nice to see that when you write a typo it's well spread :joy: 
`deleted` and `not-deleted`

aether-ui/aether/ui/api/views.py
lordmallam: Yes it can, but I rather exit as soon as I can and avoid checking `if not artefacts_result` on error requests

obdulia-losantos: To be consistent maybe if the pipeline wasn't published yet the response should also be `204` :thinking:

```python
    @action(detail=True, methods=['post'], url_path='delete-artefacts')
    def delete_artefacts(self, request, pk=None, *args, **kwargs):
        pipeline = self.get_object_or_404(pk=pk)

        if pipeline.mappingset:
            try:
                artefacts_result = utils.delete_operation(
                    f'mappingsets/{pipeline.mappingset}/delete-artefacts/',
                    request.data,
                    pipeline,
                )
                if artefacts_result:
                    return Response(artefacts_result)
            except Exception as e:
                return Response(str(e), status=status.HTTP_500_INTERNAL_SERVER_ERROR)
        else:
            pipeline.delete()

        return Response(status=status.HTTP_204_NO_CONTENT)
```

obdulia-losantos: What do you think?

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -75,7 +75,7 @@ class DeleteModal extends Component {
               </span>
               <FormattedMessage
                 id='delete.modal.all.submissions-2'
-                defaultMessage={`this ${this.props.objectType} (Submissions)`}
+                defaultMessage='this pipeline (Submissions)'
```
obdulia-losantos: Could it be "contract"?

lordmallam: No, only pipelines/mappingsets have submissions

obdulia-losantos: Right!

aether-ui/aether/ui/api/views.py
lordmallam: The implementation is to differentiate between an unpublished pipeline and a published pipeline with a missing upstream mappingset

obdulia-losantos: would it be better in that case to return a "missing artefacts" warning? (up to you) it's fine as it is.

aether-kernel/aether/kernel/api/utils.py
```diff
@@ -52,3 +56,123 @@ def merge_objects(source, target, direction):
...
+    result = {}
+    entities_opt = delete_opts.get('entities')
+    schemas_opt = delete_opts.get('schemas')
+    submissions_opt = delete_opts.get('submissions')
+    if entities_opt is True:
+        entities = models.Entity.objects.filter(mapping__id__in=mapping_ids)
+        by_schemas_list = list(
+            models.Schema.objects.filter(schemadecorators__entities__in=entities)
+            .values('name').annotate(count=Count('schemadecorators__entities')).values('count', 'name')
+        )
```
obdulia-losantos, line 155: could you change this into something easier to read? :grimacing:

aether-kernel/aether/kernel/api/utils.py
obdulia-losantos: I don't see that this is resolved. :(

aether-kernel/aether/kernel/api/utils.py
obdulia-losantos: Thanks!

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -0,0 +1,158 @@
...
+      submissions: false
+    }
+  }
+
+  render () {
+    const header = (
+      <span>
+        <FormattedMessage
+          id='delete.modal.header'
+          defaultMessage={this.props.header}
```
obdulia-losantos: ????

aether-ui/aether/ui/assets/apps/pipeline/components/PipelineOptions.jsx
```diff
@@ -0,0 +1,68 @@
...
+          onClick={this.toggle.bind(this)}>
+          <span className='details-title'>
+            <i className='fas fa-ellipsis-h' />
+          </span>
+        </button>
+        {
+          this.state.show && <ul
+            className='options'>
+            <li onClick={() => this.props.rename()}>Rename Pipeline</li>
+            <li onClick={() => this.props.delete()}>Delete Pipeline</li>
```
obdulia-losantos: :de: :gb: :es: :nigeria:

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -26,15 +26,78 @@ import { ModalError, NavBar } from '../components'
...
+  }
+
+  renderDeletionModal () {
+    if (!this.state.showDeleteModal) {
+      return null
+    }
+
+    return (
+      <DeleteModal
+        header='Delete pipeline '
```
obdulia-losantos: This is the text that must contain the format message
```
<DeleteModal
  header={
    <FormattedMessage id='pipeline.list.delete.modal.header' defaultMessage='Delete pipeline' />
  }
...
/>
```

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -26,15 +26,78 @@ import { ModalError, NavBar } from '../components'
...
+    )
+  }
+
+  renderDeleteProgressModal () {
+    if (!this.state.showDeleteProgress) {
+      return null
+    }
+    return (
+      <DeleteStatus
+        header='Deleting pipeline '
```
obdulia-losantos: Same comment
```
<DeleteSataus
  header={
    <FormattedMessage id='pipeline.list.delete.status.header' defaultMessage='Deleting pipeline' />
  }
...
/>
```

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -294,6 +323,40 @@ class Pipeline extends Component {
...
   }
 
+  renderDeletionModal () {
+    if (!this.state.showDeleteModal) {
+      return null
+    }
+
+    return (
+      <DeleteModal
+        header='Delete contract '
```
obdulia-losantos: Same comment

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -294,6 +323,40 @@ class Pipeline extends Component {
...
+    )
+  }
+
+  renderDeleteProgressModal () {
+    if (!this.state.showDeleteProgress) {
+      return null
+    }
+    return (
+      <DeleteStatus
+        header='Deleting contract '
```
obdulia-losantos: same comment

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -33,11 +33,16 @@ class DeleteModal extends Component {
   }
 
   render () {
+    const messages = {
+      header: this.props.header || 'Delete ',
+      objectType: this.props.objectType || 'pipeline'
+    }
```
obdulia-losantos:
```javascript
import { defineMessages, injectIntl } from 'react-intl'

const MESSAGES = defineMessages({
  pipeline: { id: 'pipeline', defaultMessage: 'pipeline' },
  contract: { id: 'contract', defaultMessage: 'contract' }
})

class DeleteModal extends Component {
  render () {
    const { formatMessage } = this.props.intl
    const objectType = formatMessage(this.props.objectType || 'pipeline')

    const header = (
      <span>
        <FormattedMessage
          id='delete.modal.header'
          defaultMessage='Delete {type} <b>{name}</b>?'
          values={{
            type: objectType,
            name: this.props.obj.name
          }}
        />
      </span>
    )

    // ...

        <label for='check3'>
          <FormattedMessage
            id='delete.modal.entities'
            defaultMessage='Data <b>created by</b> this {type} (Entities)'
            values={{ type: objectType }}
          />
        </label>
    
    // ....
  }
}

export default injectIntl(DeleteModal)
```

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -88,19 +88,12 @@ class DeleteModal extends Component {
...
-                  id='delete.modal.all.submissions-1'
-                  defaultMessage='submitted to'
-                />
-              </span>
               <FormattedMessage
                 id='delete.modal.all.submissions-2'
-                defaultMessage='this pipeline (Submissions)'
+                defaultMessage='Data {emphasis} this pipeline (Submissions)'
+                values={{
+                  emphasis: <b>sbumitted to</b>
```
obdulia-losantos: ytpo ;-)

But in this case `submitted to` is not "eligible" for translation.

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -141,18 +134,11 @@ class DeleteModal extends Component {
...
-              />
-            </span>
-            <FormattedMessage
-              id='delete.modal.all.data-2'
-              defaultMessage='this {objectType} (Entities)'
-              values={messages}
+              defaultMessage='Data {emphasis} this {objectType} (Entities)'
+              values={{
+                ...messages,
+                emphasis: <b>created by</b>
```
obdulia-losantos: same comment as above

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -41,7 +41,7 @@ class DeleteModal extends Component {
       <span>
         <FormattedMessage
           id='delete.modal.header'
-          defaultMessage='{header}'
+          defaultMessage='{header} {objectType} '
```
obdulia-losantos: If we run the extract "translatable" texts script here (the one that we didn't include yet) the extracted text would be `"{header} {objectType} "`...

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -19,9 +19,28 @@
...
 
 import React, { Component } from 'react'
-import { FormattedMessage } from 'react-intl'
+import { FormattedMessage, FormattedHTMLMessage, injectIntl, defineMessages } from 'react-intl'
 import { Modal } from '../../components'
 
+const MESSAGES = defineMessages({
+  delete: {
+    id: 'modal.delete.text',
+    defaultMessage: 'Delete {obj} <b>{objName}</b>'
```
obdulia-losantos, line 28: :+1:

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -19,9 +19,28 @@
...
+  },
+  submissions: {
+    id: 'modal.delete.submissions.text',
+    defaultMessage: 'Data <b>submitted to</b> this pipeline (Submissions)'
```
obdulia-losantos, line 32: :+1:

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -19,9 +19,28 @@
...
+  },
+  entities: {
+    id: 'modal.delete.entities.text',
+    defaultMessage: 'Data <b>created by</b> this {obj} (Entities)'
```
obdulia-losantos, line 36: :+1:

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -19,9 +19,28 @@
...
+  },
+  objectType: {
+    id: 'modal.delete.object.type',
+    defaultMessage: '{objType}'
```
obdulia-losantos: :disappointed:

obdulia-losantos:
```javascript
const MESSAGES = defineMessages({
  // ...
  pipeline: {
    id: 'modal.delete.object.type.pipeline',
    defaultMessage: 'pipeline'
  },
  contract: {
    id: 'modal.delete.object.type.contract',
    defaultMessage: 'contract'
  }
})
```

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -33,18 +52,19 @@ class DeleteModal extends Component {
...
       <span>
-        <FormattedMessage
-          id='delete.modal.header'
-          defaultMessage='{header} {objectType} '
-          values={messages}
+        <FormattedHTMLMessage
+          {...
+          { ...MESSAGES.delete,
+            values: {
+              obj: formatMessage(MESSAGES.objectType, { objType: this.props.objectType }),
```
obdulia-losantos:
```javascript
 obj: formatMessage(MESSAGES[this.props.objectType])
```

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -33,18 +52,19 @@ class DeleteModal extends Component {
...
+    const { formatMessage } = this.props.intl
     const header = (
       <span>
-        <FormattedMessage
-          id='delete.modal.header'
-          defaultMessage='{header} {objectType} '
-          values={messages}
+        <FormattedHTMLMessage
+          {...
+          { ...MESSAGES.delete,
```
obdulia-losantos, line 65: You could have defined the HTML message here...

```javascript
<FormattedHTMLMessage
  id='modal.delete.text'
  defaultMessage='Delete {obj} <b>{objName}</b>'
  values={{
    obj: formatMessage(MESSAGES[this.props.objectType]),
    objName: this.props.obj.name
  }}
/>
```

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -88,12 +108,10 @@ class DeleteModal extends Component {
...
-              <FormattedMessage
-                id='delete.modal.all.submissions-2'
-                defaultMessage='Data {emphasis} this pipeline (Submissions)'
-                values={{
-                  emphasis: <b>sbumitted to</b>
-                }}
+              <FormattedHTMLMessage
+                {...
+                { ...MESSAGES.submissions }
+                }
```
obdulia-losantos, line 119: This was fine except for the `emphasis`

```javascript
<FormattedHTMLMessage
  id='delete.modal.all.submissions-2'
  defaultMessage='Data <b>sbumitted to</b> this pipeline (Submissions)'
/>

File: aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx [hunk 46]
```diff
@@ -132,13 +150,14 @@ class DeleteModal extends Component {
...
-                emphasis: <b>created by</b>
-              }}
+            <FormattedHTMLMessage
+              {...
+              { ...MESSAGES.entities,
+                values: {
+                  obj: formatMessage(MESSAGES.objectType, { objType: this.props.objectType })
+                }
+              }
+              }
```
obdulia-losantos, line 165: Same comment 
```javascript
<FormattedHTMLMessage
  id='modal.delete.entities.text'
  defaultMessage='Data <b>created by</b> this {obj} (Entities)'
  values={{ obj: formatMessage(MESSAGES[this.props.objectType]) }}
/>
```

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
obdulia-losantos: @lordmallam :grimacing:

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
obdulia-losantos: @lordmallam :see_no_evil:

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteStatus.jsx
```diff
@@ -0,0 +1,182 @@
...
+  render () {
+    if (!this.props.showModal) {
+      return null
+    }
+
+    const header = (
+      <span>
+        <FormattedMessage
+          id='delete.progress.modal.header'
+          defaultMessage={this.props.header}
```
obdulia-losantos: :disappointed:

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
lordmallam: With this you get  an error `Uncaught Invariant Violation: [React Intl] An `id` must be provided to format a message.`

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
lordmallam: Decided to keep all template related messages together

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
obdulia-losantos: let me check...

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
obdulia-losantos: in that case (to reduce complexity)...

```
const { formatMessage, formatHTMLMessage } = this.props.intl

// ...

{ formatHTMLMessage(MESSAGE.entities, { obj: formatMessage(MESSAGES[this.props.objectType]) }) }
```

lordmallam: Which is what is implemented. using `FormattedHTMLMessage` tag instead

obdulia-losantos:
I did not create the contract yet but I can delete it... :wink: 

image

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
lordmallam: I rather have it as a template with interchangeable variables than having constants, except if there's a side effect am not seeing?

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
obdulia-losantos: image


```javascript

const MESSAGES = defineMessages({
  delete: {
    id: 'modal.delete.text',
    defaultMessage: 'Delete {obj} <b>{objName}</b>'
  },
  submissions: {
    id: 'modal.delete.submissions.text',
    defaultMessage: 'Data <b>submitted to</b> this pipeline (Submissions)'
  },
  entities: {
    id: 'modal.delete.entities.text',
    defaultMessage: 'Data <b>created by</b> this {obj} (Entities)'
  },
  pipeline: {
    id: 'modal.delete.object.type.pipeline',
    defaultMessage: 'pipeline'
  },
  contract: {
    id: 'modal.delete.object.type.contract',
    defaultMessage: 'contract'
  }
})

class DeleteModal extends Component {
  // ...

  render () {
    const { formatMessage } = this.props.intl
    const objType = formatMessage(MESSAGES[this.props.objectType || 'pipeline'])
    const header = (
      <span>
        <FormattedHTMLMessage {...
          { ...MESSAGES.delete,
            values: { obj: objType, objName: this.props.obj.name }
          }}
        />
      </span>
    )

  // ...

            <label for='check1'>
              <FormattedHTMLMessage {... { ...MESSAGES.submissions }} />
            </label>

  // ...

          <label for='check3'>
            <FormattedHTMLMessage {... { ...MESSAGES.entities, values: { obj: objType } }} />
          </label>

  // ...
  }
}
export default injectIntl(DeleteModal)
```

obdulia-losantos:
The expected answer :joy: 
image

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
obdulia-losantos: The extracted text would be "{my_variable}" but never "contract" or "pipeline", the purpose of all of this is to make those texts "eligible" for translation... otherwise this would be much simpler. :books:

aether-ui/aether/ui/assets/apps/pipeline/Pipeline.jsx
```diff
@@ -344,7 +344,9 @@ class Pipeline extends Component {
     }
     return (
       <DeleteStatus
-        header='Deleting contract '
+        header={
+          <FormattedMessage id='contract.delete.status.header' defaultMessage='Deleting contract ' />
+        }
```
obdulia-losantos, line 349: :+1:

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -87,7 +87,9 @@ class PipelineList extends Component {
     }
     return (
       <DeleteStatus
-        header='Deleting pipeline '
+        header={
+          <FormattedMessage id='pipeline.list.delete.status.header' defaultMessage='Deleting pipeline ' />
+        }
```
obdulia-losantos, line 92: :+1:

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteStatus.jsx
obdulia-losantos: In this case, this is useless and creates more confusion than me!!! :stuck_out_tongue: 
The simple `const header = this.props.header` is enough. You already translated it. :gb: :nigeria: :de:

lordmallam: Ahhhh... yes, true. The back and forth is throwing me off...  😄

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
```diff
@@ -0,0 +1,174 @@
...
+          this.props.objectType === 'pipeline' &&
+          <div className='check-default ml-4'>
+            <input type='checkbox' id='check1' checked={this.state.submissions}
+              onChange={(e) => {
+                this.setState({
+                  ...this.state, submissions: e.target.checked
+                })
+              }}
+            />
+            <label for='check1'>
```
obdulia-losantos: :disappointed_relieved: 

```text
Warning: Invalid DOM property `for`. Did you mean `htmlFor`?
    in label (created by DeleteModal)
```

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
```diff
@@ -59,19 +123,24 @@ class PipelineList extends Component {
                 key={pipeline.id}
                 pipeline={pipeline}
                 history={this.props.history}
+                rename={() => {}}
```
obdulia-losantos: :roll_eyes:

obdulia-losantos: why not `rename={() => { window.alert('we are working hard on this... or maybe we forgot to do it') }}`? :wink:

aether-ui/aether/ui/assets/apps/pipeline/components/DeleteModal.jsx
obdulia-losantos: if you fix this issue (3 times) I think we are done with this :scream:

aether-ui/aether/ui/assets/apps/pipeline/PipelineList.jsx
obdulia-losantos: Actually, I would hide this till it's  ready...

aether-kernel/aether/kernel/api/tests/test_utils.py
```diff
@@ -16,14 +16,95 @@
...
+            ],
+        }
+        self.project_artefacts = self.client.patch(
+            url,
+            data=data,
+            content_type='application/json',
+        ).json()
+
+    def tearDown(self):
+        self.project.delete()
```
obdulia-losantos, line 106: this happens automatically but :speak_no_evil:

aether-kernel/aether/kernel/api/utils.py
obdulia-losantos: I would love to have at least a comment here with the risks of flagging a schema with entities but not mappings as "unique"

obdulia-losantos (approved):
The only thing that bothers me is the issue with the schema "uniqueness" but I'm not going to block this PR any longer with it... :bomb:

PR #570, opened 2019-05-09
chore: (kernel) submission cascade delete

lordmallam:
Remove cascade delete on mappingset ==> submissions


aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -203,7 +203,7 @@ class AttachmentSerializer(DynamicFieldsMixin, serializers.ModelSerializer):
 
     submission = MtPrimaryKeyRelatedField(
         queryset=models.Submission.objects.all(),
-        mt_field='mappingset__project',
+        mt_field='project',
```
obdulia-losantos, line 210: Also in the `class SubmissionSerializer` the `mappingset` field is no longer required ;-)

obdulia-losantos: But I would say that **IT IS REQUIRED TO CREATE THE SUBMISSION** or no entities will be extracted...

obdulia-losantos:
Another point, what happens with the entity extraction if the submission has no mappingset? :scream:

lordmallam:
@obdulia-losantos a mappingset is required on submission create only

obdulia-losantos:
Have you tried to update a submission without mapping set?

aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -168,13 +168,11 @@ class SubmissionSerializer(DynamicFieldsMixin, serializers.ModelSerializer):
         queryset=models.Project.objects.all(),
         required=False,
     )
-    mappingset = MtPrimaryKeyRelatedField(
```
obdulia-losantos: You cannot remove this or multitenancy will not work properly :disappointed: you only need to add the option `required=False`

aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -168,13 +168,11 @@ class SubmissionSerializer(DynamicFieldsMixin, serializers.ModelSerializer):
...
-        queryset=models.MappingSet.objects.all(),
-        mt_field='project',
-    )
 
     def create(self, validated_data):
         instance = super(SubmissionSerializer, self).create(validated_data)
+        if instance.mappingset is None:
```
obdulia-losantos: The other way around, check the value before create it
```python
        if validated_data.get('mappingset') is None:
            raise serializers.ValidationError('Mappingset must be provided on initial submission')
        instance = super(SubmissionSerializer, self).create(validated_data)
```

obdulia-losantos (approved):
:see_no_evil:

PR #575, opened 2019-05-15
feat: (kernel) adds topic in schemadecorator

lordmallam:
Adds `topic_name` in `SchemaDecorator` as a JSONField defaulting to `{'name': [schemadeocrator.name]}`

Related JIRA Ticket: [AET-386](https://jira.ehealthafrica.org/browse/AET-386)


aether-kernel/aether/kernel/api/models.py
```diff
@@ -414,6 +415,7 @@ class SchemaDecorator(ExportModelOperationsMixin('kernel_schemadecorator'), Proj
     transport_rule = models.TextField(null=True, blank=True, verbose_name=_('transport rule'))
     masked_fields = models.TextField(null=True, blank=True, verbose_name=_('masked fields'))
     is_encrypted = models.BooleanField(default=False, verbose_name=_('encrypted?'))
+    topic_name = JSONField(null=True, blank=True, verbose_name=_('topic name'))
```
obdulia-losantos: I think that the field name should be just `topic`, shouldn't it?

obdulia-losantos: Does it need to be a JSON field, could it be a simple TextField?

obdulia-losantos: @shawnsarwar ?

aether-kernel/aether/kernel/api/filters.py
```diff
@@ -165,7 +165,7 @@ def mapping_filter(self, queryset, name, value):
             return queryset.filter(mappings__name=value)
 
     class Meta:
-        fields = '__all__'
+        exclude = ('topic',)
```
obdulia-losantos, line 168: Do we need to exclude the topic filed in filters? If we need it to be used in the producer it looks like that we need to filter the entries by this field...

obdulia-losantos: :woman_shrugging:

aether-kernel/aether/kernel/api/migrations/0104_schemadecorator_topic_name.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 2.2.1 on 2019-05-15 09:57
+
+import django.contrib.postgres.fields.jsonb
+from django.db import migrations
+
+def migrate__update_schemadecorator(apps, schema_editor):
+    SchemaDecorator = apps.get_model('kernel', 'SchemaDecorator')
+    for sd in SchemaDecorator.objects.all():
+        sd.save() # default topic_name to { name: [schemadecorator.name] }
```
obdulia-losantos: flake8 warning... at least 2 spaces between the piece of code and the inline comment ;-)

aether-kernel/aether/kernel/api/models.py
```diff
@@ -414,6 +415,7 @@ class SchemaDecorator(ExportModelOperationsMixin('kernel_schemadecorator'), Proj
     transport_rule = models.TextField(null=True, blank=True, verbose_name=_('transport rule'))
     masked_fields = models.TextField(null=True, blank=True, verbose_name=_('masked fields'))
     is_encrypted = models.BooleanField(default=False, verbose_name=_('encrypted?'))
+    topic = JSONField(null=True, blank=True, verbose_name=_('topic name'))
```
obdulia-losantos: `verbose_name='topic'`

aether-kernel/aether/kernel/api/tests/test_models.py
```diff
@@ -391,3 +391,18 @@ def test_model_deletion(self):
...
+        schema = models.Schema.objects.create(
+            name='schema',
+            definition=EXAMPLE_SCHEMA,
+        )
+        schemadecorator = models.SchemaDecorator.objects.create(
+            name='schema decorator',
+            project=project,
+            schema=schema,
+        )
+        self.assertEqual(schemadecorator.topic, {'name': 'schema decorator'})
```
obdulia-losantos, line 408: Could you add a couple of silly test in which you change the schemadecorator name and the topic doesn't change and the changing the topic one?

aether-kernel/aether/kernel/api/filters.py
shawnsarwar: The producer is going to use a single query that doesn't take topic into account. Don't need the filter for that specific purpose.

aether-kernel/aether/kernel/api/models.py
shawnsarwar: Let's use a JSONField in case we want to put more advanced production options here.

obdulia-losantos (approved):

PR #576, opened 2019-05-15
feat: aether client oidc

shawnsarwar:


aether-client-library/aether/client/basic_auth.py
```diff
@@ -0,0 +1,62 @@
+# Copyright (C) 2018 by eHealth Africa : http://www.eHealthAfrica.org
```
obdulia-losantos: Now it's `2019` :grimacing:

aether-client-library/aether/client/oidc.py
```diff
@@ -0,0 +1,256 @@
...
+    def bind_client(self, client):
+        client.session = self.session
+        client.authenticator = self
+
+    def get_spec(self, spec_url):
+        res = self.session.get(spec_url)
+        self.csrf = res.cookies.get('csrftoken')
+        LOG.debug(f'Set CSRFToken: {self.csrf}')
+        spec = res.json()
+        spec['host'] = f'{spec["host"]}/{self.realm}/kernel'
```
obdulia-losantos: I would never hard-code `kernel`... we are using the `GATEWAY_ID` )or something like that env var.

shawnsarwar: I stuck it into the `Client` constructor as an optional kwarg.

obdulia-losantos (approved):

PR #577, opened 2019-05-15
feat: simple producer mt

shawnsarwar:
It became apparent when working on the client MT, that the integration tests wouldn't work until we had some sort of producer multi-tenancy in place.  This patch allows topics to be created on a {realm}:{topic_name} basis. It also removes the single-tenant aether-client and moves us temporarily to pulling schemas directly from the database via the `readonlyuser`.

I'm hesitant to call this the MT producer because it won't scale well, primarily because each topic uses it's own greenlet, which will get slow eventually.  The real producer MT work is on feat/multi-tenant-producer but can wait until the optimization work in 1.6. This patch should be sufficient for 1.5, although I hate how few tests it has.


shawnsarwar:
The patch for the gateway capable client will follow, but requires these changes.

aether-kernel/sql/create_readonly_user.py
```diff
@@ -59,6 +59,7 @@
 GRANT SELECT ON kernel_mapping TO {role_id};
 GRANT SELECT ON kernel_schemadecorator TO {role_id};
 GRANT SELECT ON kernel_schema TO {role_id};
+GRANT SELECT ON multitenancy_mtinstance TO {role_id};
```
obdulia-losantos, line 77: The problem with this is that if multi-tenancy is not enabled, this table does not exists at all :disappointed:

shawnsarwar: Oh no. Looks like I need a mode switch for the producer. How can we make the readonlyuser creation reactive to the system config?

obdulia-losantos: Ask to the _DB expert_ :smiling_imp:

obdulia-losantos: If the table does not exits... create a view to simulate it  _expert says_

```sql
GRANT SELECT ON kernel_project TO {role_id};

DROP VIEW IF EXISTS multitenancy_mtinstance;
DO $$
BEGIN
  IF NOT EXISTS (SELECT FROM information_schema.tables WHERE table_name = 'multitenancy_mtinstance')
  THEN
      CREATE VIEW multitenancy_mtinstance AS
            SELECT '-'    AS realm,
                    p.id  AS instance_id
            FROM kernel_project AS p;
  END IF;
END
$$ LANGUAGE plpgsql;
```

aether-producer/conf/pip/primary-requirements.txt
```diff
@@ -11,15 +11,15 @@
...
 confluent_kafka
 flask
 gevent
 psycogreen
 psycopg2-binary
+requests
 spavro
 sqlalchemy
-urllib3>=1.25  # due to security warning
+urllib3  # due to security warning
```
obdulia-losantos: this is not longer necessary, `request` already upgraded its dependencies

obdulia-losantos:
How do you provide the `realm` value to the producer? It looks like it's somewhere in the schema definition but I cannot find where...

obdulia-losantos:
Never mind, I found it  :woman_facepalming:

shawnsarwar:
I'm still fixing integration tests in Travis, apparently.

shawnsarwar:
@obdulia-losantos  :tada:

obdulia-losantos (approved):

PR #581, opened 2019-05-20
fix: (ui) navigation

lordmallam:
Reverting https://github.com/eHealthAfrica/aether/pull/578/files#diff-a8f8351a04f88c434814614fa553b51bL42
and 
https://github.com/eHealthAfrica/aether/pull/578/files#diff-0d8a45cd945856f9b8dd65318f43cb0dR173


aether-ui/aether/ui/assets/apps/components/NavBar.jsx
```diff
@@ -39,7 +39,7 @@ class NavBar extends Component {
 
     return (
       <div data-qa='navbar' className='navbar top-nav'>
-        <Link className='top-nav-logo' title='aether' to='#'>
+        <Link className='top-nav-logo' title='aether' onClick={this.props.cb}>
```
obdulia-losantos: `to` is mandatory.... we need to find a way to solve this point :disappointed: 

https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md

obdulia-losantos (approved):

PR #583, opened 2019-05-23
fix(sync): start RQ worker along with webserver

obdulia-losantos:
Remove one useless container :dancer:


shawnsarwar (approved):
:trophy:

PR #596, opened 2019-06-05
feat: aether.sdk library

obdulia-losantos:


admbtlr (changes requested):
I think it would be better to leave the Aether README intact, rather than linking to aether.sdk - this is the place where people will look for that information about 90% of the time

obdulia-losantos:
Reverted!

admbtlr (approved):
Thank you!

PR #598, opened 2019-06-06
feat: (ui) pipeline rename

lordmallam:
Allows renaming of pipelines (mappingset) from the UI and propagates changes to kernel if published.

Related JIRA ticket: [AET-518](https://jira.ehealthafrica.org/browse/AET-518)

@sabeen we could use some better styling 😄 
Screenshot 2019-06-07 at 13 23 58


aether-ui/aether/ui/api/views.py
```diff
@@ -70,6 +71,33 @@ def delete_artefacts(self, request, pk=None, *args, **kwargs):
 
         return Response(artefacts_result)
 
+    @action(detail=True, methods=['put'], url_path='rename')
+    def rename(self, request, pk=None, *args, **kwargs):
+        pipeline = self.get_object_or_404(pk=pk)
+        new_name = request.data.get('name', pipeline.name)
+        pipeline.name = new_name
+        pipeline.save()
+        if pipeline.mappingset:
```
obdulia-losantos, line 79: Do we need to update the kernel "mappingset" in this method? :thinking:

aether-ui/aether/ui/api/views.py
```diff
@@ -70,6 +71,33 @@ def delete_artefacts(self, request, pk=None, *args, **kwargs):
 
         return Response(artefacts_result)
 
+    @action(detail=True, methods=['put'], url_path='rename')
+    def rename(self, request, pk=None, *args, **kwargs):
```
obdulia-losantos, line 75: The partial update method AKA `PATCH` didn't work for you? :see_no_evil:

aether-ui/aether/ui/api/views.py
lordmallam: Yes, I agreed with @adamvert to propagate to kernel immediately if the pipeline is already published.

aether-ui/aether/ui/assets/apps/pipeline/components/PipelineCard.jsx
```diff
@@ -44,12 +67,23 @@ class PipelineCard extends Component {
           <span className='pipeline-name'>
             // { pipeline.name }
           </span>
-          <div className='pipeline-actions'>
-            <PipelineOptions
-              delete={this.props.delete}
-            />
-            <ContractAddButton pipeline={pipeline} history={this.props.history} />
-          </div>
+          <PipelineRename
```
admbtlr: The logic here is a little weird: 
- if `state.renameView === 'form'`, `PipelineRename` will render a form, and the `pipeline-actions` div will not be rendered
- else `PipelineRename` will render `null`, and the `pipeline-actions` div _will_ be rendered
- also, `renameView` has two possible values, which actually makes it a good candidate to simplify into a boolean, e.g. `isRenaming`

It would be nice to make this a bit less spaghetti-ish, so that this render method looks more like:

```
{ this.state.isRenaming ?
  <PipelineRename /> :
  <PipelineActions /> }
```

That way, `PipelineRename` doesn't need the `view` prop.

admbtlr: (Where `PipelineOptions` could be a very simple little functional component that just needs a couple of functions passed in as props)

admbtlr (changes requested):

aether-ui/aether/ui/assets/apps/pipeline/components/PipelineCard.jsx
lordmallam: Yeah. the hide on `pipeline-actions` was from the 1st implementation, now since a modal is been used, hiding the actions is no longer necessary. Refactoring

admbtlr: Thanks @lordmallam , this is better - but do you think it would be possible to implement it like I outlined above, so that this is moved out to a simple functional component:
```
          <div className='pipeline-actions'>
            <PipelineOptions
              delete={this.props.delete}
              rename={this.onRename}
            />
            <ContractAddButton pipeline={pipeline} history={this.props.history} />
          </div>
```
called, for example, `PipelineActions`. Then the `PipelineCard` render function can become:
```
{ this.state.isRenaming ?
  <PipelineRename /> :
  <PipelineActions /> }
```
which is a lot more readable. This means you can also remove the `show` prop from `PipelineRename`, which will mean that you can easily get rid of its `renderForm` method.

admbtlr (approved):
💥

PR #602, opened 2019-06-14
feat: added xformType in AVRO schema def

obdulia-losantos:
Related to: https://jira.ehealthafrica.org/browse/AET-520

I don't have time before holidays to add a proper visualization field with the expected widget but at least the original field type is included there and allows to handle the data in kibana or any other graphical tool better than before.


obdulia-losantos:
Continued in #606

PR #603, opened 2019-06-17
fix: ODK formlist port numbers

shawnsarwar:
We were having an issue in Aether bootstrap where the server isn't using SSL. The odk module would try to serve forms over HTTP via port 80 instead of 8443 which is what ODK collect expects. This patch adds 8443 to the urls inside of XML payloads when schema is not https and the port is not already set.

### DIFF
--- a/aether-odk-module/aether/odk/api/views_collect.py
+++ b/aether-odk-module/aether/odk/api/views_collect.py
@@ -147,6 +147,8 @@ def xform_list(request, *args, **kwargs):
         xforms = xforms.filter(form_id=formID)
 
     host = request.build_absolute_uri(request.path).replace(reverse('xform-list-xml'), '')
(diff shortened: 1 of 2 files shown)


aether-odk-module/aether/odk/api/views_collect.py
```diff
@@ -146,7 +147,9 @@ def xform_list(request, *args, **kwargs):
         xforms = xforms.filter(form_id=formID)
 
     host = request.build_absolute_uri(request.path).replace(reverse('xform-list-xml'), '')
-
+    url_info = urlparse(host)
```
admbtlr, line 152: Maybe add a little comment that says something `ODK Collect will only use either https or port 8443` to explain why this is here...?

admbtlr (approved):

PR #604, opened 2019-06-18
feat: producer sasl

shawnsarwar:
Added SASL auth to the producer to align with pattern used in bootstrap / deployment. Tested with bootstrap. Fine tuned producer settings for better performance.


admbtlr (approved):
💥

PR #605, opened 2019-06-21
feat: (kernel) return mapping topics

lordmallam:
implements an endpoint to return all `schemadecorator` topics used within a `mapping`
#### Usage
GET `/mappings/{id}/topics`


shawnsarwar (approved):
:+1:

PR #608, opened 2019-07-01
fix(odk): check that the port is not 8443

obdulia-losantos:
Checking the python docs: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

The URL could be `http://hostname/formList` or  `http://hostname:80/formList` in the second case the built URLs did not contain the port `8443`.


shawnsarwar (approved):
Assuming this doesn't break when running behind the gateway, LGTM

PR #609, opened 2019-07-02
feat(odk): add geoshape, geotrace, select, rank

obdulia-losantos:
Added support for the following XForm field types:

- geotrace
- geoshape
- select: changed the values parser, from a string into a list of strings.
- rank (same as select)

image


admbtlr:
Looks good!

@djmoran64 this PR will change the format of multiselect answers from a space separated String `"Orange Apple Plum"` to an array `["Orange", "Apple", "Plum"]`. Is that a good solution?

shawnsarwar (dismissed):
:sun_with_face:

admbtlr (dismissed):

admbtlr (approved):

PR #610, opened 2019-07-03
fix(odk): XMLForm without h namespace

obdulia-losantos:
Closes: https://jira.ehealthafrica.org/browse/AET-526

Note: this fix only applies for **h** (http://www.w3.org/1999/xhtml) namespace.

The rest of expected namespaces are:

- **ev** http://www.w3.org/2001/xml-events
- **jr** http://openrosa.org/javarosa
- **odk** http://www.opendatakit.org/xforms
- **orx** http://openrosa.org/xforms
- **xsd** http://www.w3.org/2001/XMLSchema

### DIFF
--- a/aether-odk-module/aether/odk/api/tests/test_xform_utils.py
+++ b/aether-odk-module/aether/odk/api/tests/test_xform_utils.py
@@ -272,6 +272,50 @@ class XFormUtilsParsersTests(CustomTestCase):
         }
         self.assertEqual(parse_xml_to_dict(xml_str), expected)
 
+    def test__parse_xml_to_dict__missing_h_namespace(self):
+        self.assertEqual(parse_xml_to_dict('<html/>'),
+                         {'h:html': None})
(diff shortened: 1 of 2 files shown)


admbtlr:
After examining the documentation, I'm pretty sure that `<html>` etc is actually invalid. So I'm surprised that this form works in ODK Collect, and I'm no longer sure whether we should actually support it. 🤔

obdulia-losantos:
The problem with the parser is that sometimes they remove the namespace making `h:html` and `html` being equal. Actually, this happens in the `pyxform` library and that's why I made my own parser to check the conditions... and more reasons. :woman_shrugging:

admbtlr:
I guess that makes some kind of sense, as long as you check that removing the namespace won't leave you with clashes.

obdulia-losantos:
This workaround allows the use of of xforms that do not conform current XForm specifications.

PR #619, opened 2019-07-10
fix(odk): gathering choice list of one item

obdulia-losantos:
This fixes the issue with the HS.io xForm. There is one choice list with only ONE item.

list_name | name | label::English (en)
-- | -- | -- |
hc_speciality_list | xx | xx |


admbtlr (approved):
That's a pretty annoying issue...

PR #626, opened 2019-07-19
fix(odk): use hostname instead of netloc

obdulia-losantos:
`netloc` contains the `hostname` along with the current `port`; if the initial `host` value were `http://my-server:8080/odk` we were returning `http://my-server:8080:8443/odk` and not `http://my-server:8443/odk`


shawnsarwar (approved):
Doh

PR #628, opened 2019-07-19
fix(kernel): SubmissionStatsMixin [entities_count]

obdulia-losantos:
Closes: https://jira.ehealthafrica.org/browse/GTH-164

In the `ProjectStatsViewSet` it was returning the number of entities linked to a submission, so entitites without submission were never counted but they appeared in Gather. Now it returns the real number.

With this approach we could also calculate the statistics for mappings:

```python
class MappingStatsViewSet(SubmissionStatsMixin, viewsets.ReadOnlyModelViewSet):
    queryset = models.Mapping.objects.all()
    serializer_class = serializers.MappingStatsSerializer
    mt_field = 'mappingset__project'

    submissions_field = 'entities__submissions'
    entities_field = 'entities'
```


lordmallam (approved):
LGTM

PR #634, opened 2019-07-24
fix(ui): AVRO schema is restricted to record type

obdulia-losantos:
Closes:
- https://jira.ehealthafrica.org/browse/AET-429
- https://jira.ehealthafrica.org/browse/AET-497

One of the restrictions we have to the input is that the generated AVRO schema must be of type "record" and include a field named "id" of type "string" (internally an UUID4). This cannot be possible if the input is an array or a simple string, those cases derive valid AVRO schemas but not for us.


shawnsarwar (approved):
Nice note in the ticket, thanks!

PR #649, opened 2019-08-06
feat: (kernel, exm) Entity Extraction Module

lordmallam:
Extracts entity extraction from the kernel into its own module.

This PR also includes the following feats:

- Entity Extraction docker config
- Kernel bulk submissions
- Kernel Integration with Redis

#### How it works:

- All artefacts are cached on redis when created.
- Submissions are queued on redis, processed and entities extracted (multi-threaded)
- Generated entities and and updated submissions are bulk uploaded to kernel
- Processed submissions are removed from redis.


shawnsarwar:
@lordmallam I thought you were planning to make a non-django utils library?

aether-entity-extraction-module/entrypoint.sh
```diff
@@ -0,0 +1,122 @@
...
+    after_test
+}
+
+function test_all {
+    pytest
+    after_test
+}
+
+export APP_MODULE=extractor
+export DJANGO_SETTINGS_MODULE="${APP_MODULE}.settings"
```
obdulia-losantos: Do you need these?

aether-entity-extraction-module/extractor/settings.py
```diff
@@ -0,0 +1,49 @@
...
+def get_required(name):
+    try:
+        return os.environ[name]
+    except KeyError as key:
+        raise RuntimeError(f'Missing {key} environment variable!')
+
+
+DEBUG = bool(os.environ.get('DEBUG'))
+TESTING = bool(os.environ.get('TESTING'))
+SECRET_KEY = get_required('DJANGO_SECRET_KEY')
```
obdulia-losantos: is this a django app?

aether-entity-extraction-module/entrypoint.sh
lordmallam: Yes I do aether-entity-extraction-module/extractor/settings.py

lordmallam: Yes I do ==> [Settings](https://github.com/eHealthAfrica/aether/pull/649/files/02895201e0f4a74d4485510e92660ab60cd53c88#diff-4b73dab268e37dbad83df071901c8e7c)

aether-entity-extraction-module/extractor/settings.py
lordmallam: No it is not, but because I have to use feats from django-sdk, it had to be partly setup as a django app, thus the need for a solely python library 😢

obdulia-losantos: As far as I check it, you are only using your `TaskHelper` class and the `aether.sdk.utils.request` method (actually a wrapper of `requests.request`). Does it worth? Why don't create the `TaskHelper` class here and use the `requests` library directly?

lordmallam: `TaskHelper` class is also used in kernel https://github.com/eHealthAfrica/aether/pull/649/files#diff-1910d7e6a2668bff54e163b34f442549R23

shawnsarwar:
@lordmallam Is this all ready for final review?

shawnsarwar (dismissed):
This looks great. *We need to hold of merging* until we get the K8s deployment out of `alpha` as merging this will break the setup. That should be done this week.

obdulia-losantos (approved):

PR #655, opened 2019-08-12
chore: release tag with commit hash

obdulia-losantos:


MrPink (approved):
thanks 🙏

admbtlr (approved):
So now we'll be duplicating the images on DockerHub, once with version number, once with commit hash, correct?

PR #674, opened 2019-09-06
chore: RC latest changes.

shawnsarwar:


obdulia-losantos:
Do we want to continue with 1.5 or begin with 1.6? I mean if this is still multitenancy related...

shawnsarwar:
@obdulia-losantos Still MT related. I wouldn't merge the extractor changes as that's major, but the producer changes are required to launch the MT prod instance on CCloud. I think you're right that this might well be the last release on 1.5.

obdulia-losantos (approved):

PR #678, opened 2019-09-18
fix: (kernel) submission validate catch invalid id

lordmallam:


aether-kernel/aether/kernel/api/views.py
```diff
@@ -400,7 +400,14 @@ def validate(self, request, *args, **kwargs):
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
-        mappingset = get_object_or_404(models.MappingSet.objects.all(), pk=mappingset_id)
+        try:
+            mappingset = get_object_or_404(models.MappingSet.objects.all(), pk=mappingset_id)
+        except Exception as e:
```
obdulia-losantos, line 405: In this case you are catching the 404 exception :scream: 

It's better if you use the `is_uuid` method declared in the filters.

https://github.com/eHealthAfrica/aether/blob/develop/aether-kernel/aether/kernel/api/filters.py#L234

```python
def is_uuid(value):
    try:
        uuid.UUID(value)
        return True
    except ValueError:
        # `value` is not a valid UUID
        return False
```

obdulia-losantos (changes requested):
Almost there :grimacing:

aether-kernel/aether/kernel/api/views.py
lordmallam: Yes, I considered that, and still feel a missing mappingset translates to a bad request since the uuid is passed as the body. You get a `No MappingSet matches the given query.` error in this case

obdulia-losantos: :thinking: 

But is it the same error? 404 not found (with a valid UUID) and 400 Bad Request with a non valid one?

We can live with an 400 error in this case, it's not a big deal ;-)

obdulia-losantos (dismissed):

obdulia-losantos (approved):

PR #683, opened 2019-09-25
feat: (kernel) bulk delete by filters

lordmallam:
Makes available bulk delete by filters

Usage
`/{model_name}/filtered-deleted/?{query_string}`

### DIFF
--- a/aether-kernel/aether/kernel/api/tests/test_views.py
+++ b/aether-kernel/aether/kernel/api/tests/test_views.py
@@ -20,6 +20,7 @@ import dateutil.parser
 import json
 from unittest import mock
 import uuid
+import random
 
 from django.contrib.auth import get_user_model
 from django.core.files.uploadedfile import SimpleUploadedFile
@@ -31,6 +32,7 @@ from rest_framework import status
 
 from aether.kernel.api import models
 from aether.kernel.api.entity_extractor import run_entity_extraction
+from aether.kernel.api.tests.utils.generators import generate_project
 
 from . import (
     EXAMPLE_MAPPING,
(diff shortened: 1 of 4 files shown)


aether-kernel/aether/kernel/api/tests/test_views.py
```diff
@@ -928,3 +930,22 @@ def test_submission_validate(self):
...
+        mapping = models.Mapping.objects.first()
+        filtered_count = models.Entity.objects.filter(mapping=mapping.pk).count()
+        url = f'{reverse("entity-filtered-delete")}?mapping={str(mapping.id)}'
+        response = self.client.delete(url)
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+        self.assertEqual(
+            models.Entity.objects.count(),
+            (entity_count - filtered_count)
+        )
+
```
obdulia-losantos, line 948: could you add a test with no query string and check that everything is deleted? :scream:

aether-kernel/aether/kernel/api/views.py
```diff
@@ -53,7 +53,23 @@
 )
 
 
-class ProjectViewSet(MtViewSetMixin, viewsets.ModelViewSet):
+class FilteredMixin(object):
+    @action(detail=False, methods=['delete'], url_path='filtered-delete')
```
obdulia-losantos: Question, why not `bulk-delete`? :thinking:

obdulia-losantos:
Do you know this library https://github.com/miki725/django-rest-framework-bulk?

It could be a starting point for all the bulk actions (sadly, it's not maintained)

aether-kernel/aether/kernel/api/views.py
```diff
@@ -53,7 +53,23 @@
...
+    def filtered_delete(self, request, *args, **kwargs):
+        qs = super(FilteredMixin, self).get_queryset()
+        filter_class = self.filter_class
+        try:
+            filtered_projects = filter_class(self.request.GET, queryset=qs)
```
obdulia-losantos: stupid comment, have to tried to do directly `qs.delete()` the `get_queryset()` method should have already filtered the query set, I think so :thinking:

obdulia-losantos:
Proposal, why don't you move this piece of code to `aether-django-sdk-library` so we can use it in the rest of modules?

aether-kernel/aether/kernel/api/views.py
lordmallam: 🤷‍♂  Bulk delete in my opinion feels like you have to pass the objects (ids) to be deleted.

obdulia-losantos: right

aether-kernel/aether/kernel/api/views.py
lordmallam: Nope, Filter isn't applied on `get_queryset()`

obdulia-losantos: That's weird...

lordmallam:
@obdulia-losantos I will move this to the django-sdk... Hold on.. 😁

obdulia-losantos:
:scream: :scream: :scream: :scream:

obdulia-losantos (approved):

PR #691, opened 2019-10-06
feat: (odk) Add visualization to the schema definition from XLSForms

lordmallam:
### Add @aether_default_visualization to the schema definition from XLSForms

Allows a custom column `instance::aether_default_visualization` on the xls form with values representing chart formats.

Adds values of the column to the generated schema as `{' @aether_default_visualization': '{value}'}` for corresponding fields

JIRA Ticket: [AET-545](https://jira.ehealthafrica.org/browse/AET-545)


aether-odk-module/aether/odk/api/xform_utils.py
```diff
@@ -649,13 +655,15 @@ def __get_xform_instance_skeleton(xml_definition):
...
     instance = __get_xform_instance(xform_dict, with_root=True)
-    for xpath, has_children in __get_all_paths(instance):
+    for xpath, has_children, visualizer in __get_all_paths(instance):
         schema[xpath] = {
             'xpath': xpath,
             'type': 'group' if has_children else 'string',
             'required': False,
             'label': __get_xform_label(xform_dict, xpath, itexts),
         }
+        if visualizer:
```
obdulia-losantos: This is an internal estructure, you don't need to be so cautious, also either to use the same complicated key name ;-)

```python
        schema[xpath] = {
            # ...
            'visualizer': visualizer,
        }
```

aether-odk-module/aether/odk/api/xform_utils.py
```diff
@@ -297,6 +299,7 @@ def parse_xform_to_avro_schema(
         current_name = xpath.split('/')[-1]
         current_doc = definition.get('label')
         current_choices = definition.get('choices')
+        current_extended_property = definition.get(VISUALIZATION_KEY)
```
obdulia-losantos: Why not `current_visualizer`? (simpler is better)

```python
        current_visualizer = definition.get('visualizer')
```

You only need the `VISUALIZATION_KEY` for the resultant AVRO schema, internally you can use simpler and easier names.

aether-odk-module/aether/odk/api/xform_utils.py
```diff
@@ -314,6 +317,9 @@ def parse_xform_to_avro_schema(
         if current_choices:
             current_field['@aether_lookup'] = current_choices
 
+        if current_extended_property:
```
obdulia-losantos: Also you don't need to check if the value exists, the `__clean_dict` method will remove the entry if the value is `None`. (This is valid also for the `if current_choices` line above)

shawnsarwar (dismissed):

aether-odk-module/aether/odk/api/xform_utils.py
```diff
@@ -955,12 +963,17 @@ def __get_all_paths(dictionary):
...
+            is_dict = isinstance(v, dict)
             if k.startswith('@'):  # ignore attributes
                 continue
             keys = parent_keys + [k]
             xpath = '/' + '/'.join(keys)
-            paths.append((xpath, isinstance(v, dict)))
-            if isinstance(v, dict):
+            if is_dict:
+                if VISUALIZATION_KEY in v:
+                    visualizer = v[VISUALIZATION_KEY]
```
obdulia-losantos: Maybe... to simplify this

```python
visualizer = v.get(VISUALIZATION_KEY) if is_dict else None
```

obdulia-losantos (approved):

PR #692, opened 2019-10-08
chore: merge develop into offline extraction feature branch

admbtlr:


aether-kernel/setup.cfg
```diff
@@ -14,4 +14,4 @@ omit = *migrations*, *tests*
 [coverage:report]
 omit = *migrations*, *tests*, *settings.py
 show_missing = true
-fail_under = 100
+fail_under = 95
```
obdulia-losantos, line 17: One step back...

admbtlr: Ha! True. So what do you think? Should we accept 95%, or write the extra tests?

obdulia-losantos: We can accept 95% but I would suggest to write the missing tests in a near future.

admbtlr: Agreed. Can you approve then please?

obdulia-losantos (approved):

PR #694, opened 2019-10-08
fix: change from using instance:: tag to bind:: with aet namespace

shawnsarwar:
PyXForm did not like our use of instance:: for adding arbitrary tags to instance data. It should work, but it doesn't.

To allow propagation of default_visualization from an XLSForm -> @aether_default_visualization, we're now adding a namespace to the form (`aet`) and then using `bind::aet::{name-of-tag}` -> value which is then included in the schema for that element as `@aether_{name-of-tag}` -> value.


aether-odk-module/aether/odk/api/tests/files/demo-xform.xml
```diff
@@ -1,11 +1,5 @@
-<h:html
-        xmlns="http://www.w3.org/2002/xforms"
-        xmlns:ev="http://www.w3.org/2001/xml-events"
-        xmlns:h="http://www.w3.org/1999/xhtml"
-        xmlns:jr="http://openrosa.org/javarosa"
-        xmlns:odk="http://www.opendatakit.org/xforms"
-        xmlns:orx="http://openrosa.org/xforms"
-        xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
```
obdulia-losantos, line 1: I know that this sounds weird, but I broke this file into several lines to test that our custom xml parser works. The point is to break the tags in different lines not everything in the same looooooooooong line. The previous parser crashed with this case. If you don't mind could you revert the breaking lines I had before :grimacing:

shawnsarwar: No problem, for all the broken up lines or just the ones in this element?

obdulia-losantos: All of them :woman_shrugging:

shawnsarwar: Also, is this kind of pretty-print a standard I can find? My sublime add-on doesn't do anything so nice?

obdulia-losantos: It was manual :hand:

aether-odk-module/aether/odk/api/tests/files/demo-xform.xml
```diff
@@ -324,8 +326,7 @@
     </odk:rank>
     <group ref="/Something_that_is_not_None/iterate">
       <label>Indicate loop elements</label>
-      <repeat
-              nodeset="/Something_that_is_not_None/iterate">
+      <repeat nodeset="/Something_that_is_not_None/iterate">
```
obdulia-losantos: And this one...

shawnsarwar: :+1:

obdulia-losantos (approved):

PR #706, opened 2019-10-18
fix(client): remove bravado libs restrictions

obdulia-losantos:
Closes: https://jira.ehealthafrica.org/browse/AET-572


shawnsarwar (approved):
Yikes, sorry about that.

PR #716, opened 2019-10-23
fix: (exm) multi tenancy entity post

lordmallam:


aether-entity-extraction-module/extractor/manager.py
```diff
@@ -247,7 +234,29 @@ def push_to_kernel(self):
...
+
+            if current_entity_size:
+                entities = [self.PROCESSED_ENTITIES.pop() for _ in range(current_entity_size)]
+                # group entities by realm
+                entities_by_realm = {}
+                for entity in entities:
+                    realm = submission_tenant[entity['submission']]
+                    if realm not in entities_by_realm:
+                        entities_by_realm[realm] = []
+                    entities_by_realm[realm].append(entity)
```
obdulia-losantos: I long time ago a read that the pythonic way of programming is to force the exceptions, not trying to avoid them

```python
try:
    entities_by_realm[realm].append(entity)
except KeyError:
    entities_by_realm[realm] = [entity]
```

In this way you increase the execution speed (it only fails the first time and you don't check the realm condition in each iteration). I think one of the things we want to achieve with this "extractor" is high performance otherwise I would have said nothing. :wink:

aether-entity-extraction-module/extractor/utils.py
```diff
@@ -65,22 +69,30 @@ class Task(NamedTuple):
...
-        headers={'Authorization': f'Token {settings.KERNEL_TOKEN}'},
+        headers=headers,
     )
 
     res.raise_for_status()
     return json.loads(res.content.decode('utf-8'))
 
 
-def get_from_redis_or_kernel(id, type, tenant, redis=None):
+def get_from_redis_or_kernel(id, _type, tenant, redis=None):
```
obdulia-losantos: maybe `model`? or `model_type`? There is a convention in python that the names starting with underscore are private :woman_shrugging:

aether-entity-extraction-module/extractor/utils.py
```diff
@@ -18,8 +18,10 @@
 
 import json
 import collections
+import logging
 from aether.python.redis.task import TaskHelper
 from aether.python.utils import request
+from base64 import b64encode
```
obdulia-losantos: still a thing?

aether-entity-extraction-module/extractor/settings.py
```diff
@@ -37,10 +30,13 @@ def get_required(name):
 REDIS_PORT = os.environ.get('REDIS_PORT', 6379)
 REDIS_DB = os.environ.get('REDIS_DB', 0)
 REDIS_PASSWORD = os.environ.get('REDIS_PASSWORD', '')
-REDIS_INSTANCE = os.environ.get('REDIS_INSTANCE', None)
+REDIS_INSTANCE = os.environ.get('REDIS_INSTANCE')
 KERNEL_TOKEN = os.environ.get('AETHER_KERNEL_TOKEN', '')
 KERNEL_URL = os.environ.get('AETHER_KERNEL_URL', '')
-DEFAULT_REALM = os.environ.get('DEFAULT_REALM', 'aether')
+KERNEL_URL_INTERNAL = os.environ.get('AETHER_KERNEL_URL_INTERNAL', 'http://kernel:8000')
```
obdulia-losantos: Are you using `AETHER_KERNEL_URL_INTERNAL`  or just `AETHER_KERNEL_URL`?

obdulia-losantos (changes requested):
The `exm` tests are not included in travis :scream:

scripts/test_travis.sh
```diff
@@ -33,6 +33,7 @@ case "$1" in
     ;;
 
     core)
+        ./scripts/test_container.sh exm
```
obdulia-losantos, line 36: I would have said that kernel comes first and after the modules that depend on it :wink:

lordmallam: Actually it's the other way around, kernel depends on exm

obdulia-losantos: :open_mouth:

obdulia-losantos (approved):

PR #724, opened 2019-10-30
fix: keycloak admin client location

obdulia-losantos:
The `latest` keycloak version is now the `7.0.1` in which they changed the keycloak folder from `./keycloak` to `/opt/jboss/keycloak`, that made the setup fail.


umarphaarook (approved):
👍

PR #726, opened 2019-11-04
feat: add active flag in Project/Survey and xForm

obdulia-losantos:
Closes: https://jira.ehealthafrica.org/browse/GTH-160

Open questions: 

1- Do we want the same behavior for CouchDB-Sync?
2- Do we want to propagate the active "flag" too?
3- How should UI behave with inactive projects? (Even if so far there is no real implementation for projects)


aether-couchdb-sync-module/aether/sync/api/models.py
```diff
@@ -40,8 +40,8 @@
 +==================+     +==================+
 | project_id       |<-+  | id               |
 | name             |  |  | name             |
-+------------------+  |  | avro_schema      |
-                      |  +~~~~~~~~~~~~~~~~~~+
+| active           |  |  | avro_schema      |
```
lordmallam, line 43: @obdulia-losantos you need to share this trade secret. 😉

lordmallam (approved):

PR #727, opened 2019-11-05
refactor: (kernel) Entity post validation

lordmallam:
improves submission and entity post performance using redis


aether-kernel/aether/kernel/api/models.py
```diff
@@ -688,7 +684,7 @@ def clean(self):
 
     def save(self, *args, **kwargs):
         try:
-            self.full_clean()
+            self.clean()
```
obdulia-losantos, line 691: Why `clean` and not `full_clean`?

https://docs.djangoproject.com/en/2.2/ref/models/instances/#validating-objects

aether-kernel/aether/kernel/api/models.py
```diff
@@ -677,7 +672,8 @@ def clean(self):
         else:
             self.modified = '{}-{}'.format(datetime.now().isoformat(), self.id)
 
-        if self.schema:
+        if self.schemadecorator:
+            self.schema = self.schemadecorator.schema  # redundant values taken from schema decorator
```
obdulia-losantos, line 676: What happens when there is no `schemadecorator`? How do we validate the `payload`? :thinking:

aether-kernel/aether/kernel/api/models.py
lordmallam: `full_clean` revalidates all linked objects, which is already done on the [serializer,](https://github.com/eHealthAfrica/aether/blob/refactor-kernel-entity-post/aether-kernel/aether/kernel/api/serializers.py#L381) causing unnecessary calls to the db.  That is the major strain on the current performance.

obdulia-losantos: Could you write a comment explaining why, otherwise someone would change it back in the future :grimacing:

aether-kernel/aether/kernel/api/models.py
obdulia-losantos: This is what I mean

```python
        if self.schemadecorator:  # redundant values taken from schema decorator	
            self.schema = self.schemadecorator.schema
        if self.schema:
            # validate payload ....
```

aether-kernel/aether/kernel/api/models.py
obdulia-losantos: :+1:

obdulia-losantos (approved):
:crossed_fingers:

PR #734, opened 2019-11-18
fix(storage): file path includes tenant and project id

obdulia-losantos:
Group attachments by project not only by submission


admbtlr (dismissed):
We should bear in mind that this is a breaking change

obdulia-losantos:
We could also run a script to rename all the uploaded files with the new path pattern. :thinking:

admbtlr:
Good idea. Like a kind of migration. I guess this won't actually affect any existing projects (or will it?), so that might not be necessary, but I just wanted to point out that we should increment the major version to make sure we don't trip ourselves up in the future.

shawnsarwar (dismissed):

lordmallam (approved):

PR #738, opened 2019-11-28
chore: allow external collaborators

obdulia-losantos:
As part of: https://jira.ehealthafrica.org/browse/AET-557

The problem with the external collaborators is that we are only building pushing branches in Travis and we need to enable also the "pull request" option.

image

The point here is that we cannot execute the release step in a fork and also we need to change our repo settings to include the Travis CI PR build.

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,8 @@ services:
 branches:
   only:
     - develop
-    - ^release\-[0-9]+\.[0-9]+$
-    - ^[0-9]+(\.[0-9]+){2}$
+    - /^release\-[0-9]+\.[0-9]+$/
+    - /^[0-9]+(\.[0-9]+){2}$/
 
 # use matrix to parallelize tests
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,10 @@ matrix:
         #  branch   develop
         #  branch   release-#.#
         #  tag      #.#.#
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -32,13 +38,16 @@ matrix:
     - name: "Aether release"
       stage: deploy
       # release only in:
-        #  branch   develop
-        #  branch   release-#.#
-        #  tag      #.#.#
-      if: (branch = develop) OR (branch =~ ^release\-[0-9]+\.[0-9]+$) OR (tag =~ ^[0-9]+(\.[0-9]+){2}$)
-      before_install:
```
MrPink, line 39: did you mean to remove this

MrPink (dismissed):

.travis.yml
obdulia-losantos: I moved it to the release script :grimacing: 

https://github.com/eHealthAfrica/aether/pull/738/commits/2a65f0fec83375e5230bd4b7ecd739f7a51f4731#diff-0d28f82f2d1cc0deb12146b212778bc9R81-R87

obdulia-losantos: If we finally move to GitHub actions, I prefer to have all the login in the scripts so the migration will require less steps :crossed_fingers:

shawnsarwar (approved):

PR #747, opened 2019-12-23
refactor: limit MT filtering on entity serializers 

lordmallam:


aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -353,26 +353,6 @@ class EntitySerializer(DynamicFieldsMixin, DynamicFieldsModelSerializer):
         source='submission.attachments',
     )
 
-    project = MtPrimaryKeyRelatedField(
```
obdulia-losantos, line 356: This is required to check that the user is not taking projects from another tenant :(

obdulia-losantos: An option could be to remove the "select" in the form changing the style

https://www.django-rest-framework.org/api-guide/fields/#style

```python
    project = MtPrimaryKeyRelatedField(
        queryset=models.Project.objects.all(),
        required=False,
        style={'input_type': 'text'},
    )
```

:thinking:

obdulia-losantos: Ok, I tried this option and is not working :disappointed:

lordmallam: Yes, i agree  but that is used for the drf web forms and serializer validations (to ensure the artefacts selected are from the same realm) the validation is already handled, relying on redis [here](https://github.com/eHealthAfrica/aether/blob/mt-cache/aether-kernel/aether/kernel/api/serializers.py#L361). The question now is if the drf web forms are import enough for the expensive db calls once the filters are introduced.

obdulia-losantos: We introduced Django ORM cache (#756), would you mind checking if that is enough for the performance issue? There is also another PR (#763) in develop that could be included in release 2.0 to improve the performance.

lordmallam: Profiled with Django ORM Cache, there's some improvements as shown in the image. The difference is with this PR there is no db call for MT validation. But if the current performance is ok @shawnsarwar, then I can close this PR. 

Screenshot 2020-01-13 at 11 21 16

shawnsarwar: What's our next move with this? We need to reduce/ cache the calls somehow. @obdulia-losantos, is there a way we can limit these calls without breaking DRF?

obdulia-losantos: :thinking:

obdulia-losantos: I included this snippet in the multitenancy test:
```python
    def test_views(self):
        # ...

        # try to create an entity linked to a project that belongs to this realm
        entity_data_1 = {
            'project': str(obj1.pk),
            'name': 'playing with realms',
            'status': 'Pending Approval',
            'payload': {
                'id': str(obj1.pk),
            }
        }
        response = self.client.post(
            reverse('entity-list'),
            json.dumps(entity_data_1),
            content_type='application/json'
        )
        self.assertEqual(response.status_code, status.HTTP_201_CREATED, response.json())

        # try to assign entity 1 to a project that belongs to another realm
        response = self.client.patch(
            reverse('entity-detail', kwargs={'pk': entity_1}),
            json.dumps({'project': str(obj2.pk)}),
            content_type='application/json'
        )
        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
        data = response.json()
        self.assertIn('project', data)
        self.assertEqual(
            data['project'],
            [f'Invalid pk "{str(obj2.pk)}" - object does not exist.'],
        )

        # try to create an entity linked to a project that belongs to another realm
        entity_data_2 = {
            'project': str(obj2.pk),
            'name': 'playing with realms',
            'status': 'Pending Approval',
            'payload': {
                'id': str(obj2.pk),
            }
        }
        response = self.client.post(
            reverse('entity-list'),
            json.dumps(entity_data_2),
            content_type='application/json'
        )
        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
        data = response.json()
        self.assertIn('project', data)
        self.assertEqual(
            data['project'],
            [f'Invalid pk "{str(obj2.pk)}" - object does not exist.'],
        )
```

**The tests pass in the `develop` branch.**

If I comment the following out:
```python
class EntitySerializer(DynamicFieldsMixin, DynamicFieldsModelSerializer):
    # ...
    project = MtPrimaryKeyRelatedField(
        queryset=models.Project.objects.all(),
    )
```
I have this:
```text
======================================================================
FAIL: test_views (aether.kernel.api.tests.test_multitenancy.MultitenancyTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/code/aether/kernel/api/tests/test_multitenancy.py", line 185, in test_views
    self.assertIn('project', data)
AssertionError: 'project' not found in {'payload': ['Value must be valid JSON.']}

----------------------------------------------------------------------
 ``` 

So those field entries are required to prevent create/update instances that belong to other realms. We cannot remove them, if we do that we need to include the check in the serializer `save` method.

When I tried the same tests in the `mt-cache` branch I have this error:

```python
Traceback (most recent call last):
  File "/code/aether/kernel/api/tests/test_multitenancy.py", line 185, in test_views
    self.assertEqual(response.status_code, status.HTTP_201_CREATED, response.json())
AssertionError: 400 != 201 : ["['No associated project. Check you provided the correct Submission, Mapping and Schema Decorator']"]
```

obdulia-losantos: The new tests are included in this PR: https://github.com/eHealthAfrica/aether/pull/771

shawnsarwar:
Suggested improvements achieved by caching and incorporated into #788

PR #756, opened 2020-01-09
feat(kernel): enable django ORM cache

obdulia-losantos:


aether-kernel/aether/kernel/settings.py
```diff
@@ -71,11 +71,8 @@
 
 # Redis Configuration
 # ------------------------------------------------------------------------------
-REDIS_DB = os.environ.get('REDIS_DB', 0)
```
shawnsarwar, line 74: @obdulia-losantos Why did this get removed? I think Kernel 2.0 needs this regardless of caching for EXM

shawnsarwar: Ah, because it's in the SDK now?

obdulia-losantos: yeah! it's handled by the env var `REDIS_REQUIRED` included in the `entrypoint.sh` file, I'm a lazy person :grimacing:

shawnsarwar (approved):

PR #758, opened 2020-01-10
chore: drop couchdb sync support

obdulia-losantos:


shawnsarwar (dismissed):

lordmallam (approved):
We should probably keep a branch with this feature

obdulia-losantos:
I was thinking on extracting this to its own repo and make it read only. :thinking:

shawnsarwar:
If we branch and make / close a PR, we should be able to delete the artifacts and have an opportunity to resurrect the branch in the future. Might be harder to find though.

PR #764, opened 2020-01-13
fix(travis): change deployment repository to GCS

obdulia-losantos:
Related to: https://jira.ehealthafrica.org/browse/AET-592

Only deploy kernel, odk, ui and producer (the rest of modules are not deployed)
Pending: push images to GCS not to docker hub.


scripts/release.sh
```diff
@@ -61,34 +61,6 @@ function release_app {
     echo "${LINE}"
 }
 
-function release_gcs {
```
obdulia-losantos, line 64: moved to `deploy.sh` file

MrPink (dismissed):

MrPink (approved):
👏

PR #775, opened 2020-01-19
test: (exm) test coverage multiprocessing

lordmallam:


aether-entity-extraction-module/entrypoint.sh
```diff
@@ -43,9 +43,16 @@ function test_flake8 {
 
 function test {
     rm -R /code/.coverage* 2>/dev/null || true
-    coverage run -m pytest "${@:1}"
+    coverage run \
+    --concurrency=multiprocessing \
+    --parallel-mode \
+    --parallel \
```
obdulia-losantos: it's only `--parallel-mode`; `parallel` is used in the config file.

aether-entity-extraction-module/extractor/tests/test_manager.py
```diff
@@ -149,14 +174,23 @@ def test_entity_extraction(self):
...
+        manager.entity_extraction(self.wrong_task)
+        self.assertEqual(len(manager.PROCESSED_SUBMISSIONS), 1)
+        self.assertNotIn(TENANT, manager.realm_entities)
+        with self.assertRaises(Exception):
+            manager.entity_extraction(self.wrong_mapping_task)
+            self.assertEqual(len(manager.PROCESSED_SUBMISSIONS), 2)
+        self.assertEqual(len(manager.PROCESSED_SUBMISSIONS), 1)
+
     @mock.patch(
         'extractor.utils.kernel_data_request'
```
obdulia-losantos: It's `'extractor.manager.kernel_data_request'`, please add the following in the test to check that you are using that mocked method. You don't need to mock the response.

```python
    @mock.patch('extractor.manager.kernel_data_request', side_effect=Exception)
    def test_push_to_kernel__error(self, mock_kernel_data_request):
        manager = ExtractionManager(self.redis)

        # ...

        mock_kernel_data_request.assert_has_calls([
            mock.call(
                url='entities/',
                method='post',
                data=mock.ANY,
                realm=TENANT,
            ),
            mock.call(
                url='submissions/bulk_update_extracted/',
                method='patch',
                data=mock.ANY,
            ),
        ])
```

aether-entity-extraction-module/setup.cfg
```diff
@@ -31,7 +31,9 @@ description-file = README.md
 
 [tool:pytest]
 python_files = tests/test*.py
-addopts = --maxfail=100 -s --capture=no
+addopts =
+    --maxfail=100
+    -s
 
 [coverage:run]
 branch = true
```
obdulia-losantos, line 39: you could add here the option `parallel =  true` and remove it from the `entrypoint.sh` file

obdulia-losantos (approved):

PR #786, opened 2020-01-21
fix: entity POST performance

shawnsarwar:


shawnsarwar:
Wrong base for merge. Retry.

PR #788, opened 2020-01-22
refactor: increase Kernel Entity input speed

shawnsarwar:
Fixes the django_cacheops settings to optimize input of artifacts.
Changes Base of EntitySerializer from ModelSerializer.
Greatly improves entity input throughput and stability.


aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -294,7 +306,7 @@ def create(self, validated_data):
                 entities.append(entity)
             except Exception as e:
                 errors.append(str(e))
-
+                break
```
obdulia-losantos: wouldn't be faster just...?
```python
            try:
                # ...
            except Exception as e:
                # reject ALL if ANY invalid entities are included
                raise(serializers.ValidationError(e))
```

aether-kernel/aether/kernel/api/validators.py
```diff
@@ -20,7 +20,7 @@
 from aether.python import exceptions, validators
 from django.utils.translation import ugettext as _
 
-from .redis import in_same_project_and_cache
+# from .redis import in_same_project_and_cache
```
obdulia-losantos: :grimacing: :wastebasket:

aether-kernel/aether/kernel/api/validators.py
```diff
@@ -83,37 +83,9 @@ def wrapper_validate_schema_input_definition(data):
...
+            _('Schema Decorator MUST be provided with entities')
         )
-
-    if not _artefacts_in_same_project:
-        raise ValidationError(
-            _('Submission, Mapping and Schema Decorator MUST belong to the same Project')
-        )
-
-    return _possible_project
+    return _schema_decorator.project
```
obdulia-losantos: you still need to check that the rest of provided artefacts belong to the same project...

obdulia-losantos: I mean, the schema decorator is king but if the submission or mapping are present you also need to check that they belong to the same project as the schema decorator.

```python
        _artefact_dict = {	
            'schema_decorators': str(_schema_decorator.pk),	
            'submissions': str(_submission.pk) if _submission else None,	
            'mappings': str(_mapping.pk) if _mapping else None,	
        }	
        if not in_same_project_and_cache(_artefact_dict, _schema_decorator.project):
            raise ValidationError(	
                _('Submission, Mapping and Schema Decorator MUST belong to the same Project')	
            )
```

obdulia-losantos: or simpler
```python
if _submission and _submission.project != _schema_decorator.project:
    raise ValidationError(_('Submission MUST belong to the same Project as Schema Decorator'))

if _mapping and _mapping.project != _schema_decorator.project:
    raise ValidationError(_('Mapping MUST belong to the same Project as Schema Decorator'))
```

aether-kernel/aether/kernel/api/serializers.py
```diff
@@ -281,23 +293,21 @@ class Meta:
 class EntityListSerializer(serializers.ListSerializer):
 
     def create(self, validated_data):
-        errors = []
         entities = []
         # remove helper field and validate entity
+        # only do this once since all schema decoratos are the same.
```
obdulia-losantos: Is this assumption true?
Can't I push entities from different projects?

aether-kernel/aether/kernel/api/tests/test_views.py
```diff
@@ -1013,9 +1013,8 @@ def test_update_by_filters(self):
         )
         response_data = response.json()
         filtered_list = models.Entity.objects.filter(mapping=mapping.pk)
-
-        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
-        self.assertIn('is not a valid choice', response_data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertIn('updated', response_data)
```
obdulia-losantos: does this mean that any "status" is valid? Not only the two ones we had?

aether-kernel/aether/kernel/api/validators.py
```diff
@@ -83,37 +83,38 @@ def wrapper_validate_schema_input_definition(data):
...
-        raise ValidationError(
-            _('Submission, Mapping and Schema Decorator MUST belong to the same Project')
-        )
+        if not _artefacts_in_same_project:
+            raise ValidationError(
+                _('Submission, Mapping and Schema Decorator MUST belong to the same Project')
+            )
 
-    return _possible_project
+    return _project
```
obdulia-losantos, line 120: :+1:

aether-kernel/aether/kernel/api/tests/test_serializers.py
```diff
@@ -337,46 +336,16 @@ def test__serializers__create_and_update(self):
...
-        )
-        self.assertTrue(entity_7.is_valid(), entity_7.errors)
-        with self.assertRaises(ValidationError) as ve:
-            entity_7.save()
-        self.assertIn(
-            'No associated project. Check you provided the correct Submission, Mapping and Schema Decorator',
-            str(ve.exception)
-        )
-
-        entity_8 = serializers.EntitySerializer(
```
obdulia-losantos, line 365: I think you can keep this test :wink:

aether-kernel/aether/kernel/api/serializers.py
shawnsarwar: You're totally right.

aether-kernel/aether/kernel/api/tests/test_views.py
shawnsarwar: @obdulia-losantos, good point. I thought that since we have an enum set in the model for this field that when we do instance.save() we'd get an error. That doesn't seem to be the case. Do we need to perform this validation in the update section of the serializer?

obdulia-losantos: Because you are not using the ModelSerializer you need to declare the status field including the choices.

```python
status = serializers.ChoiceField(choices=models.ENTITY_STATUS_CHOICES)
```

https://www.django-rest-framework.org/api-guide/fields/#choicefield

aether-kernel/aether/kernel/api/serializers.py
obdulia-losantos: Could you add a test in which you send two entities from two different projects and see what's happening there?

aether-kernel/aether/kernel/api/tests/test_serializers.py
```diff
@@ -344,8 +345,12 @@ def test__serializers__create_and_update(self):
             context={'request': self.request},
         )
         self.assertTrue(entity_6.is_valid(), entity_6.errors)
-        entity_6.save()
-        self.assertEqual(entity_6.data['project'], mapping.data['project'])
+        with self.assertRaises(ValidationError) as ve:
+            entity_6.save()
+        self.assertIn(
```
obdulia-losantos: Indent the block inside the with statement.
```python
with self.assertRaises(ValidationError) as ve:
    entity_6.save()
    self.assertIn(
        'Schema Decorator MUST be provided with entities',
        str(ve.exception)
    )
```

aether-kernel/aether/kernel/api/tests/test_serializers.py
```diff
@@ -358,10 +363,9 @@ def test__serializers__create_and_update(self):
         with self.assertRaises(ValidationError) as ve:
             entity_7.save()
         self.assertIn(
-            'No associated project. Check you provided the correct Submission, Mapping and Schema Decorator',
+            'Schema Decorator MUST be provided with entities',
             str(ve.exception)
         )
```
obdulia-losantos: Same comment as above. Why was it wrong? :thinking:

shawnsarwar: Bad paste. Surprised the test passed like this. :man_shrugging:

aether-kernel/aether/kernel/api/tests/test_serializers.py
obdulia-losantos: still pending ;-)

shawnsarwar: :sob:

obdulia-losantos (approved):

PR #795, opened 2020-01-26
fix: EXM Multiprocess / Persist Failure

shawnsarwar:
- Fixed shutdown bug on stop.
- Refactored MP to have an Extraction Process Pool (CPU intensive) and Kernel Pool (Kernel Intensive)
- All stages now persisted in Redis in case of failure.
- Permanent failure (Quarantine in Redis) possible in cases Kernel returns 400 (Bad Request) on submission /kernel publication.
- Don't try to submit entities until their parent submission is accepted
- Quarantine Entities associated with malformed submissions preemptively.
- Stepdown from bulk submission -> single submission by factor of 2 to identify bad artifacts.


lordmallam (approved):
LGTM... That was intense.. 😄

PR #804, opened 2020-02-13
chore: disable logging by default - enable UWSGI stats

MrPink:
- disable logging for successful requests
- enable UWSGI metrics server, the future plan would be so we can monitor UWSGI more closely with Prometheus.

### DIFF
--- a/aether-kernel/conf/uwsgi/config.ini
+++ b/aether-kernel/conf/uwsgi/config.ini
@@ -35,11 +35,16 @@ touch-reload      = /var/run/aether/uwsgi.pid
 chdir             = /code/
 module            = django.core.wsgi:get_wsgi_application()
 
+# Logging
+disable-logging   = true
+if-env            = UWSGI_ENABLE_LOGGING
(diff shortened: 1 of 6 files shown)


aether-kernel/conf/uwsgi/config.ini
```diff
@@ -35,12 +35,17 @@ touch-reload      = /var/run/aether/uwsgi.pid
 chdir             = /code/
 module            = django.core.wsgi:get_wsgi_application()
 
+# Logging
+disable-logging   = true
+if-env            = UWSGI_ENABLE_LOGGING
+disable-logging   = $(UWSGI_ENABLE_LOGGING)
+endif             =
```
obdulia-losantos, line 42: I wonder if in this case the correct expression would be...

```properties
# Logging
disable-logging   = true
if-env            = UWSGI_ENABLE_LOGGING
disable-logging   = false
endif             =
```

:thinking:

obdulia-losantos (approved):

PR #817, opened 2020-03-16
feat(producer): access kernel via REST API

obdulia-losantos:
Reorganize producer source code, split functionality into different modules.

Remove create read-only database user. This script can only be executed by the postgres root user and in most of the cases we have no access to it and require an standard user.

Added option in producer to access Kernel via REST API, not only DB. This option is handled by the environment variable `KERNEL_ACCESS_TYPE` with possible values `api` (default) or `db`. I didn't remove `db` option in order to keep current behavior, but we should consider to avoid this option and finally drop it.


shawnsarwar (approved):
This is excellent.

PR #823, opened 2020-04-23
fix: producer lockups

shawnsarwar:
fix: cast max connections picked up from ENV to integer
fix: set and handle timeouts for requesting a database connection from the pool


shawnsarwar:
Closing for now as many of the issues were addressed by the Producer refactor.

PR #835, opened 2020-05-19
feat(odk): upgrade pyxform@1.1.0

obdulia-losantos:


aether-odk-module/aether/odk/api/xform_utils.py
```diff
@@ -974,19 +982,28 @@ def __get_all_paths(dictionary):
...
             if k.startswith('@'):  # ignore attributes
                 continue
             keys = parent_keys + [k]
             xpath = '/' + '/'.join(keys)
-            paths.append((xpath, isinstance(v, dict)))
+            paths.append((xpath, is_dict or is_list))
             if is_dict:
                 walk(v, keys)
+            elif is_list:
+                # in pyxform 1.x.x there could be duplicated entries like:
```
lordmallam, line 998: What if there aren't duplicates?

obdulia-losantos: If there is only one is the first condition, if they are different... we have a problem.

The reason of having it duplicated is this one: https://github.com/XLSForm/pyxform/pull/380

lordmallam (approved):

PR #839, opened 2020-05-29
fix: rethink error messages level

obdulia-losantos:
Some of the error messages are only a warning and don't require a further action from our side. Let's reduce the sentry entries and only have the issues we should fix, the rest of them will still appear in our server logs.


lordmallam (approved):
Good idea. LGTM

PR #888, opened 2020-08-21
fix: (producer) single entities were not triggering check_updates()

shawnsarwar:
When the `check_update({last_offset})` function was called via the KernelAPI, we use the querystring:
`modified_gt={last_offset}`

This would mean that any returned entity would be new and should trigger an update. However, the subsequent check was for:
`count > 1`.


obdulia-losantos (approved):
:joy:

PR #909, opened 2020-10-30
feat: allow webhooks to create submissions

shawnsarwar:
- feat: allow webhooks to create submissions without special knowledge of our body structure

Often times with webhooks and other similar mechanisms the user has no control over the structure of the request, with the only interface being the URL to which the webhook will publish data. To allow aether to receive such arbitrarily structured data as submissions, we're allowing lists and dicts to be posted directly against `/submissions.json` _if_ they have included the `mappingset` as a query parameter. In the case of a list, it will be treated as a bulk submission, whereas dicts will be treated as a single submission. This does not affect the standard expected format of:
```
{
  "mappingset": "{id}":
  "payload": {}
}

### DIFF
--- a/aether-kernel/aether/kernel/api/views.py
+++ b/aether-kernel/aether/kernel/api/views.py
@@ -373,18 +373,17 @@ class SubmissionViewSet(MtViewSetMixin, FilteredMixin, ExtractMixin, ExporterMix
         return super(SubmissionViewSet, self).get_serializer(*args, **kwargs)
 
     def create(self, request, *args, **kwargs):
-        if isinstance(request.data, list):
-            if (mappingset_id := request.GET.get('mappingset')):
-                request._full_data = [{
+        if (mappingset_id := request.GET.get('mappingset')):
+            if isinstance(request.data, list):
+                  request._full_data = [
+                      {'mappingset': mappingset_id, 'payload': i}
+                      for i in request.data
+                  ]
+            elif not request.data.get('mappingset'):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: aether-kernel/aether/kernel/api/views.py [hunk 1]
```diff
@@ -372,6 +372,21 @@ def get_serializer(self, *args, **kwargs):
...
+                    'mappingset': mappingset_id,
+                    'payload': i
+                } for i in request.data]
+        elif not (mappingset_id := request.data.get('mappingset')):
+            mappingset_id = request.GET.get('mappingset')
+            request._full_data = {
+                'mappingset': mappingset_id,
+                'payload': request.data
+            }
+        return super(SubmissionViewSet, self).create(request, *args, **kwargs)
```
obdulia-losantos:
```suggestion
    def create(self, request, *args, **kwargs):
        if (mappingset_id := request.GET.get('mappingset')):
            if isinstance(request.data, list):
                  request._full_data = [
                      {'mappingset': mappingset_id, 'payload': i}
                      for i in request.data
                  ]
            elif not request.data.get('mappingset'):
                request._full_data = {
                    'mappingset': mappingset_id,
                    'payload': request.data,
                }
        return super(SubmissionViewSet, self).create(request, *args, **kwargs)
```

obdulia-losantos (dismissed):

obdulia-losantos (dismissed):

